PR #12, opened 2020-10-01
fix: update selfsigned transient dependency

thatguynamedandy:
# Checklist:

- [ ] I have added the functionality to the [test app](./blob/master/CONTRIBUTING.md#build-test-application)
- [ ] I have [linked the new component](<(https://github.com/Legal-and-General/canopy/blob/master/CONTRIBUTING.md#invision-dsm)>) to adobe DSM (if appropriate)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22352,12 +22352,6 @@
         "safe-buffer": "^5.1.1"
       }
     },
-    "node-forge": {
-      "version": "0.9.0",
-      "resolved": "https://registry.npmjs.org/node-forge/-/node-forge-0.9.0.tgz",
-      "integrity": "sha512-7ASaDa3pD+lJ3WvXFsxekJQelBKRpne+GOVbLbtHYdd7pFspyeuJHnWfLplGf3SwKGbfs/aYl5V/JCIaHVUKKQ==",
-      "dev": true
-    },
     "node-gyp": {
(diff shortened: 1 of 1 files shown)


elenagarrone (approved):

package-lock.json
```diff
@@ -22352,12 +22352,6 @@
         "safe-buffer": "^5.1.1"
       }
     },
-    "node-forge": {
```
aaronBery, line 22355: node-forge was a top level dependency before. selfsigned has gone up by a patch increment and gained the dependency `node-forge` which is fine,  but i'm trying to work out how this can now be removed from here?

owensgit: @thatguynamedandy  was that `selfsigned` was previously using this top-level `node-forge` install, but the newer version/best practice requires it to be its own specified install?

owensgit (approved):

package-lock.json
thatguynamedandy: I'll be honest I'm not entirely sure why npm moved node-forge down a layer in the dependency tree, I've checked through and there is no reference to the old one. I think I'll go ahead and merge but keep an eye on it.

aaronBery: cool, if there's no reference seems good to go :-)

aaronBery (approved):

PR #17, opened 2020-10-05
fix(quick-link): fixed visited link colour

NemirStephan:
A quick actions visited link colour should match its unvisited colour

This is the same fix as https://github.com/canopy-collective/canopy/pull/403

### DIFF
--- a/projects/canopy/src/lib/quick-action/quick-action.stories.ts
+++ b/projects/canopy/src/lib/quick-action/quick-action.stories.ts
@@ -78,9 +78,14 @@ export const link = () => ({
   `,
   props: {
     content: text('text', 'Send us a message', groupId),
-    link: text('link', '/mailbox', groupId),
+    link: text('link', 'https://google.com', groupId),
     target: select('target', ['_blank', '_self'], '_blank', groupId),
-    icon: select('icon', iconsArray.map(icon => icon.name), 'secure-messaging', groupId),
+    icon: select(
+      'icon',
+      iconsArray.map(icon => icon.name),
+      'secure-messaging',
+      groupId,
+    ),
   },
 });
 
@@ -93,6 +98,11 @@ export const button = () => ({
   `,
   props: {
     content: text('text', 'Load more', groupId),
-    icon: select('icon', iconsArray.map(icon => icon.name), 'repeat', groupId),
+    icon: select(
+      'icon',
+      iconsArray.map(icon => icon.name),
+      'repeat',
+      groupId,
+    ),
   },
 });
--- a/projects/canopy/src/styles/variables/components/_quick-action.scss
+++ b/projects/canopy/src/styles/variables/components/_quick-action.scss
@@ -4,7 +4,7 @@
   --quick-action-hover-color: var(--color-white);
   --quick-action-active-bg-color: var(--color-super-blue-dark);
   --quick-action-active-color: var(--color-white);
-  --quick-action-visited-color: var(--link-visited-color);
+  --quick-action-visited-color: var(--link-color);
   --quick-action-visited-hover-bg-color: var(--color-super-blue-dark);
   --quick-action-visited-hover-color: var(--color-white);
 }


paul-request:
I don't have an issue with this code, but my concern is that the Canopy component is supposed to be a button, but actually spits out a span. So the consumer is left to implement this. However, if the quick action is just a link to a page (which has been the case for 100|% of the quick links in our app) then it should not be a button (semantically and for accessibility). So by saying this is a button so it doesn't need a visited state is incorrect as we have designs that should be implemented as links and therefore do need a distinct visited state.

 Or this component should be changed to output a button if it can only be a button? And if the consumer wants to have links, they could just use the classes perhaps?

thatguynamedandy:
@paul-request I'm not sure that I've seen that, I think the component should be either a button or a span, it should be used as an attribute selector not a component selector. So in the [docs](https://legal-and-general-canopy.netlify.app/?path=/story/components-quick-action--link).

```html
<button lg-quick-action>
  <lg-icon name="repeat"></lg-icon>
  Load more
</button>
```
or
```html
<a lg-quick-action
  href="/mailbox">
  <lg-icon name="secure-message"></lg-icon>
  Send us a message
</a>
```
Does that resolve it? or did I miss something, it shouldn't be a span that sounds a bit dodgy.

paul-request:
> @paul-request I'm not sure that I've seen that, I think the component should be either a button or a span, it should be used as an attribute selector not a component selector. So in the [docs](https://legal-and-general-canopy.netlify.app/?path=/story/components-quick-action--link).
> 
> ```
> <button lg-quick-action>
>   <lg-icon name="repeat"></lg-icon>
>   Load more
> </button>
> ```
> 
> or
> 
> ```
> <a lg-quick-action
>   href="/mailbox">
>   <lg-icon name="secure-message"></lg-icon>
>   Send us a message
> </a>
> ```
> 
> Does that resolve it? or did I miss something, it shouldn't be a span that sounds a bit dodgy.

So in your second example, if we make the visited state the same as unvisited as per this PR, then if you implement as a link like you have(and like we already have), then we will be causing issue I believe with the visited link state.

NemirStephan:
@paul-request and @thatguynamedandy. Do you think we should leave things as they are in canopy and just change MY Account so the quick actions are buttons rather than links? Although, of those quick actions point to external U, in which case they really should stay as links

paul-request:
> @paul-request and @thatguynamedandy. Do you think we should leave things as they are in canopy and just change MY Account so the quick actions are buttons rather than links? Although, of those quick actions point to external U, in which case they really should stay as links

I think the point I am making is that they should be links in our app, and making them buttons to get around the issue of not wanting a visited state is the wrong thing to do. That would also be worse for accessibility.

thatguynamedandy:
Ah ok I think I get you now. So a couple of philosophies that we have taken is, mostly for general knowledge...

1. Avoid redundant parent elements, so sometimes in angular people ignore the host element, meaning a simple alert might look like the following when rendered..
```<alert><div class="alert">{{content}}</div></alert>```
We generally favour host binding classes and not having the wrapper div to make the markup more like
```<alert class="alert">{{content}}</alert>```

2. When the component adds behaviour to existing native elements we generally use component attribute selectors to prevent us manually mapping all of the attributes over. The current button `lg-button` also follows this approach which we borrowed from material ui, but it does open it to abuse. I think the quick action had that in mind, but also to cater for the could be a button, could be a link scenario.

3. We've generally steered clear of relying on consumers to add classes manually, we generally style typography natively  and encapsulate the rest behind components. Partly because we don't really want people to start adding their own overrides, in an ideal world the consumers shouldn't need to add any css.

All of that said, I do see your point, I wonder if we have any examples of quick actions being used as buttons or if they should always be links and therefore have a better designed visited state? I also wonder how important the visited link is in terms of a11y?

paul-request:
Ok, so digging into accessibility a bit more, it would seem almost impossible to proved a meaningful visited state using colour only (https://webaim.org/blog/wcag-2-0-and-link-colors/). So, based on that, it is probably safe to remove it

aaronBery:
> @paul-request I'm not sure that I've seen that, I think the component should be either a `button` or a `link`, it should be used as an attribute selector not a component selector. So in the [docs](https://legal-and-general-canopy.netlify.app/?path=/story/components-quick-action--link).
> 
> ```
> <button lg-quick-action>
>   <lg-icon name="repeat"></lg-icon>
>   Load more
> </button>
> ```
> 
> or
> 
> ```
> <a lg-quick-action href="/mailbox">
>   <lg-icon name="secure-message"></lg-icon>
>   Send us a message
> </a>
> ```
> 
> Does that resolve it? or did I miss something, it shouldn't be a span that sounds a bit dodgy.

Is it worth creating a separate issue to decide on the best semantics for the component going forward?

owensgit (approved):

paul-request (approved):

owensgit (approved):

thatguynamedandy:
Good conversation everyone thank you, I think the consensus is to merge. Piggy backing slightly on the conversation over in [Gov.uk](https://github.com/alphagov/govuk-design-system-backlog/issues/34#issuecomment-399995390). If future testing determines that the visible state adds more to a11y an UX than we have accounted for we can revisit.

PR #19, opened 2020-10-07
feat(button): adds icon control to buttons

thatguynamedandy:
# Description

Allows for the creation of two new button types
- Button with text and icon, with icon positioning via **iconPosition** property
- Button with icon only, set up via **iconButton** property

Also contains a refactor of the stylesheet setup to match other components in the repository. I don't think those changes quantify as a breaking change, however it does remind me that we need to document this better. 

I've configured the storybook knobs to allow users to try out all the different settings, while still keeping the short cuts to the primary variants to match the [DSM](https://legalandgeneral.invisionapp.com/dsm/legalandgeneral/canopy/nav/build/asset/componentContainers/5eb4178bd037c04646b5b9e7/tab/design?mode=edit). There are two storybook knob panels
- **properties** which have a direct mapping to the component properties
- **content** which provide some flexibility to control the content inside the story, but do not match to component properties.

## Requirements

Storybook link: https://deploy-preview-19--legal-and-general-canopy.netlify.app/?path=/story/components-button--primary-button
Design link: https://legalandgeneral.invisionapp.com/share/A4X6XFIYKFV#/416751146_Button
Screenshot: 

Screenshot 2020-10-07 at 11 12 15
Screenshot 2020-10-07 at 11 11 45
Screenshot 2020-10-07 at 11 10 44

# Checklist:

- [x] I have added the functionality to the [test app](./blob/master/CONTRIBUTING.md#build-test-application)
- [x] I have [linked the new component](<(https://github.com/Legal-and-General/canopy/blob/master/CONTRIBUTING.md#invision-dsm)>) to adobe DSM (if appropriate)


projects/canopy/src/lib/button/button.notes.ts
```diff
@@ -16,23 +16,50 @@ Import the component in your module:
...
-~~~
+~~~html
 <a lg-button href="#" [variant]="primary">Link</a>
 ~~~
 
+## Button with icon with text
+
+Icons can be content projected into the button, the button component will then handle the styling of that icon.
+The \`ButtonIconPosition\` property can be used to locate the icon on the left hand side of the text, the default is to the right hand side.
+Content projection is used as Canopy Icons need loading via the iconRegistry which enables tree shaking of icons.
```
owensgit, line 33: _Super minor_, but it would be nice to add syntax highlighting to the `iconRegistry` name.

```suggestion
Content projection is used as Canopy Icons need loading via the `iconRegistry` which enables tree shaking of icons.
```

projects/canopy/src/lib/button/button.notes.ts
```diff
@@ -16,23 +16,50 @@ Import the component in your module:
...
+
+~~~html
+<button lg-button type="button" [ButtonIconPosition]="left" >
```
owensgit, line 36: I think the input name is wrong, could also drop the brackets...

```suggestion
<button lg-button type="button" iconPosition="left" >
```

projects/canopy/src/lib/button/button.notes.ts
```diff
@@ -16,23 +16,50 @@ Import the component in your module:
...
+  Add item
+  <lg-icon name="add" />
+</button>
+~~~
+
+## Button with icon only
+
+Buttons can be setup as icon only buttons, this is done in a similar way to a button with icon and text.
+You will still need to include text content inside the button as this information is vital to screen readers.
+The \`iconButton\` property is used to visually hide that text from sighted users.
```
owensgit, line 46: Maybe drop the bit about sighted users, as there are many impediments that could play into it? Perhaps this bit could just be something like...

> The \`iconButton\` property is used to visually hide that text

projects/canopy/src/lib/button/button.notes.ts
```diff
@@ -16,23 +16,50 @@ Import the component in your module:
...
+~~~
+
 ## Inputs
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
 | \`\`variant\`\` | The variant of button: \`\`solid-primary\`\`, \`\`solid-secondary\`\`, \`\`outline-primary\`\`, \`\`outline-secondary\`\`, \`\`reverse-primary\`\`, \`\`reverse-secondary\`\`; | string | primary | No |
 | \`\`fullWidth\`\` | If the button has to span full width or not. For 'sm' and 'md' sized screens, the button will always be full width and this input has no affect | boolean | false | No |
 | \`\`loading\`\` | If the button shows a loading spinner and is also disabled | boolean | false | No |
+| \`\`iconPosition\`\` | The position of the icon in the button | string | right | No |
```
owensgit, line 62: I think `iconButton` is missing from this table

owensgit (commented):
Nice addition and refactor Andy! 👍 

Just a few comments from me.

projects/canopy/src/lib/button/button.notes.ts
thatguynamedandy: oh good spot, search and replace gone wrong!

projects/canopy/src/lib/button/button.notes.ts
thatguynamedandy: Good call.

projects/canopy/src/lib/button/button.notes.ts
thatguynamedandy: Added 👍

owensgit (dismissed):

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -223,28 +223,47 @@
...
 
           <button lg-button variant="solid-primary" [loading]="true">
-            Test loading button
+            Loading button
           </button>
+          <br />
+
+          <button lg-button variant="solid-secondary">
+            Button with icon
+            <lg-icon [name]="'add'"></lg-icon>
```
aaronBery, line 252: i think Owen mentioned a similar occurrence in that we can drop the brackets and the inner quotes

owensgit (approved):

projects/canopy-test-app/src/app/app.component.html
thatguynamedandy: I think we have a few of these occurrences, I'll be honest I wasn't sure what common convention was. I've swapped it on these ones. 👍

james-hayden-ellis:
Design Review: Button - Icon Only

The button's not quite square, it should be 44x44px - it seems to be both fractionally too wide and too short. The Add icon also seems to be missing its rounded caps - do we have the right SVG uploaded or is it just rendering like that at that particular scale?

Button - Icon

Add Icon

james-hayden-ellis:
Design Review: Button - Icon + Label and Label + Icon

Both buttons seem to have an extra 4px padding on each side...

Button - Icon + Label
Button - Label + Icon

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -9,7 +9,7 @@
             <lg-breadcrumb variant="light" lgMarginBottom="none">
               <lg-breadcrumb-item>
                 <a href="#">
-                  <lg-icon [name]="'home'"></lg-icon>
+                  <lg-icon name='home'></lg-icon>
```
elenagarrone, line 12: Sorry, just my OCD :D I think we always use `"` instead of `'` ;)

projects/canopy/src/lib/button/button.component.scss
```diff
@@ -1,4 +1,127 @@
-@import 'scss-modules/button';
```
elenagarrone, line 1: Nice one, I hated how I set this up! 👍 
Although I think we should mark it as a breaking change since these modules are used in some places (canopy landing page and possibly one more).

elenagarrone (changes requested):
Great work 👍  Just adding the `request changes` flag because I think it should be a breaking change.

projects/canopy/src/lib/button/button.component.scss
thatguynamedandy: I was wondering about that as a breaking change, probably better to be safe rather than sorry. I'll add some notes to the commit.

thatguynamedandy: tagged as a breaking change.

james-hayden-ellis:
Button: Icon-only

Thanks for the fix, looks great 👍🏻

projects/canopy/src/styles/variables/components/button/_button.scss
```diff
@@ -5,5 +5,9 @@
   --btn-disabled-border-color: var(--disabled-color);
   --btn-disabled-color: var(--color-taupe-grey);
 
+  --btn-padding-horizontal: 0.6875rem; // 11
```
aaronBery, line 8: Just looking at the button padding before this PR... it was 

`padding: var(--space-xs);`

If it was previously important to be using a standardised spacing value it feels strange to be discarding that especially for the sake of a pixel... i'm not sure what the right thing to do here is but I think its a good discussion point before it potentially becomes widespread

james-hayden-ellis:
Button: Label + Icon

This seems to have gone the other way, now it’s 8px short either side, plus a pixel short between the label and the icon itself...

Button - Label + Icon

Button: Icon + Label

This one's also gone the other way, but it's 8px short on the left and 9px on the right.

Button - Icon + Label

projects/canopy/src/lib/button/button.component.ts
```diff
@@ -47,7 +47,18 @@ export class LgButtonComponent {
     return this.fullWidth;
   }
 
+  @Input() iconPosition: ButtonIconPosition;
+  @HostBinding('class.lg-btn--icon-left') get leftIconClass() {
+    return this.iconPosition === 'left';
```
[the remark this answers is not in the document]
thatguynamedandy: Links to this https://github.com/Legal-and-General/canopy/pull/20#discussion_r502309398

thatguynamedandy:
Hi @james-hayden-ellis thanks for double checking, I had another crack at it. I haven't got the invision app downloaded to do the alpha fade test, but just to double check these are the spaces to run against?
Screenshot 2020-10-12 at 16 32 08
Screenshot 2020-10-12 at 16 31 56

projects/canopy/src/styles/variables/components/button/_button.scss
thatguynamedandy: The initial t-shirt sizing approach was mostly a development led idea during the early stages. The design team have been pretty happy to adopt it as it all conforms to the 4px design grid, I think it generally works well for things like layout and spacing. However when we get down to the slightly fiddle detail like inside the buttons I've been happy to concede that we might need a bit more granularity. We can keep an eye on it though and try and keep to the pre-defined spaces where possible

james-hayden-ellis:
Hi @thatguynamedandy - There should be 20px on the left, 8px between the label and the icon and 20px on the right, and that goes for both (icon left and icon right) going by our Canopy Sketch Library.

elenagarrone (dismissed):

projects/canopy/src/lib/button/button.component.spec.ts
```diff
@@ -49,6 +49,38 @@ describe('LgButtonComponent', () => {
     });
   });
 
+  describe('the icon position input', () => {
+    describe('when not specified', () => {
+      it('should not set a class modifier', () => {
+        expect(fixture.nativeElement.getAttribute('class')).not.toContain(
+          'lg-btn--icon-',
```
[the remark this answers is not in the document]
thatguynamedandy: good call I've modified the `it`

owensgit (approved):

elenagarrone (approved):

projects/canopy/src/styles/variables/components/button/_button.scss
aaronBery: yeah your right it might be one for further conversations. it might just be the disconnect between our lint rules for using rem and the practice of contorting them to work with a pixel based design that seems strange. either way it's not a show stopper as we are already doing it to a certain degree in the code base

PR #20, opened 2020-10-08
feat(filter-buttons): add filter-buttons component

AlexandruHoisan:
# Description

Provides a set of components to implement filter buttons in a filter pannel.

Also contains a refactor of the storybook tree to match the DSM. The form folder was removed from the storybook tree.

## Requirements

Storybook link: [https://deploy-preview-20--legal-and-general-canopy.netlify.app/?path=/story/components-filter-buttons--select-one](https://deploy-preview-20--legal-and-general-canopy.netlify.app/?path=/story/components-filter-buttons--select-one)
Design link:[https://legalandgeneral.invisionapp.com/share/2SZ192MTHZY#/432421864_Filter_Buttons]( https://legalandgeneral.invisionapp.com/share/2SZ192MTHZY#/432421864_Filter_Buttons)
Screenshot: 

Select One Variant

Screenshot 2020-10-08 at 18 05 21
Screenshot 2020-10-08 at 18 05 40

Select Multiple Variant

Screenshot 2020-10-08 at 18 05 50
Screenshot 2020-10-08 at 18 06 07

# Checklist:

- [x] I have added the functionality to the [test app](./blob/master/CONTRIBUTING.md#build-test-application)
- [x] I have [linked the new component](<(https://github.com/Legal-and-General/canopy/blob/master/CONTRIBUTING.md#invision-dsm)>) to adobe DSM (if appropriate)


projects/canopy-test-app/src/app/app.component.html
```diff
@@ -287,6 +287,22 @@
...
+              Test filter group select one
+              <lg-filter-button value="red">Red</lg-filter-button>
+              <lg-filter-button value="yellow">Yellow</lg-filter-button>
+              <lg-filter-button value="green">Green</lg-filter-button>
+              <lg-filter-button value="blue">Blue</lg-filter-button>
+            </lg-filter-group>
+
+            <lg-filter-group variant="select-multiple">
+              Test filter group select multiple
+              <lg-filter-button formControlName="redFilter" value="true">Red</lg-filter-button>
```
thatguynamedandy, line 300: This is the bit that I wasn't sure about with the `select-multiple`. I wondered if a checkbox group should still just have one form control that has an array of values as opposed to individually wired checkboxes.

I realise we don't currently have a checkbox group, so there is a bit of a gap there.

There is an example of checkbox groups in..
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Examples

some examples on Angular Material as well..
https://material.angular.io/components/checkbox/examples

projects/canopy/src/lib/filter-buttons/filter-button.notes.ts
```diff
@@ -0,0 +1,156 @@
+export const notes = `
+# Filter Button Component
+
+## Purpose
+Provides a set of components to implement filter buttons in a filter pannel. Filter buttons help users to narrow down large data sets, like their documents or payments, by giving them a clear and friendly option for filtering the data. Filter buttons are usually included as part of a filter panel or overlay.
```
thatguynamedandy, line 5:
```suggestion
Provides a set of components to implement filter buttons in a filter panel. Filter buttons help users to narrow down large data sets, like their documents or payments, by giving them a clear and friendly option for filtering the data. Filter buttons are usually included as part of a filter panel or overlay.
```

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+      moduleMetadata({
+        imports: [ReactiveFormsModule, LgFilterButtonModule],
+        declarations: [
+          ReactiveFormFilterSelectOneComponent,
+          ReactiveFormMultipleComponent,
+        ],
+      }),
+    ],
+    'in-dsm': {
+      id: '5f75aab1766566c38799e764',
```
thatguynamedandy, line 26: I think we can map these stories individually to the DSM components. I created the IDs as I'm not sure if you would have the privileges.

single
id: 5f75d9e2a5c09c75c88bd402
https://legalandgeneral.invisionapp.com/dsm/legalandgeneral/canopy/nav/build/asset/componentContainers/5f75d9e2a5c09cad678bd401/tab/angular?mode=edit

multi
id: 5f75d9f774621bd9596c8520
https://legalandgeneral.invisionapp.com/dsm/legalandgeneral/canopy/nav/build/asset/componentContainers/5f75d9f774621b49ea6c851f?mode=edit

I recently did it on the [icon button PR](https://github.com/Legal-and-General/canopy/pull/19/files#diff-9fb0a83f451b7408baea42e9a6b55011R163)

I think this should do it....

```js
export const selectOne = () => ({
 ...
});
selectOne.story = {
  parameters: {
    'in-dsm': {
      id: '5f75d9f774621bd9596c8520'
    }
  }
}
```
I apologise as I realise this setup is bit tricky to test!

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+  template: `
+        <lg-reactive-form
+        [disabled]="disabled"
+        [label]="label"
+        (datesChange)="datesChange($event)">
+    </lg-reactive-form>
+    `,
+  props: {
+    label: text('label', 'Select a date range'),
+    datesChange: action('datesChange'),
```
thatguynamedandy, line 44:
```suggestion
    dateChange: action('dateChange'),
```

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+  template: `
+      <lg-reactive-form-multiple
+      [disabled]="disabled"
+      [label]="label"
+      (typesChange)="typesChange($event)">
+    </lg-reactive-form-multiple>
+    `,
+  props: {
+    label: text('label', 'Select a document type(s)'),
+    typesChange: action('typesChange'),
```
thatguynamedandy, line 59:
```suggestion
    typeChange: action('typeChange'),
```

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
thatguynamedandy: I think the singular is more standard, but super minor.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button-select-multiple.stories.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.form.get('policy');
+  }
+
+  @Output() typesChange: EventEmitter<void> = new EventEmitter();
+
+  form: FormGroup;
+
+  constructor(public fb: FormBuilder) {
+    this.form = this.fb.group({
+      p60: [{ value: '', disabled: false }],
```
thatguynamedandy, line 65: I wondered if we should make these a  bit more generic, to just 'option 1', 'option 2' etc. Or in the [past I have used brand colours as options](https://legal-and-general-canopy.netlify.app/?path=/story/components-form-radio--standard).

It's very minor, but might be nice if we have some form of consistency.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
```diff
@@ -0,0 +1,99 @@
...
+  font-size: var(--text-fs--8-size);
+  margin-bottom: var(--space-xs);
+  cursor: pointer;
+
+  .lg-icon {
+    vertical-align: sub;
+    margin-right: var(--space-xxxs);
+  }
+
+  .filter-button__icon-selected {
```
thatguynamedandy, line 23:
```suggestion
  .lg-filter-button__icon-selected {
```
It's probably good practice to add the `lg` prefix to these. We do this mostly as `ViewExncapsulation` is currently set to `None` which means there is a risk of class clashing where people use a different library alongside Canopy.  Unlikely with such specific names, but it does keep it consistent.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
```diff
@@ -0,0 +1,99 @@
...
+.lg-filter-button__input {
+  @include lg-visually-hidden;
+}
+
+.lg-filter-button__label {
+  display: inline-block;
+  max-height: var(--space-lg);
+  border: solid var(--border-width) var(--filter-button-bg-color-active);
+  border-radius: var(--space-md);
+  padding: var(--space-xxxs) var(--space-sm) var(--space-xxxs) var(--space-sm);
```
thatguynamedandy, line 12:
```suggestion
  padding: var(--space-xxxs) var(--space-sm);
```
I think this can be shortened

projects/canopy/src/lib/filter-buttons/filter-button.interface.ts
```diff
@@ -0,0 +1 @@
+export type FilterVariant = 'select-one' | 'select-multiple';
```
thatguynamedandy, line 1: We have an ongoing discussion where we think these should be enums so that they can be referenced in code.
```ts
enum FilterVariant {
  SelectOne,
  SelectMultiple
}
```
Which means in the code we can do this instead of referring to the string value...
```ts
if (this.variant === FilterVariant.SelectMultiple) {
```
This comment is more for awareness, although I think this approach might be better, I'd rather we log it in an issue and then fix throughout the codebase, cc @aaronBery who I think has looked into it too)

projects/canopy/src/lib/filter-buttons/filter-group.component.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../styles/mixins';
+
+.lg-filter-group {
+  display: block;
+  margin-bottom: var(--component-margin);
+
+  .lg-label {
```
thatguynamedandy, line 7: I might sound a bit fussy on this one, but where possible we try and avoid the components being aware of each other on the css. An easy way round it is to add the [LgMarginDirective](https://legal-and-general-canopy.netlify.app/?path=/story/directives--margin)  and then override the default margin that way. This adds a utility class which helps the components to have a slightly looser coupling. 

```html
  <legend lg-label [attr.for]="id" lgMarginBottom="xs">
    <ng-content></ng-content>
  </legend>
```

projects/canopy/src/lib/forms/forms.module.ts
```diff
@@ -17,6 +18,7 @@ import { LgValidationModule } from './validation/validation.module';
     LgInputModule,
     LgLabelModule,
     LgRadioModule,
+    LgFilterButtonModule,
```
thatguynamedandy, line 21: For anyone else reviewing we had an interesting discussion around wether to deprecate this LgFormsModule. It's useful in that it's just one module to import, but as it grows in size people may end up increasing their bundle size with form components they aren't using. Unless of course the tree shaking is more clever than I think.

thatguynamedandy (commented):
@AlexandruHoisan I'm super impressed by this PR! Amazing for a first time contribution. 🎉  I had a few comments and general discussion points. The only potentially tricky comment is the one about whether the `select-multiple` variant should behave more like a checkbox group, e.g. have an array of options as opposed to multiple form controls.

projects/canopy-test-app/src/app/app.component.html
AlexandruHoisan: The filter checkbox group is built similar to the ones in the examples above. I did some changes and now you can implement the select-multiple variant either way, with each checkbox with its individual formcontrol as it is now(I left this working as it is, because I saw that in the majority of cases, users were using them like this), and with a formArray and we push the values in that array and we get the array of values like above. For the second implementation should I also update the notes on how the user should implement it on their side, as it needs a bit of logic with the pushing and removing of values in the formArray?

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
AlexandruHoisan: I am not sure how to test those, any pointers will be appreciated.:)

projects/canopy/src/lib/filter-buttons/filter-button/filter-button-select-multiple.stories.ts
AlexandruHoisan: Will change them to be more generic.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
AlexandruHoisan: Will keep in mind in the future.

projects/canopy-test-app/src/app/app.component.html
thatguynamedandy: It probably should be reflected in the notes, it might be worth checking with @aaronBery, @owensgit and @elenagarrone, just in case I am wrong on the array approach which I assumed would be the more commonly used one, but I may be wrong 😀

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
thatguynamedandy: to be honest it's quite tricky, the `npm run dsm-storybook:preview` might help. In terms of docs there is a bit in the repo.

https://github.com/Legal-and-General/canopy/blob/master/CONTRIBUTING.md#invision-dsm

Which links out to more comprehensive docs on invision..

https://support.invisionapp.com/hc/en-us/articles/360028510211-Configuring-the-Storybook-DSM-Integration

projects/canopy-test-app/src/app/app.component.html
AlexandruHoisan: Either way, I think it's better to have both and let the user choose. I would also choose the array one now, as it makes more sense for our needed implementation, but I guess it depends on the users scenario. 
The notes will look something like this:
Screenshot 2020-10-12 at 12 33 42

projects/canopy/src/lib/filter-buttons/filter-button.interface.ts
elenagarrone: I personally don't agree with the enum but I think I wrote my reasons in an issue. I'll try to find it as this is not the first time I wrote about it ;) Definitely needs a decision but I think we can leave this as is for now.

projects/canopy/src/lib/filter-buttons/filter-group.component.html
```diff
@@ -0,0 +1,7 @@
+<fieldset [attr.aria-describedby]="ariaDescribedBy">
```
elenagarrone, line 1: I've noticed that in storybook the code is appearing as `<fieldset aria-describerdby>`. If `ariaDescribedBy` is undefined I believe is better to hide the attribute as this might confuse the screen readers.

 I would simply do something like: `<fieldset [attr.aria-describedby]="ariaDescribedBy || null">`

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
```diff
@@ -0,0 +1,21 @@
+<input
+  (click)="onCheck()"
+  [attr.aria-describedby]="variant === 'select-multiple' ? ariaDescribedBy : null"
```
elenagarrone, line 3: Here as well. If `ariaDescribedBy` is undefined I would set it to `null`.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
```diff
@@ -0,0 +1,21 @@
...
+  [attr.checked]="checked === true ? true : null"
```
elenagarrone, line 4: Could this be `[attr.checked]="checked || null"`?

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
```diff
@@ -0,0 +1,21 @@
...
+  [attr.disabled]="disabled === true ? '' : null"
```
elenagarrone, line 5: And this as well.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
```diff
@@ -0,0 +1,21 @@
...
+  [attr.value]="value"
+  class="lg-filter-button__input"
+  [type]="variant === 'select-multiple' ? 'checkbox' : 'radio'"
+/>
+
+<label [attr.for]="id" class="lg-filter-button__label">
+  <span class="filter-button__icon-wrapper">
+    <lg-icon class="filter-button__icon-add" name="add"></lg-icon>
+    <lg-icon class="filter-button__icon-subtract" name="subtract-spot-outline"></lg-icon>
+    <lg-icon class="filter-button__icon-selected" name="checkmark"></lg-icon>
```
elenagarrone, line 17: I think the `add` and `selected` icons could use the `checked` value to decided wether to be added to the DOM or not instead of always be present and handled by the css.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
```diff
@@ -0,0 +1,99 @@
+@import '../../../styles/mixins';
+
+.lg-filter-button__input {
+  @include lg-visually-hidden;
+}
+
+.lg-filter-button__label {
+  display: inline-block;
+  max-height: var(--space-lg);
+  border: solid var(--border-width) var(--filter-button-bg-color-active);
```
elenagarrone, line 10: We usually use `btn` instead of `button` in our variables ;)

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
elenagarrone: Yes, I think it's better to add the prefix so that it's consistent with what we've done until now 👍

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+@Component({
+  selector: 'lg-filter-button',
+  templateUrl: './filter-button.component.html',
+  styleUrls: ['./filter-button.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LgFilterButtonComponent implements OnInit, ControlValueAccessor {
+  @Input() checked = false;
+  @Input() id = `lg-filter-button-${++nextUniqueId}`;
```
elenagarrone, line 31: The `id` and `name` don't match because you are increasing the `nextUniqueId` in both places. You might want to save the increased value into a variable and then refer to that.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    @Self() @Optional() public control: NgControl,
+    private filterGroup: LgFilterGroupComponent,
+    private domService: LgDomService,
+    private errorState: LgErrorStateMatcher,
+    @Optional()
+    @Host()
+    @SkipSelf()
+    private controlContainer: FormGroupDirective,
+  ) {
+    if (this.control != null) {
```
elenagarrone, line 72: Could we use `!==`?

projects/canopy/src/lib/forms/forms.module.ts
elenagarrone: I definitely agree on the `LgFormsModule`. I can't see any need for it.

projects/canopy-test-app/src/app/app.component.html
elenagarrone: I agree with @thatguynamedandy , I think having the array would be enough but I think I'm ok leaving both the solutions.

elenagarrone (commented):
Great work @AlexandruHoisan 🎊 Just few small comments from me 👍

projects/canopy/src/lib/filter-buttons/filter-group.component.ts
```diff
@@ -0,0 +1,153 @@
...
+})
+export class LgFilterGroupComponent implements ControlValueAccessor {
+  private _name = `lg-filter-group-${nextUniqueId++}`;
+
+  @Input() id = `lg-filter-group-id-${nextUniqueId++}`;
+  @Input() disabled = false;
+  @Input() ariaDescribedBy: string;
+  _variant: FilterVariant = 'select-one';
+  @Input()
+  set variant(variant: FilterVariant) {
```
owensgit, line 39: Just wondering why we're using a getter + setter on the variant input? It doesn't seem to be doing any logic.

owensgit (dismissed):
I know we've been through the code together already, but just one small question from me 🙂 

Also: I can see there's a separate commit for adding this component to the public API, so I'm wondering, should we look into adding it same commit that contains all the new component changes (**c5cef73**)? Reason I suggest is that the Semantic Release plugin will create a new version of Canopy at this commit, which will technically be broken when it is packaged up, as the component won't be available to use. Would be good to get @thatguynamedandy , @aaronBery  and @elenagarrone 's views on this too.

thatguynamedandy:
I agree with @owensgit on the commits I think we should just aim to have two...
```
feat(filter-buttons): add filter-buttons component 
refactor(form-storybook): remove form folder from storybook tree 
```
The second one is quite distinct, but we could bundle all the other functionality into one.

projects/canopy/src/lib/filter-buttons/filter-button.interface.ts
aaronBery: As far as I can remember @elenagarrone (and I may be wrong) you mentioned a speaker at one of the angular meet-ups was an advocate for types over enums but I don't think you managed to follow up on what the rationale was. If you can dig that out it would be amazing :-) the only downside of enums I can think of is that I imagine it adds to bundle size whereas types are just annotations. However there are some clear disadvantages too. In reviewing this PR I'm not using my IDE so values that have type security can look like magic strings when read on a GitHub diff which adds an overhead in validating that they are not. either way would be good to hear the arguments 👍

thatguynamedandy: Moving to [here](https://github.com/Legal-and-General/canopy/issues/25) for now just to ease the PR through

projects/canopy/src/lib/filter-buttons/filter-group.component.ts
AlexandruHoisan: they were left behind after I tested something. Got rid of them. :)

projects/canopy/src/lib/filter-buttons/filter-button.notes.ts
AlexandruHoisan: Done

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
AlexandruHoisan: Done

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button-select-multiple.stories.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-group.component.scss
AlexandruHoisan: Done. Will use this in the future in those kind of cases.

projects/canopy/src/lib/filter-buttons/filter-group.component.html
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.html
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.scss
AlexandruHoisan: Good to know. Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
AlexandruHoisan: Nice one. Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button-select-multiple.stories.ts
```diff
@@ -0,0 +1,51 @@
...
+  template: `
+    <form [formGroup]="form">
+      <lg-filter-group
+        [variant]="variant"
+        [valueArrayName]="valueArrayName"
+        [formArrayName]="valueArrayName"
+      >
+        {{ label }}
+        <lg-filter-button value="red">Red</lg-filter-button>
+        <lg-filter-button value="yellow">Yellow </lg-filter-button>
```
thatguynamedandy, line 17: minor again, but some spaces around the text (and on Blue)...
```suggestion
        <lg-filter-button value="yellow">Yellow</lg-filter-button>
```

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,129 @@
...
+let uniqueId = 0;
+
+@Component({
+  selector: 'lg-filter-button',
+  templateUrl: './filter-button.component.html',
+  styleUrls: ['./filter-button.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LgFilterButtonComponent implements OnInit, ControlValueAccessor {
+  nexUniqueId = ++uniqueId;
```
thatguynamedandy, line 32: tiny typo..
```suggestion
  nextUniqueId = ++uniqueId;
```

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (this.variant === 'select-multiple') {
+      if (this.control) {
+        return this.errorState.isControlInvalid(this.control, this.controlContainer);
+      } else {
+        return this.errorState.isControlInvalid(this.controlArray, this.controlContainer);
+      }
+    } else {
+      return this.errorState.isControlInvalid(this.control, this.controlContainer);
+    }
+  }
```
thatguynamedandy, line 64: This is more of just a style thing, so take it or leave it, but I'm a fan of the early return here as I think it makes it a bit more readable
```suggestion
  if (this.variant === 'select-multiple') {
    if (this.control) {
      return this.errorState.isControlInvalid(this.control, this.controlContainer);
    } 
    return this.errorState.isControlInvalid(this.controlArray, this.controlContainer);
  }
  return this.errorState.isControlInvalid(this.control, this.controlContainer);
}
```

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -287,6 +287,22 @@
...
 
+            <lg-filter-group formControlName="filter">
+              Test filter group select one
+              <lg-filter-button value="red">Red</lg-filter-button>
+              <lg-filter-button value="yellow">Yellow</lg-filter-button>
+              <lg-filter-button value="green">Green</lg-filter-button>
+              <lg-filter-button value="blue">Blue</lg-filter-button>
+            </lg-filter-group>
+
+            <lg-filter-group variant="select-multiple" valueArrayName="filters" formArrayName="filters">
```
thatguynamedandy, line 298: In my head I expected this to be like the below, but I haven't at this stage done any background research or really looked into it, so I may be talking nonsense, I'll have a read to get my head around it more. 
```
<lg-filter-group variant="select-multiple" formControlName="filter">
```

projects/canopy/src/lib/filter-buttons/filter-group.component.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (this.variant === 'select-multiple' && this.valueArrayName) {
+      const formArray: FormArray = this.controlContainer.control.get(
+        this.valueArrayName,
+      ) as FormArray;
+
+      if (event.target.checked) {
+        formArray.push(new FormControl(event.target.value));
+      } else {
+        let i = 0;
+        formArray.controls.forEach((ctrl: FormControl) => {
```
thatguynamedandy, line 134: Do you need to assign `i`? or can you do the below?
```suggestion
        formArray.controls.forEach((ctrl: FormControl, i) => {
```

projects/canopy-test-app/src/app/app.component.html
AlexandruHoisan: to get the values in an array, like ['redFilter', 'greenFilter'], we need to use a control container, either a form group or a form array, so we can contain more controls.The FormControlName, on the filter-group component, is only supported with radio buttons as they change the same value on the same control, as opposed to pushing and removing values.

projects/canopy/src/lib/filter-buttons/filter-group.component.scss
```diff
@@ -0,0 +1,25 @@
...
+    display: inline-block;
+    margin-right: var(--space-xs);
+  }
+
+  .lg-filter-button:last-of-type {
+    margin-right: 0;
+  }
+}
+
+.lg-filter-group.no-icons {
```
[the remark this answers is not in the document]
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button.stories.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export default {
+  title: 'Components/Filter Buttons',
+  parameters: {
+    decorators: [
+      withKnobs,
+      moduleMetadata({
+        imports: [ReactiveFormsModule, LgFilterButtonModule],
+        declarations: [
+          ReactiveFormFilterSelectOneComponent,
```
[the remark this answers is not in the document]
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button-select-multiple.stories.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-button/filter-button.component.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/filter-buttons/filter-group.component.ts
AlexandruHoisan: no need to assign it, worked like a charm.

thatguynamedandy:
@AlexandruHoisan are you ok if I close this one down for now?

AlexandruHoisan:
> @AlexandruHoisan are you ok if I close this one down for now?
Yea, I think thats for the best at the moment. 👍

PR #45, opened 2020-10-23
fix(input): match height with designs

### LINKED ISSUE #43: Input fields and buttons should be the same height
At some stage out code implementation has deviated from the designs.

Input field and buttons should be the correct height.

Currently our button implementation is correct (calculated height of 44px) however our input fields need to be aligned with design.

owensgit:
# Description

Fixes an issue where inputs were set at the wrong height (50px) instead of the height specified in the designs (44px).

Fixes #43

## Requirements

Storybook link: https://deploy-preview-45--legal-and-general-canopy.netlify.app/
Design link: https://legalandgeneral.invisionapp.com/d/main#/console/19910472/416751159/inspect
Screenshot:
Screenshot 2020-10-23 at 15 42 09

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/input/input.scss
+++ b/projects/canopy/src/lib/forms/input/input.scss
@@ -4,8 +4,8 @@
   display: block;
   border: solid var(--border-width) var(--border-color);
   border-radius: var(--border-radius-sm);
-  line-height: var(--line-height-sm);
-  padding: var(--space-xs) var(--space-sm);
+  line-height: var(--input-line-height);
+  padding: var(--input-vertical-padding) var(--space-sm);
   outline: 0;
   max-width: 100%;
 
@@ -22,9 +22,7 @@
   &:focus {
     border-left-width: var(--keyline-width);
     border-color: var(--border-focus-color);
-    padding-left: calc(
-      var(--space-sm) - var(--keyline-width) + var(--border-width)
-    );
+    padding-left: calc(var(--space-sm) - var(--keyline-width) + var(--border-width));
 
     @include lg-outer-focus-outline();
   }
--- a/projects/canopy/src/styles/variables/_main.scss
+++ b/projects/canopy/src/styles/variables/_main.scss
@@ -43,6 +43,8 @@
 
   --keyline-width: 0.1875rem;
 
+  --form-field-vertical-padding: 0.5625rem; // 9px
+
   /* Separator */
   --separator-color: var(--color-platinum);
 }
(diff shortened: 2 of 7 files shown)


projects/canopy/src/styles/variables/_main.scss
```diff
@@ -43,6 +43,9 @@
 
   --keyline-width: 0.1875rem;
 
+  --input-line-height: 1.5;
+  --input-vertical-padding: 0.5625rem; // 9px
```
owensgit, line 47: This follows the same approach used for buttons, where we tweak the padding to match the required height. The value `0.5625rem` is the same in both. I did consider creating a new CSS var to contain this value, but I couldn't think of a nice location and name for the var and how it could sit alongside other vars long-term. In the end, I felt that just having the value twice isn't so bad. Happy to chat about this though!

thatguynamedandy: you could put these into a new file at `'variables/components/input';`?
I might pass over to @elenagarrone to review and spend a bit more time thinking about it while I am off, but aside from that I'm ok with the approach, and thanks for picking this up @owensgit

projects/canopy/src/lib/forms/input/input.scss
```diff
@@ -4,8 +4,8 @@
   display: block;
   border: solid var(--border-width) var(--border-color);
   border-radius: var(--border-radius-sm);
-  line-height: var(--line-height-sm);
-  padding: var(--space-xs) var(--space-sm);
+  line-height: var(--input-line-height);
```
elenagarrone, line 7: Would this need to be done also for the select style?

owensgit: Ah, that's a good point, I'll look into that 👍

elenagarrone (approved):
Looks good to me :) Thanks for doing this 👍

projects/canopy/src/styles/variables/_main.scss
owensgit: Thanks @thatguynamedandy, I've created that new file. @elenagarrone also suggested taking a look at the **select** component - which turned out to be the wrong height too, so I've updated that and created `'variables/components/select'`. As the vertical padding values are the same across 3 places now, I've stored that in a new shared variable alongside the other generic form variables.

AlexandruHoisan (approved):

owensgit:
@thatguynamedandy @aaronBery - I think I just need one more thumb from one you for this PR to be mergeable 🙂

thatguynamedandy (approved):
Nice one @owensgit, good fix!

PR #61, opened 2020-11-12
Upgrade to Prettier 2.1.2

### LINKED ISSUE #42: Dependencies to remove
Hi all, I'm just using this issue to log dependencies that we should probably consider removing as they are appear to be stale or no longer maintained.

**Tslint** (we should move to ESlint soon anyway)

- "tslint-eslint-rules"
- "tslint-import-group-ordering"

owensgit:
# Description

This is split into 3 commits which center around the update of Prettier to v2.1.2 - originally raised by Dependabot in [this PR](https://github.com/Legal-and-General/canopy/pull/29).

1. **npm(deps-dev): bump prettier from 1.19.1 to 2.1.2** 
Original commit that bumps Prettier version
2. **style(linter): disable ts-lint rules that conflict with prettier v2** 
Prettier's new formatting rules clash with ts-lint, this commit removes those ts-lint rules, allowing Prettier to format the code without linting errors
3. **style(formatting): reformat all code with Prettier v2**
Runs Prettier on the whole codebase, reformatting all files under the new rules. I wrote a [general summary of the changes](https://github.com/Legal-and-General/canopy/pull/29#issuecomment-724117084) in the original Dependabot PR.

## Requirements

I recommend looking at the changes in each commit individually, as the wholesale Pretter formating commit is huge and just contains the reformatting in one go.

Storybook link: https://deploy-preview-61--legal-and-general-canopy.netlify.app/

# Checklist:


tslint.json
```diff
@@ -19,49 +18,6 @@
       "rxjs/Rx",
       "rxjs/internal/operators"
     ],
-    "import-group-ordering": {
```
elenagarrone, line 22: Why did you remove `"tslint-import-group-ordering"? `Was it causing issues? It's quite a good package and I really like the grouping of imports :(

package.json
```diff
@@ -120,6 +120,7 @@
     "ts-node": "^9.0.0",
     "tsickle": "^0.39.0",
     "tslint": "^6.1.3",
+    "tslint-config-prettier": "^1.18.0",
```
thatguynamedandy, line 123: I just had a query on this one, and I'm a bit out of the loop on Angular eslint/tslint. This dependency looks like it's on the decline with no updates for a couple of years, but should the future plan be to remove tslint and therefore this dependency anyway? If this PR is purely concerned with Prettier then I think thats ok and we can defer the eslint one until later, just wanted to check with others (cc @elenagarrone)

tslint.json
owensgit: I removed it as part of #42 , as it's been flagged as potentially stale and no longer maintained. I think it's also not compatible with ES Lint which is what we'll be moving toward. But having said that, perhaps we could just look at removing or replacing `tslint-import-group-ordering` when we do the ES Lint migration, and leave it as is for now? I don't think it's causing any issues.

package.json
elenagarrone: In my opinion this is fine for now as we will tackle eslint in a separate PR (hopefully soon).

owensgit: Yes I saw this is as more of a stop gap until we do the migration to ES Lint, in which case we'd need to swap this out for the ES Lint equivalent one - [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier).

tslint.json
elenagarrone: I would personally prefer to do that. In MYA we were able to use another package once moved to eslint so this could be part of the migration from `tslint` to `eslint` (cc @thatguynamedandy).

owensgit: Sounds good, also great to hear you're already using a replacement package

owensgit: I've removed that commit for now, and we can revisit when we do the migration to ES Lint 👍

thatguynamedandy: works for me 👍

thatguynamedandy (approved):
Nice one @owensgit thanks for tackling this one!

elenagarrone (approved):
Thanks Owen, great work 👍

PR #77, opened 2020-11-23
Adds Input suffixes and prefixes

thatguynamedandy:
# Description

Initially raising this as a draft PR to get some visibility on it, it proved to be slightly larger than expected and any help getting it over the line would be most appreciated. It's probably worth reviewing one commit at a time.

Outstanding work
- [x] Check up on a11y best practice
- [x] Cross browser testing
- [x] Self review of the code

## Requirements

Create the ability to add input prefixes and suffixes to input fields which may contain buttons. To enable this work a couple of smaller commits were made.
- Addition of a small button variant
- Addition of a small spinner variant
- Addition of disabled property to button
- Addition of an 'add-on' button variant. 

As you can see from the designs the buttons are within the borders of the input field. This left me with two possible implementation options.
1. Absolutely position the input suffix/prefix inside the field. This would work except that the suffix could contain text and therefore be of a variable size, we would have to use JavaScript to detect the size the suffix before offsetting it
2. Add a wrapper around the input field, and move the focus events up to the wrapping element. We can then put the input fields, suffixes and prefixes inside that wrapper and use flexbox to control them.

I ended up going with option 2 as I think option 1 would lead to a flicker as the JavaScript calculated the suffix width.

From an a11y point of view I have linked the prefix and suffix to the described-by attribute of the input field. I don't treat the button any different from the text span, although I think a bit more testing on this would be nice. There is a similar [discussion on GOV.uk at](https://github.com/alphagov/govuk-design-system-backlog/issues/134) the moment

Screenshot 2020-11-23 at 21 46 01Screenshot 2020-11-23 at 21 45 23Screenshot 2020-11-23 at 21 46 19

https://deploy-preview-77--legal-and-general-canopy.netlify.app/?path=/story/components-form-input--with-button-suffix

https://legalandgeneral.invisionapp.com/d/main#/console/19910472/416751159/preview

# Checklist:


alexcanning:
Should I be able to do this?

2020-11-24_11-33-38_am

alexcanning:
Focus on secondary buttons falls beneath primary button

2020-11-24_11-37-43_am

thatguynamedandy:
https://github.com/Legal-and-General/canopy/pull/77#issuecomment-732885220
I fixed the focus issue.
Screenshot 2020-11-25 at 10 23 52

thatguynamedandy:
https://github.com/Legal-and-General/canopy/pull/77#issuecomment-732884499

You can in theory put more suffixes in than you would ever need, there is some responsibility on the user to be sensible. I fixed the alignment issue in that screen shot though.

projects/canopy/src/lib/button/button.interface.ts
```diff
@@ -1,10 +1,11 @@
-export type Variant =
+export type ButtonVariant =
```
elenagarrone, line 1: My personal preference would be to have the types without the `Button` prefix since they all come from `button.interface.ts` but not too fussed about it.

projects/canopy/src/lib/button/button.notes.ts
```diff
@@ -56,8 +56,10 @@ The \`iconButton\` property is used to visually hide that text.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
-| \`\`variant\`\` | The variant of button: \`\`solid-primary\`\`, \`\`solid-secondary\`\`, \`\`outline-primary\`\`, \`\`outline-secondary\`\`, \`\`reverse-primary\`\`, \`\`reverse-secondary\`\`; | string | primary | No |
+| \`\`variant\`\` | The variant of button: \`\`solid-primary\`\`, \`\`solid-secondary\`\`, \`\`outline-primary\`\`, \`\`outline-secondary\`\`, \`\`reverse-primary\`\`, \`\`reverse-secondary\`\`, \`\`add-on\`\`; | string | solid-primary | No |
+| \`\`size\`\` | The size of the button | ButtonSize [\`\`sm\`\`, \`\`md\`\`] | \`\`md\`\` | No |
 | \`\`fullWidth\`\` | If the button has to span full width or not. For 'sm' and 'md' sized screens, the button will always be full width and this input has no affect | boolean | false | No |
+| \`\`disabled\`\` | Programmatically disable the button via this property | boolean | false | No |
```
elenagarrone, line 62: I probably missed something but why do we need a `disabled` input when a user can simply add the attribute to the html button element itself?

projects/canopy/src/lib/button/button.stories.ts
```diff
@@ -27,25 +28,29 @@ const contentGroupId = 'content';
...
     <button
       lg-button
-      [variant]="variant"
+      [disabled]="disabled"
+      [fullWidth]="fullWidth"
       [iconButton]="iconButton"
       [iconPosition]="iconPosition"
       [loading]="loading"
-      [fullWidth]="fullWidth"
+      [size]="size"
```
elenagarrone, line 36: I'm wondering if maybe we should have a single `size` input that includes `fullWidth` instead of having them as two separate inputs. They both affect the size of the button so I think it makes sense 🤔

projects/canopy/src/lib/forms/input/input-field.component.ts
```diff
@@ -47,28 +80,108 @@ export class LgInputFieldComponent {
...
+        element,
+      );
+    });
+    this._prefixChildren = elements;
+  }
+  get prefixChildren() {
+    return this._prefixChildren;
+  }
+
+  @Input() id = `lg-input-${nextUniqueId++}`;
```
elenagarrone, line 143: I think it would be better to save `nextUniqueId++` into a separate const and then use it, otherwise if we end up reusing it, it will increase each time.

projects/canopy/src/lib/forms/input/input-field.component.ts
```diff
@@ -47,28 +80,108 @@ export class LgInputFieldComponent {
...
+  /*
+    The input field control element mimics the border of the input field.
+    This allows us to add buttons and icons inside the input field.
+    Lack of IE11 support for :focus-within necessitates us doing this in JS
+  */
+  private hasFocus = false;
+  private hasHover = false;
+  private disabledStateChanges: Subscription;
+
+  ngAfterContentInit() {
```
elenagarrone, line 156: I would like us to start using return types a bit more. Could you please add the return types for these functions?

elenagarrone (changes requested):
Great work! I just had few small comments plus the followings for storybook:

- the `with button`: the button doesn't have text so the screen reader only reads "button"
- the `with suffix`: the example has the percentage but the label is still "Name" which makes no sense ;) this is just me being annoying :D
- the `with prefix`: same as per above comment
- the `with multiple button suffixes`: both the buttons don't have text so nothing is read to the screen readers apart from "button"

package.json
```diff
@@ -135,22 +135,18 @@
   },
   "lint-staged": {
     "projects/canopy/src/**/*.svg": [
-      "npm run build:icons",
-      "git add"
```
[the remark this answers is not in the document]
thatguynamedandy: Ah good spot, it seems to be a mixed bag. The 'ts' files worked and the changes were committed, but the SVGs and SCSS files were updated and not added to the commit. It would be good to fix this at some stage. For now I've removed the commit from this PR.

projects/canopy/src/lib/button/button.component.ts
```diff
@@ -57,8 +61,14 @@ export class LgButtonComponent {
     return this.iconButton;
   }
 
-  constructor(private renderer: Renderer2, private hostElement: ElementRef) {
+  @Input() size: ButtonSize;
+  @HostBinding('class.lg-btn--sm') get sizeClass() {
+    return this.size === 'sm';
+  }
+
+  constructor(private renderer: Renderer2, public hostElement: ElementRef) {
```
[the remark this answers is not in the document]
thatguynamedandy: This one should now be cleared up with this https://github.com/Legal-and-General/canopy/pull/81. The convention being to use 'hostElement', unless the 'ElementRef' is for a different element of course. So in the mainline branch we currently only have 'hostElement'

projects/canopy/src/lib/button/button.stories.ts
owensgit: I guess the downside of doing this is that you may want a button to be both `sm` _and_ `fullWidth`, which would be easy to do with separate inputs. If they were combined, we may need more options such as `sm-full-width`, `md-full-width` and so on, which could get long.

projects/canopy/src/lib/button/button.interface.ts
thatguynamedandy: My reason for this change was the chance of a naming clash in certain composite components. An example scenario could be a component that needs an `AlertVariant` and a `ButtonVariant`. This change does also bring a bit of consistency with the other interfaces which all have prefixes.

projects/canopy/src/lib/button/button.stories.ts
thatguynamedandy: Yeah it might a good idea, it could be linked to this issue.. https://github.com/canopy-collective/canopy/issues/251. I didn't actually change anything around that on this PR, I just put it into alphabetical order. (which I realise doesn't make it easier for the reviewer :-) )

projects/canopy/src/lib/forms/input/input-field.component.ts
thatguynamedandy: This was mostly reformatting and not changed in the PR. I'm open to changing it though but I'm not sure I quite follow. The idea is that it should increase each time, it was a convention I shamefully [borrowed from material ui](https://github.com/angular/components/blob/master/src/material/input/input.ts#L94)

projects/canopy/src/lib/forms/input/input-field.component.ts
thatguynamedandy: Good idea, I've added a bunch to this file, I wonder if there is a lint warning we can add?

projects/canopy/src/lib/button/button.notes.ts
thatguynamedandy: The reason I added it on this is that if you disable the form control and there is a button suffix it will automatically disable the button. It's much easier to do this programmatically via the component instance as opposed to going into the markup to do it. I can see that it does provide two options to do the same thing though, not sure if there is a cleaner alternative.

projects/canopy/src/lib/forms/input/input-field.component.scss
```diff
@@ -1,9 +1,115 @@
 @import '../../../styles/mixins';
-@import './input.scss';
+
+:root {
+  --suffix-margin: var(--space-sm);
+  --suffix-button-margin: 0.3125rem;
+}
 
 .lg-input-field {
+  $block: &;
```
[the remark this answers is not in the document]
thatguynamedandy: I went for block as in the block in**B**EM, but in hindsight that may be misread, did you read it as block like `display:block`?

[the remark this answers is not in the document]
owensgit: Yeah it took me a couple of takes to realise what was going on, maybe renaming it could make the code easier to scan. Then again, it's pretty clear once you know what's going on. So happy either way really :)

owensgit (approved):

thatguynamedandy:
@elenagarrone on these ones...

> the with button: the button doesn't have text so the screen reader only reads "button"
the with suffix: the example has the percentage but the label is still "Name" which makes no sense ;) this is just me being annoying :D
the with prefix: same as per above comment
the with multiple button suffixes: both the buttons don't have text so nothing is read to the screen readers apart from "button"

I've hopefully fixed the first one I was accidentally not rendering the text when it was an icon button. On the other ones I've tweaked the default text for those examples, people could still override them with silly things. I'm hoping by fixing the first one I've also fixed the last one, but there was text in the close button, so I'm slightly surprised that didn't read out, I might need to borrow your screen reader skills.

projects/canopy/src/lib/button/button.interface.ts
elenagarrone: Yes, makes sense 👍

projects/canopy/src/lib/forms/input/input-field.component.ts
elenagarrone: Yes, it should increase each time a new component is instantiated.
I have had to fix multiple instances where we did the following mistake: 
```
@Input() id = `lg-input-${nextUniqueId++}`;
anotherVariable = `my-variable-${nextUniqueId++}`
```
which produces:
```
@Input() id = `lg-input-1`;
anotherVariable = `my-variable-2`
```
while it should be:
```
@Input() id = `lg-input-1`;
anotherVariable = `my-variable-1`
```

I know in this component this is not an issue atm but it did happen multiple times and I'm just concerned that someone might come along and do the same.

projects/canopy/src/lib/forms/input/input-field.component.ts
elenagarrone: Yes I'm sure there is as we have it in our project. Might be one for when we create the global linting?

elenagarrone:
> @elenagarrone on these ones...
> 
> > the with button: the button doesn't have text so the screen reader only reads "button"
> > the with suffix: the example has the percentage but the label is still "Name" which makes no sense ;) this is just me being annoying :D
> > the with prefix: same as per above comment
> > the with multiple button suffixes: both the buttons don't have text so nothing is read to the screen readers apart from "button"
> 
> I've hopefully fixed the first one I was accidentally not rendering the text when it was an icon button. On the other ones I've tweaked the default text for those examples, people could still override them with silly things. I'm hoping by fixing the first one I've also fixed the last one, but there was text in the close button, so I'm slightly surprised that didn't read out, I might need to borrow your screen reader skills.

The text in the first and last button is still not read and the reason is that you've set the `lg-btn-text` to have `visibility: hidden`. Most screen readers will not read elements with `visibility: hidden` or `display: block`. I would suggest you use the mixin `lg-visually-hidden` or the class `.lg-visually-hidden`.

projects/canopy/src/lib/forms/input/input-field.component.ts
thatguynamedandy: ah ok that makes sense, so in theory with the first example it still works as the IDs are unique, but it's harder to ascertain which elements join together as the numerical part of the id is different. (pushed a change)

thatguynamedandy:
https://github.com/Legal-and-General/canopy/pull/77#issuecomment-734750673
Good spot on this @elenagarrone , it was actually a bug on the original icon button implementation, I have tagged a separate commit onto this PR to fix it.

elenagarrone (dismissed):
Great work 👍

elenagarrone (approved):

owensgit (approved):
Great addition Andy! 👍

PR #86, opened 2020-11-27
feat(segment): add segment component

elenagarrone:
# Description

This PR implements the segment component. The way the segment stack is controlled by the `stack` input which by default is set to be falsey.

## Requirements

The segment component should be either inline at all breakpoints or stack whenever needed (sm, md, lg breakpoint).
It takes the full width of the container element.

Storybook link: https://deploy-preview-86--legal-and-general-canopy.netlify.app/
Design link: https://legalandgeneral.invisionapp.com/share/NSZDSDYUWFE#/screens/416751156_Segment
Screenshots:

Screenshot 2020-11-27 at 17 39 02
Screenshot 2020-11-27 at 17 39 21
Screenshot 2020-11-27 at 17 39 43

# Checklist:

- [x] I have [linked the new component](<(https://github.com/Legal-and-General/canopy/blob/master/docs/CONTRIBUTING.md#invision-dsm)>) to adobe DSM (if appropriate)

### DIFF
--- a/projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
+++ b/projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
@@ -1,0 +1,93 @@
+@import '../../../styles/mixins';
+
+.lg-radio-button--segment {
+  display: flex;
+  flex-grow: 1;
+  flex-basis: 0;
+  align-items: stretch;
+  padding: 0 var(--space-xxxs);
+  margin: var(--space-xxxs) 0;
+
+  + .lg-radio-button--segment {
+    border-left: var(--radio-segment-separator-border-width) solid
+      var(--radio-segment-separator-border-color);
+  }
+
+  .lg-radio-button__label {
+    display: flex;
+    padding: var(--space-xxxs) var(--space-sm);
+    border-radius: var(--border-radius-sm);
+    min-width: var(--radio-segment-label-min-width);
+    flex-grow: 1;
+    text-align: center;
+    align-items: center;
+    justify-content: center;
+
+    &::before,
+    &::after {
+      display: none;
+    }
+  }
+
+  .lg-radio-button__input {
+    &:checked + .lg-radio-button__label {
+      background-color: var(--radio-bg-color);
+      color: var(--radio-color);
+      transition: all var(--animation-duration) var(--animation-fn);
+    }
+
+    &:disabled:checked + .lg-radio-button__label {
+      background-color: var(--radio-disabled-color);
+      color: var(--radio-color);
+    }
+
+    &:disabled + .lg-radio-button__label {
+      color: var(--disabled-color);
+      cursor: not-allowed;
+    }
+
+    &:focus + .lg-radio-button__label {
+      @include lg-inner-focus-outline(var(--default-focus-color));
+    }
+
+    .lg-radio-button--error &:checked + .lg-radio-button__label {
+      background-color: var(--form-error-color);
+      color: var(--radio-color);
+    }
+  }
+
+  $stack-breakpoints: 'sm', 'md', 'lg' !default;
+
+  @each $stack-breakpoint in $stack-breakpoints {
+    &.lg-radio-button--stacked-#{$stack-breakpoint} {
+      @include lg-breakpoint($stack-breakpoint, 'max-width') {
+        display: flex;
+        flex-flow: column wrap;
+        margin: 0;
+      }
+
+      .lg-radio-button__label {
+        @include lg-breakpoint($stack-breakpoint, 'max-width') {
+          margin: calc(var(--space-xxxs) + var(--radio-segment-separator-border-width)) 0
+            var(--space-xxxs);
+        }
+      }
+
+      + .lg-radio-button--segment {
+        @include lg-breakpoint($stack-breakpoint, 'max-width') {
+          border-left: 0;
+        }
+
+        &::before {
+          @include lg-breakpoint($stack-breakpoint, 'max-width') {
+            content: '';
+            display: inline-block;
+            width: 100%;
+            height: var(--radio-segment-separator-border-width);
+            background-color: var(--radio-segment-separator-border-color);
+          }
+        }
+      }
+    }
+  }
+}
--- a/projects/canopy/src/lib/forms/radio/radio-group--segment.component.scss
+++ b/projects/canopy/src/lib/forms/radio/radio-group--segment.component.scss
@@ -1,0 +1,17 @@
+@import '../../../styles/mixins';
+
+.lg-radio-group__segment {
+  display: inline-flex;
+  border: var(--border-width) solid var(--radio-border-color);
+  border-radius: var(--border-radius-sm);
+}
+
+$stack-breakpoints: 'sm', 'md', 'lg' !default;
+
+@each $stack-breakpoint in $stack-breakpoints {
+  .lg-radio-group--stack-#{$stack-breakpoint} .lg-radio-group__segment {
+    @include lg-breakpoint($stack-breakpoint, 'max-width') {
+      display: block;
+    }
+  }
+}
--- a/projects/canopy/src/lib/forms/radio/radio-group.component.html
+++ b/projects/canopy/src/lib/forms/radio/radio-group.component.html
@@ -3,7 +3,15 @@
     <ng-content></ng-content>
   </legend>
   <ng-content select="lg-hint"></ng-content>
-  <ng-content select="lg-radio-button, lg-filter-button"></ng-content>
+
+  <div *ngIf="variant === 'segment'; else generic" class="lg-radio-group__segment">
+    <ng-content select="lg-segment-button"></ng-content>
+  </div>
+
+  <ng-template #generic>
+    <ng-content select="lg-radio-button, lg-filter-button"></ng-content>
+  </ng-template>
+
   <ng-content select="lg-validation"></ng-content>
 </fieldset>
 <ng-content select="lg-hint"></ng-content>
--- a/projects/canopy/src/lib/forms/radio/radio.interface.ts
+++ b/projects/canopy/src/lib/forms/radio/radio.interface.ts
@@ -1,1 +1,2 @@
-export type RadioVariant = 'radio' | 'filter';
+export type RadioVariant = 'radio' | 'filter' | 'segment';
+export type RadioStackBreakpoint = 'sm' | 'md' | 'lg';
--- a/projects/canopy/src/lib/forms/radio/radio.notes.ts
+++ b/projects/canopy/src/lib/forms/radio/radio.notes.ts
@@ -17,13 +17,31 @@ Import the Radio Module into your application:
 and in your HTML:
 
 ~~~html
-<lg-${name.toLowerCase()}-group [inline]="true" formControlName="color">
+<lg-${name.toLowerCase()}-group ${
+  name === 'Radio' ? '[inline]="true"' : ''
+} formControlName="color">
   Colour
   <lg-hint>Please select a colour</lg-hint>
   <lg-${name.toLowerCase()}-button value="red">Red</lg-${name.toLowerCase()}-button>
   <lg-${name.toLowerCase()}-button value="yellow">Yellow</lg-${name.toLowerCase()}-button>
 </lg-${name.toLowerCase()}-group>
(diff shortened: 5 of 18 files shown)


projects/canopy/src/lib/forms/radio/radio.interface.ts
```diff
@@ -1 +1,2 @@
-export type RadioVariant = 'radio' | 'filter';
+export type RadioVariant = 'radio' | 'filter' | 'segment';
+export type RadioStack = 'sm' | 'md' | 'lg';
```
owensgit, line 2: This is great 👍   Just tested it out in Storybook. The only thing is that first I found the name a bit confusing at first, and had to get my head around the code before I understood exactly what it does. I wonder if we could rename this to `RadioStackBreakpoint`?

Also, I think this is only used for the Segment right? Should we consider including that in the name too? Maybe `RadioSegmentStackBreakpoint`. A bit long I suppose 🤔 

Also I think it would be a good idea to include a short paragraph in the notes about how the stacking works. 

Maybe at some point we should consider separate note files for each of the radio button styles, so we can be more description on their specific features and use-cases.

owensgit (commented):
Looks great Elena, thanks for adding this. Just one comment / question from me.

projects/canopy/src/lib/forms/radio/radio.interface.ts
elenagarrone: Maybe we can just call it `RadioBreakpoint`? 
The idea is to expand this to all the radio and checkbox variants and get rid of the `inline` input.

Regarding the additional paragraph in the notes, have you seen that I've included the explanation in the `LgRadioGroupComponent` inputs? do you think it needs a bit more text?

thatguynamedandy: I'm inclined to agree with the renaming of this one, `RadioStackBreakpooint` was the first thing that jumped to my mind, it's a bit long but descriptive enough.

projects/canopy/src/lib/forms/radio/radio.notes.ts
```diff
@@ -33,7 +35,13 @@ and in your HTML:
 | \`\`id\`\` | HTML ID attribute, auto generated if not provided | string | 'lg-radio-group-id-\${nextUniqueId++}' | No |
 | \`\`name\`\` | Set the name value for all inputs in the group, auto-generated if not provided | string | 'lg-radio-group-\${nextUniqueId++}' | No |
 | \`\`value\`\` | Set the default checked radio button, must match the value of the radio button | string | null | No |
-| \`\`inline\`\` | If true, displays the radio buttons inline rather than stacked | boolean | false | No |
+${
+  name === 'Radio'
+    ? `| \`\`inline\`\` | If true, displays the radio buttons inline rather than stacked | boolean | false | No |`
+    : name === 'Segment'
+    ? `| \`\`stack\`\` | Stack the radio buttons at a given breakpoint | 'sm', 'md', 'lg' | undefined | No |`
```
thatguynamedandy, line 42: I think we have a 'false' option here as well, I was wondering if we should set the default to `false` rather than undefined?

thatguynamedandy:
I noticed this, which I think we know about as the emphasis is on the developer to set it up correctly, should we drop something in the notes?

Screenshot 2020-12-03 at 16 25 06

projects/canopy/src/lib/forms/radio/radio.interface.ts
thatguynamedandy: I think in terms of the text we should add a bit more, maybe linked to this https://github.com/Legal-and-General/canopy/pull/86#issuecomment-738121194

owensgit: I quite like `RadioStackBreakpooint` too, it makes it very clear 👍 

In terms of the notes, maybe we could add a small paragraph in the "Purpose" section, something along the lines of..

```
Radios are displayed inline, unless given a `RadioStackBreakpoint`, which tells them at which breakpoint should they appear stacked on top of each other (currently only implemented for the Segment component).
```

What do you think?

elenagarrone: Cool, I'll rename the type to `RadioStackBreakpooint` 👍 

Regarding the notes, I haven't made the `stack` input visible to anything other than the segment's notes because we don't want people to know about it and use it incorrectly in the other variants. I will add a new section to the segments notes to explain that functionality a bit better as you suggested 👍

projects/canopy/src/styles/variables/components/_radio.scss
```diff
@@ -0,0 +1,15 @@
...
+  --radio-color: var(--color-white);
+  --radio-border-color: var(--border-color);
+  --radio-disabled-color: var(--color-taupe-grey);
+
+  --radio-inner-height: 0.5rem;
+  --radio-inner-width: var(--radio-inner-height);
+  --radio-outer-height: 1rem;
+  --radio-outer-width: var(--radio-outer-height);
+
+  --radio-segment-separator-border-width: calc(var(--border-width) / 10);
```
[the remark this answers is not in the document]
elenagarrone: Why have I done this?! 🤔  I'll have a look at it but I would expect it to be `var(--border-width)` 😅

projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
```diff
@@ -0,0 +1,93 @@
...
+    border-left: var(--radio-segment-separator-border-width) solid
+      var(--radio-segment-separator-border-color);
+  }
+
+  .lg-radio-button__label {
+    display: flex;
+    padding: var(--space-xxxs) var(--space-sm);
+    border-radius: var(--border-radius-sm);
+    min-width: var(--radio-segment-label-min-width);
+    width: 100%;
```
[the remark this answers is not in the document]
elenagarrone: I tried that before and it wasn't working, the text doesn't wrap automatically when we reach the `min-width`:
Screenshot 2020-12-04 at 11 54 28

I also tried to set `flex-basis` instead of `min-width` but that causes issues with the length of the segment when we set `inline-flex`:
Screenshot 2020-12-04 at 11 58 52

I might be missing something ;)

projects/canopy/src/lib/forms/radio/radio.interface.ts
elenagarrone: Updated.

projects/canopy/src/lib/forms/radio/radio.notes.ts
elenagarrone: Updated.

projects/canopy/src/lib/forms/radio/radio-group--segment.component.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../../styles/mixins';
+
+.lg-radio-segment {
```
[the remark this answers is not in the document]
elenagarrone: Updated.

projects/canopy/src/styles/variables/components/_radio.scss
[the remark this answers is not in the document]
elenagarrone: Updated.

projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
```diff
@@ -0,0 +1,93 @@
+@import '../../../styles/mixins';
+
+.lg-radio-button--segment {
+  display: flex;
```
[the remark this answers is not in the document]
elenagarrone: Alex and I discussed the fact that the segment could take the full width width of the container - I think it was to make development easier to be honest as I was unsure of how much effort it would take (very little turned out lol).
I think this is better.

Regarding the second point in the design the labels are all the same width: https://legalandgeneral.invisionapp.com/share/NSZDSDYUWFE#/screens/416751156_Segment

@alexcanning could you confirm this please?

projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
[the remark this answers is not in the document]
elenagarrone: Updated to use `flex-grow: 1`.

owensgit (commented):
Looking good from my point of view, although I'm wondering if there should be a default stack behaviour on a small mobile viewport? Just to echo @thatguynamedandy comment above, here's a screenshot of the segments with `stack` set to `false`. I wonder if we should always stack them on this sized viewport, or if there could be a scenario where you'd want them inline here -- perhaps if there is just two small, one-word segments?

Screenshot 2020-12-10 at 09 23 33

elenagarrone:
> Looking good from my point of view, although I'm wondering if there should be a default stack behaviour on a small mobile viewport? Just to echo @thatguynamedandy comment above, here's a screenshot of the segments with `stack` set to `false`. I wonder if we should always stack them on this sized viewport, or if there could be a scenario where you'd want them inline here -- perhaps if there is just two small, one-word segments?
> 
> CC @alexcanning
> 
> Screenshot 2020-12-10 at 09 23 33

The preference was to keep them inline because if the options fit in the viewport they should be displayed like that but maybe we could set the default to be stacked at 'sm' breakpoint instead of inline? I guess it depends on the number of items we are expecting to see the most in the segment. Any idea @alexcanning?

projects/canopy/src/lib/forms/radio/radio-button--segment.component.scss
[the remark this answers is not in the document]
alexcanning: Yes this is correct - I think the rationale is that we'd always want it to be the full width of whatever container it was in. If you're designing a form, the whole form contents are likely to be constrained to a given number of columns, no?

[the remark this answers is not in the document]
alexcanning: I guess I should also say the the maximum width should be 100% because, say for instance your options were Yes / No, you wouldn't want this to expand to the full width

alexcanning:
> The preference was to keep them inline because if the options fit in the viewport they should be displayed like that but maybe we could set the default to be stacked at 'sm' breakpoint instead of inline? I guess it depends on the number of items we are expecting to see the most in the segment. Any idea @alexcanning?

The preference would always be for horizontal version to be the default, with the designers and developers consciously identifying circumstances where the stacked variant is required.

alexcanning:
Looks great overall! The only thing I noticed is that the instance on Storybook seems to have an initial option set. One of the main purposes of this control when compared with a radio group is that the segment control should not have an initial value set. This is to allow for situations when we need the user to make a conscious choice between several options.

thatguynamedandy:
> Looks great overall! The only thing I noticed is that the instance on Storybook seems to have an initial option set. One of the main purposes of this control when compared with a radio group is that the segment control should not have an initial value set. This is to allow for situations when we need the user to make a conscious choice between several options.

This is an interesting one, as I always assumed that setting an initial value on a radio button is wrong, but do we have a slightly different need? drawing comparison from GOV.uk

https://design-system.service.gov.uk/components/radios/
```
Do not pre-select radio options as this makes it more likely that users will:

- not realise they’ve missed a question
- submit the wrong answer
```

alexcanning:
A few sources recommend always having a radio button selected by default, e.g.:

https://www.nngroup.com/articles/radio-buttons-default-selection/

Although some of the thinking on why you should do this seems to differ.

We should default radio groups to their safest choice, or perhaps providing confirmation as part of our journeys.

elenagarrone:
@alexcanning @thatguynamedandy I've updated both the story for the segment and the test app example to not have any initial value.

thatguynamedandy (dismissed):

projects/canopy/src/lib/forms/radio/radio-group.component.html
```diff
@@ -3,7 +3,15 @@
     <ng-content></ng-content>
   </legend>
   <ng-content select="lg-hint"></ng-content>
-  <ng-content select="lg-radio-button, lg-filter-button"></ng-content>
+
+  <div *ngIf="variant === 'segment'" class="lg-radio-group__segment">
```
paul-request, line 7: Not a blocker but may e the `if else` syntax is better here?

paul-request (approved):

thatguynamedandy (approved):

paul-request (approved):

PR #105, opened 2020-12-08
fix(forms): update checkbox and radio template

elenagarrone:
# Description

The checkbox and radio template were setting the `disabled` and `checked` attributes
with `attr.checkbox` and `attr.disabled`. `attr` sets a "custom" attribute called `checked` and `disabled` which don't match the actual native implementation. 
This bug is not noticeable unless you handle those attributes dynamically and they get out of sync. 

Below an example of the issue I was seeing where I was passing the checked input from the `lg-toggle` component but the styles were still picking up the `:checked` pseudo selector:

Screenshot 2020-12-08 at 12 02 21

# Checklist:


projects/canopy/src/lib/forms/toggle/toggle.stories.common.ts
```diff
@@ -54,11 +64,13 @@ export const createToggleStory = (variant: string) => ({
...
     [label]="label"
     variant="${variant}"
+    [checked]="checked"
     (toggleChange)="toggleChange($event)">
   </lg-reactive-form>`,
   props: {
     toggleChange: action('toggleChange'),
     label: text('label', 'I will bring my Umbrella if it is raining'),
     disabled: boolean('disabled', false),
+    checked: boolean('reactivily checked', false),
```
pauleustice, line 74: Samll tpyo ehre

pauleustice (approved):

thatguynamedandy (dismissed):

paul-request (dismissed):

projects/canopy/src/lib/forms/toggle/toggle.stories.common.ts
```diff
@@ -46,6 +52,10 @@ export class ReactiveToggleFormComponent {
     });
     this.form.valueChanges.subscribe((val) => this.toggleChange.emit(val));
   }
+
+  ngOnChanges(changes): any {
```
aaronBery, line 56: do we need a return type here? can it be implicitly void

elenagarrone: The `any` was a mistake but I've added `void` instead as I think we should start to be more explicit with typings.

paul-request (approved):

owensgit (approved):

Issue #70, opened 2020-11-19
Suppressing component margin on buttons (and maybe other compoments)


andy-polhill:
By default we add a bottom margin to a lot of components. This gives them the correct vertical spacing and means they can be dropped into a component without any additional work, it also keeps the vertical rhythm consistent. A good example of this is form fields in a form.

However as @p-m-p pointed out this can be pain in certain scenarios, a particular example: when using buttons in a non vertical layout the margin needs to be individually over written each time a button is used.

The [lgMargin directive](https://legal-and-general-canopy.netlify.app/?path=/story/directives--margin) helps to do this on the odd occasions, but a facility to do it globally would help. One option could be to override the css variables, currently we have one called `--component-margin` which controls the vertical spacing. We could extend this to be more granular and have `--button-margin` which maps to `--component-margin` . So you could then override either the button variable to control the buttons or the global one to control all component margins. I'm not sure the best way of doing the overrides, e.g. could it be done [programatically](https://css-tricks.com/updating-a-css-variable-with-javascript/) in someway without any visual quirks.

Xinchro:
Closing this as not an issue. Won't implement.

PR #107, opened 2020-12-09
Sort code field

owensgit:
# Description

Creates a new Sort Code component, based on the current Date Field component.

Also fixes a couple of issues found with the Date Field (and also Sort Code) component (see other commits).

## Requirements

Please test the component as normal and also validation rules as part of a form, for example the Validation Form story.

Screenshot:
Screenshot 2020-12-08 at 16 29 07

# Checklist:

### DIFF
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -409,6 +409,11 @@
               <lg-filter-checkbox value="blue">Blue</lg-filter-checkbox>
             </lg-filter-multiple-group>
 
+            <lg-sort-code formControlName="sortCode">
+              Sort Code
+              <lg-hint>Must be 6 digits long</lg-hint>
+            </lg-sort-code>
+
             <button lg-button type="submit" variant="solid-primary">Submit</button>
           </form>
 
--- a/projects/canopy/src/lib/forms/date/date-field.component.html
+++ b/projects/canopy/src/lib/forms/date/date-field.component.html
@@ -10,7 +10,7 @@
   >
     <div class="lg-date-field__date">
       <label lg-label [for]="dateId">Day</label>
-      <lg-input-field>
+      <lg-input-field lgMarginBottom="none">
         <input
           lgInput
           [id]="dateId"
@@ -25,9 +25,10 @@
         />
       </lg-input-field>
     </div>
+
     <div class="lg-date-field__month">
       <label lg-label [for]="monthId">Month</label>
-      <lg-input-field>
+      <lg-input-field lgMarginBottom="none">
         <input
           lgInput
           [id]="monthId"
@@ -42,9 +43,10 @@
         />
       </lg-input-field>
     </div>
+
     <div class="lg-date-field__year">
       <label lg-label [for]="yearId">Year</label>
-      <lg-input-field>
+      <lg-input-field lgMarginBottom="none">
         <input
           lgInput
           [id]="yearId"
--- a/projects/canopy/src/lib/forms/input/input-field.component.ts
+++ b/projects/canopy/src/lib/forms/input/input-field.component.ts
@@ -9,7 +9,9 @@ import {
   ContentChildren,
   QueryList,
   OnDestroy,
+  Optional,
 } from '@angular/core';
+import { FormGroupDirective } from '@angular/forms';
 
 import { Subscription } from 'rxjs';
 
@@ -21,6 +23,7 @@ import { LgInputDirective } from './input.directive';
 import { LgButtonComponent } from '../../button';
 import { LgSuffixDirective } from '../../suffix/suffix.directive';
 import { LgPrefixDirective } from '../../prefix/prefix.directive';
+import { LgErrorStateMatcher } from '../validation/error-state-matcher';
 
 let nextUniqueId = 0;
 
@@ -52,8 +55,11 @@ export class LgInputFieldComponent implements AfterContentInit, OnDestroy {
   }
 
   @HostBinding('class.lg-input-field--error')
-  get errorClass(): LgValidationComponent {
-    return this._validationElement;
+  get errorClass(): boolean {
+    return this.errorState.isControlInvalid(
+      this._inputElement.control,
+      this.controlContainer,
+    );
   }
 
   @HostBinding('class.lg-input-field--block')
@@ -144,7 +150,12 @@ export class LgInputFieldComponent implements AfterContentInit, OnDestroy {
 
   @Input() id = `lg-input-${this._id++}`;
 
-  constructor(private domService: LgDomService) {}
+  constructor(
+    private domService: LgDomService,
+    private errorState: LgErrorStateMatcher,
+    @Optional()
+    private controlContainer: FormGroupDirective,
+  ) {}
 
   /*
     The input field control element mimics the border of the input field.
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.html
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.html
@@ -1,0 +1,73 @@
+<fieldset [attr.aria-describedby]="ariaDescribedBy">
+  <legend lg-label [id]="labelId">
+    <ng-content></ng-content>
+  </legend>
+
+  <ng-content select="lg-hint"></ng-content>
+  <div
+    class="lg-sort-code__fields"
+    [formGroup]="sortCodeFormGroup"
+    #sortCodeFormDirective="ngForm"
+  >
+    <div class="lg-sort-code__input">
+      <label lg-label class="lg-visually-hidden" [for]="firstId"
+        >First two digits of your sort code</label
+      >
+
+      <lg-input-field lgMarginBottom="none">
+        <input
+          lgInput
+          [id]="firstId"
+          (blur)="onBlur()"
+          (input)="first.setValue($event.target.value)"
+          formControlName="first"
+          inputmode="numeric"
+          maxlength="2"
+          placeholder="00"
+          size="2"
+          type="text"
+        />
+      </lg-input-field>
+    </div>
+    <div class="lg-sort-code__input">
+      <label lg-label class="lg-visually-hidden" [for]="secondId"
+        >Second two digits of your sort code</label
+      >
+      <lg-input-field lgMarginBottom="none">
+        <input
+          lgInput
+          [id]="secondId"
+          (blur)="onBlur()"
+          (input)="second.setValue($event.target.value)"
+          formControlName="second"
+          inputmode="numeric"
+          maxlength="2"
+          placeholder="00"
+          size="2"
+          type="text"
+        />
+      </lg-input-field>
+    </div>
+    <div class="lg-sort-code__input">
+      <label lg-label class="lg-visually-hidden" [for]="thirdId"
+        >Third two digits of your sort code</label
+      >
+      <lg-input-field lgMarginBottom="none">
+        <input
+          lgInput
+          [id]="thirdId"
+          (blur)="onBlur()"
+          (input)="third.setValue($event.target.value)"
+          formControlName="third"
+          inputmode="numeric"
+          maxlength="2"
+          placeholder="00"
+          size="2"
+          type="text"
+        />
+      </lg-input-field>
+    </div>
+  </div>
+
+  <ng-content select="lg-validation"></ng-content>
+</fieldset>
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
@@ -1,0 +1,15 @@
+.lg-sort-code-field {
+  display: block;
+  margin-bottom: var(--component-margin);
+  line-height: var(--line-height-sm);
+}
+
+.lg-sort-code__fields {
+  display: flex;
+}
+
+.lg-sort-code__input {
+  flex-direction: column;
+  margin-right: var(--space-sm);
+  width: 3.75rem; // 60px
+}
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
@@ -25,6 +25,7 @@ import { LgErrorStateMatcher } from '../validation/error-state-matcher';
 import { LgValidationComponent } from '../validation/validation.component';
 import { LgSortCodeComponent } from './sort-code.component';
 import { LgInputModule } from '../input';
+import { LgButtonComponent } from '../../button/button.component';
 
 const errorId = 'test-error-id';
 const hintId = 'test-hint-id';
@@ -33,12 +34,13 @@ const errorStateMatcherMock = mock(LgErrorStateMatcher);
 
 @Component({
   template: `
-    <form (ngSubmit)="submit()" [formGroup]="form" #testForm="ngForm">
+    <form [formGroup]="form" #testForm="ngForm">
       <lg-sort-code formControlName="sortCode">
         Sort Code
         <lg-hint id="${hintId}">Must be 6 digits long</lg-hint>
         <lg-validation id="${errorId}">Error</lg-validation>
       </lg-sort-code>
+      <button lg-button type="submit" variant="solid-primary">Submit</button>
     </form>
   `,
 })
@@ -91,7 +93,7 @@ describe('SortCodeComponent', () => {
         declarations: [
           TestSortCodeComponent,
           LgSortCodeComponent,
-          MockComponents(LgHintComponent, LgValidationComponent),
+          MockComponents(LgHintComponent, LgValidationComponent, LgButtonComponent),
         ],
         providers: [
           {
@@ -293,6 +295,27 @@ describe('SortCodeComponent', () => {
       expect(fixture.componentInstance.form.controls.sortCode.errors).toEqual({
         invalidField: true,
       });
+    });
+  });
+
+  describe('when parent form is submitted', () => {
+    let sortCodeComponent: LgSortCodeComponent;
+
+    beforeEach(() => {
+      sortCodeDebugElement = fixture.debugElement.query(
+        By.directive(LgSortCodeComponent),
+      );
+      sortCodeComponent = sortCodeDebugElement.componentInstance;
+    });
+
+    it('should submit the form group', () => {
+      const spy = spyOn(sortCodeComponent.formGroupDirective, 'onSubmit');
+      const buttonDebugElement = fixture.debugElement.query(
+        By.directive(LgButtonComponent),
+      );
+      buttonDebugElement.nativeElement.click();
+
+      expect(spy).toHaveBeenCalled();
     });
   });
 });
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
@@ -1,0 +1,298 @@
+import {
+  Component,
+  DebugElement,
+  EventEmitter,
+  Input,
+  Output,
+  ViewChild,
+} from '@angular/core';
+import { waitForAsync, ComponentFixture, TestBed } from '@angular/core/testing';
+import {
+  FormBuilder,
+  FormGroup,
+  FormGroupDirective,
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { By } from '@angular/platform-browser';
+
+import { MockComponents } from 'ng-mocks';
+import { instance, mock } from 'ts-mockito';
+import { skip } from 'rxjs/operators';
+
+import { LgHintComponent } from '../hint/hint.component';
+import { LgErrorStateMatcher } from '../validation/error-state-matcher';
+import { LgValidationComponent } from '../validation/validation.component';
+import { LgSortCodeComponent } from './sort-code.component';
+import { LgInputModule } from '../input';
+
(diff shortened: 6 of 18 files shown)


projects/canopy/src/lib/forms/date/date-field.component.html
```diff
@@ -10,7 +10,7 @@
   >
     <div class="lg-date-field__date">
       <label lg-label [for]="dateId">Day</label>
-      <lg-input-field>
+      <lg-input-field lgMarginBottom="none">
```
elenagarrone, line 13: I'm wondering if maybe the input shouldn't have margin by default? 🤔  Not specific to this PR but something to consider 🤔  cc @thatguynamedandy

projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
```diff
@@ -0,0 +1,15 @@
...
+}
+
+.lg-sort-code__fields {
+  display: flex;
+}
+
+.lg-sort-code__input {
+  flex-direction: column;
+  margin-right: var(--space-sm);
+  width: 3.75rem; // 60px
```
elenagarrone, line 14: I'm wondering if we should introduce a `mixin` to translate `px` into `rem`. Maybe it would make it easier to understand without having to add the comment of the corresponding `px` value?

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    expect(/lg-input-sort-code-second-\d{1,3}/.test(secondInput.nativeElement.id)).toBe(
+      true,
+    );
+    expect(/lg-input-sort-code-third-\d{1,3}/.test(thirdInput.nativeElement.id)).toBe(
+      true,
+    );
+  });
+
+  it('replaces empty fields with an empty string if sort code is not complete', (done) => {
+    fixture.detectChanges();
```
elenagarrone, line 205: I think you could move all these `fixture.detectChanges()` in a before each.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -0,0 +1,206 @@
...
+  sortCodeFormGroup: FormGroup;
+  first: FormControl;
+  second: FormControl;
+  third: FormControl;
+  valueChanges: Subscription;
+  hasError: boolean;
+
+  @Input() value: string;
+  @Input() disabled: false;
+  @Input() firstId = `lg-input-sort-code-first-${this.uniqueId++}`;
```
elenagarrone, line 52: Careful here with `++`, I think what you really want is:
```
@Input() firstId = `lg-input-sort-code-first-${this.uniqueId}`;
@Input() secondId = `lg-input-sort-code-second-${this.uniqueId}`;
@Input() thirdId = `lg-input-sort-code-third-${this.uniqueId}`;
``

File: projects/canopy/src/lib/forms/sort-code/sort-code.component.ts [hunk 5]
```diff
@@ -0,0 +1,206 @@
...
+    // submit the group when the parent form is submitted
+    this.parentFormGroupDirective.ngSubmit
+      .pipe(filter(({ type }) => type === 'submit'))
+      .subscribe((event) => {
+        this.formGroupDirective.onSubmit(event);
+        this.ngControl.control.updateValueAndValidity();
+      });
+  }
+
+  isControlInvalid(control: NgControl, form: FormGroupDirective) {
```
elenagarrone, line 142: This should probably have a return type of `boolean` (I think).

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -0,0 +1,206 @@
...
+        errors.invalidField = true;
+      }
+    });
+    if (Object.keys(errors).length) {
+      return errors;
+    }
+    return null;
+  }
+
+  onBlur() {
```
elenagarrone, line 175: Could you add the return type please?

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
```diff
@@ -0,0 +1,73 @@
...
+    class="lg-sort-code__fields"
+    [formGroup]="sortCodeFormGroup"
+    #sortCodeFormDirective="ngForm"
+  >
+    <div class="lg-sort-code__input">
+      <label lg-label class="lg-visually-hidden" [for]="firstId"
+        >First two digits of your sort code</label
+      >
+
+      <lg-input-field lgMarginBottom="none">
```
aaronBery, line 17: I was wondering whether there is too much repetition going on here and whether we could loop over a set of fields...I think I'm probably leaning towards this implementation though out of fear of it becoming over engineered. be good to hear what others think though...

projects/canopy/src/lib/forms/sort-code/sort-code.notes.ts
```diff
@@ -0,0 +1,45 @@
...
+~~~
+
+## Inputs
+
+### LgSortCodeComponent
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| \`\`value\`\` | Existing 6 digit sort code string that will prepopulate the three input fields appropriately | string | '' | No |
+| \`\`disabled\`\` | Sets the three inner input fields to disbaled if \`\`true\`\` | boolean | \`\`false\`\` | No |
+| \`\`firstId\`\` | HTML ID attribute for the first number input, auto generated if not provided | string | \`\`lg-input-date-\${nextUniqueId++}\`\` | No |
```
elenagarrone, line 36: `nextUniqueId++` is incorrect here I think - it should be  just `id`?

elenagarrone (changes requested):
Great work @owensgit, just few small comments 👍

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -0,0 +1,206 @@
...
+  hasError: boolean;
+
+  @Input() value: string;
+  @Input() disabled: false;
+  @Input() firstId = `lg-input-sort-code-first-${this.uniqueId++}`;
+  @Input() secondId = `lg-input-sort-code-second-${this.uniqueId++}`;
+  @Input() thirdId = `lg-input-sort-code-third-${this.uniqueId++}`;
+  @Input() ariaDescribedBy: string;
+
+  _hintElement: LgHintComponent;
```
aaronBery, line 57: could this along with the `_validationElement` be marked as private?  at that point to be inline with our getter setter pattern do we then want a 

```
get hintElement() {
  return this._hintElement;
}
```

projects/canopy/src/lib/forms/date/date-field.component.html
owensgit: I always assumed that the input field needed margin because more often than not, you'd be adding them to a form stacked on top of each other, so having to add the correct margin to each one would be cumbersome and prone to error. But that's just been my assumption, rather than really considering it. What's the rationale for removing it?

projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
owensgit: That could be quite nice, it also saves you from having to do the calculation each time. I'm happy to add one in this PR and use it in this instance, and we could have an issue open to clean up the rest of the values (which I'd be happy to do also 🙂 ). I'll check with the others in Slack using poll to see if the majority agrees.

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
owensgit: Yes, good point 👍

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Ah yes this has come up a few times, I totally forgot 🤦

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+  @Output() sortCodeChange: EventEmitter<{
+    sortCode: string;
+  }> = new EventEmitter();
+
+  form: FormGroup;
+
+  @ViewChild('testForm')
+  testFormDirective: FormGroupDirective;
+
+  constructor(public fb: FormBuilder) {
```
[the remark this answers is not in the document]
owensgit: Hmm yeah maybe not, I'll check it out

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
owensgit: Not 100% sure about this. It's an approach copied from the Date Field component, so I assumed there was a reason that I hadn't thought of. I think I'd prefer to keep it consistent and raise an issue to remove if everyone thinks they're superfluous .

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Yes I think so too

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Ah yes, will do 👍

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
owensgit: I'm more included to keep it separately implemented fields, rather than a loop, but happy to change it if there's a strong consensus to.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Ah yes, I think it could 👍

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (this.sortCodeFormGroup.touched) {
+      this.onTouched();
+    }
+  }
+
+  public onChange(sortCode: string) {
+    this.value = sortCode;
+  }
+
+  public onTouched(_?: any) {}
```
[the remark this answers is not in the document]
owensgit: Hmm good question. I'm not 100% familiar with the `ControlValueAccessor`, so I don't what the best practice is. From what I can see after a quick search, there are examples out there of people doing both public and private, but I can't find anything explaining which is preferred. Does anyone else know?

projects/canopy/src/lib/forms/sort-code/sort-code.notes.ts
owensgit: I think `nextUniqueId++` is right, as that's the name of the variable that generates it. Unless I'm missing something? I actually also just removed the `++` too, as per your other comment 🙂

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
owensgit: I may have misread the question, but I was reading it as; should this method be public or private 🙂

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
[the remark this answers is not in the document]
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+      this.sortCodeChange.emit(val);
+    });
+  }
+}
+
+describe('SortCodeComponent', () => {
+  let fixture: ComponentFixture<TestSortCodeComponent>;
+  let component: TestSortCodeComponent;
+  let sortCodeFieldInstance: LgSortCodeComponent;
+  let sorCodeDebugElement: DebugElement;
```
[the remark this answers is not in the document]
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
owensgit: Done ✅

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
aaronBery: sorry badly worded by me. I was suggesting we omit the public keyword as we generally do in the codebase (but we do have a few inconsistencies that need cleaning up)

projects/canopy/src/lib/forms/date/date-field.component.html
thatguynamedandy: @owensgit is right with the intention, but there is also a downside. The hope was that the majority of time it would be easier as the default margin is in place. Links a bit to this https://github.com/Legal-and-General/canopy/issues/70

projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
thatguynamedandy: I need to have a think about this, I guess the designs are in pixels and thats where the pain comes from. Is the real fix to get the designs also done in rem/em. Or just switch to pixels, I think it needs a bit of reading up.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
thatguynamedandy: they were left that way so that they could be overwritten if there was a need (copied from material ui). It seems unlikely but there could be a scenario where you want control over the id, for example interacting with a third party piece of code outside of Angular.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
thatguynamedandy: I think we are probably a bit inconsistent with our use of `public` / `private` like you say, I wonder if there is a way of enforcing more of a convention.

projects/canopy/src/lib/forms/sort-code/sort-code.notes.ts
thatguynamedandy: I think we've used that convention elsewhere, just to imply that a default is provided, but it's not a static default.

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
thatguynamedandy: I think I'm ok with this for now, just in terms of simplicity

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
thatguynamedandy: would this enforce us to add return types?
https://stackoverflow.com/questions/42793701/is-there-a-way-to-enforce-method-return-types-on-typescript-classes-via-a-tslint

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
```diff
@@ -0,0 +1,73 @@
...
+  </legend>
+
+  <ng-content select="lg-hint"></ng-content>
+  <div
+    class="lg-sort-code__fields"
+    [formGroup]="sortCodeFormGroup"
+    #sortCodeFormDirective="ngForm"
+  >
+    <div class="lg-sort-code__input">
+      <label lg-label class="lg-visually-hidden" [for]="firstId"
```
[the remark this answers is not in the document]
owensgit: Yes good point. From my own testing, it seems to work well with Voice Over. I'll make sure our tester goes through it properly 👍

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -409,6 +409,11 @@
               <lg-filter-checkbox value="blue">Blue</lg-filter-checkbox>
             </lg-filter-multiple-group>
 
+            <lg-sort-code formControlName="sortCode" required>
```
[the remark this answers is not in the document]
owensgit: Ah good spot, this should only be for template driven forms. This was part of a little test I was doing previously, but forgot to remove it

projects/canopy/src/lib/canopy.module.ts
```diff
@@ -52,6 +53,7 @@ const modules = [
   LgPrefixModule,
   LgQuickActionModule,
   LgSeparatorModule,
+  LgSortCodeModule,
```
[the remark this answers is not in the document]
owensgit: Ah yes, remove this ✅

projects/canopy/src/lib/forms/input/input-field.component.ts
```diff
@@ -52,8 +55,11 @@ export class LgInputFieldComponent implements AfterContentInit, OnDestroy {
   }
 
   @HostBinding('class.lg-input-field--error')
-  get errorClass(): LgValidationComponent {
-    return this._validationElement;
+  get errorClass(): boolean {
```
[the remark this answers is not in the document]
owensgit: Tests added ✅ 
https://github.com/Legal-and-General/canopy/pull/107/files#diff-55317529bf721b2597b6815ba808b632246a35375cad53e658d7c58f186e1e00

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
```diff
@@ -0,0 +1,73 @@
+<fieldset [attr.aria-describedby]="ariaDescribedBy">
+  <legend lg-label id="sortCodeLabel">
```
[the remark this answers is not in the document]
owensgit: Ah yes good point, I've made this unique now ✅ 

Also made it an input to make it consistent with the other IDs

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.valueChanges = this.sortCodeFormGroup.valueChanges.subscribe(
+      (sortCode: SortCodeField) => {
+        const first = sortCode.first || '';
+        const second = sortCode.second || '';
+        const third = sortCode.third || '';
+        this.onChange(`${first}${second}${third}`);
+      },
+    );
+
+    // submit the group when the parent form is submitted
```
aaronBery, line 134: this could do with some test coverage around it

aaronBery: I created a PR into this branch @owensgit  (https://github.com/owensgit/canopy/pull/10)

aaronBery: i pushed this straight into the branch as Owen is unable to review this for a few days

aaronBery (dismissed):

aaronBery:
i've signed the commits as they were unverified

projects/canopy/src/lib/forms/date/date-field.component.html
elenagarrone: Yes, agree but on the other hand it's very easy to use the margin directive. Otherwise when you have an input not followed by another input you have to override it.

projects/canopy/src/lib/forms/sort-code/sort-code.component.scss
elenagarrone: Yes, at the end of the day `rem` is mostly useful when applying font sizes. Not 100% sure we need to use it for width/height.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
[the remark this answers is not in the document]
elenagarrone: I'm sure there is something in the linter that can be specified - not for this story though.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
elenagarrone: Yes, there is definitely a rule. I think it's worth doing it when moving to eslint though as it might need quite a bit of refactoring.

projects/canopy/src/lib/forms/sort-code/sort-code.component.html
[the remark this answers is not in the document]
elenagarrone: I did test it with Voiceover and it reads well 👍

elenagarrone (approved):

aaronBery (approved):
Ah my review fell off after signing the commits. Re-approving :-)

PR #118, opened 2020-12-15
feat(card): add form journey card type

AlexandruHoisan:
# Description

Adds the 'form-journey' card type to the already existing cards.

FYI: The desktop version will be slightly different from the design version as we opted to use the existing grid for page positioning and card width.
The product card was changed to use card-content instead of card-header, and the card-header suffered minor modifications.
All the other cards examples look exactly the same and did not suffer design modifications.

## Requirements

Design link: [https://legalandgeneral.invisionapp.com/share/J4ZQ1VDY96K#/436223077_AN_03-0_ChangeBankOutsideWindow_Annuity_XL](https://legalandgeneral.invisionapp.com/share/J4ZQ1VDY96K#/436223077_AN_03-0_ChangeBankOutsideWindow_Annuity_XL)
[https://legalandgeneral.invisionapp.com/share/P9ZEVI5KVBW#/screens/437597110_AN_02-0_ChangeBeneficiariesForm_Annuity_XL](https://legalandgeneral.invisionapp.com/share/P9ZEVI5KVBW#/screens/437597110_AN_02-0_ChangeBeneficiariesForm_Annuity_XL)
Screenshot: 
Screenshot 2020-12-15 at 16 40 51
Screenshot 2020-12-15 at 11 07 54

# Checklist:

### DIFF
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -89,237 +89,285 @@
     <div lgRow>
       <div lgCol="12" lgColLg="9" lgColMdOffset="0">
         <lg-card lgPaddingTop="none">
-          <lg-tabs [headingLevel]="2" label="Annuities">
-            <lg-tab-item>
-              <lg-tab-item-heading>Tab 1</lg-tab-item-heading>
-              <lg-tab-item-content>
-                <div class="lg-tabs__content-section">
-                  <p>
-                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi posuere
-                    nec leo nec hendrerit. Pellentesque eu lacinia tortor. Donec urna
-                    felis, dictum et faucibus et, interdum ut urna. Nam eleifend eleifend
-                    mi vel blandit. Morbi rutrum a odio in pharetra. Quisque vitae lacus
-                    efficitur, elementum mauris in, porta nisl. Praesent porttitor
-                    accumsan ante, sed efficitur nunc placerat in. Etiam non lorem leo.
-                    Aenean magna lacus, iaculis at velit non, congue commodo dui.
-                    Pellentesque tempus elementum tempor.
-                  </p>
-                </div>
-              </lg-tab-item-content>
-            </lg-tab-item>
-            <lg-tab-item>
-              <lg-tab-item-heading>Tab 2</lg-tab-item-heading>
-              <lg-tab-item-content>
-                <div class="lg-tabs__content-section">
-                  <p>
-                    Maecenas laoreet tristique ligula, mollis ullamcorper est faucibus
-                    eget. Aenean quis placerat arcu. Sed efficitur odio nunc, id facilisis
-                    orci accumsan eget. Nunc feugiat elit eget imperdiet faucibus.
-                    Maecenas faucibus sit amet nisi a ultricies. Donec id scelerisque
-                    ante, eget tempus dui. Fusce semper ex eu lorem pellentesque
-                    tristique. Proin non augue quis orci lobortis rhoncus. Aliquam quis
-                    luctus ipsum. Maecenas vulputate porta mauris, id lacinia turpis
-                    feugiat sed. Aenean et commodo elit, a dignissim massa. Fusce
-                    facilisis laoreet orci quis vehicula. Curabitur eu lacus vitae libero
-                    laoreet hendrerit at quis magna.
-                  </p>
-                </div>
-              </lg-tab-item-content>
-            </lg-tab-item>
-            <lg-tab-item>
-              <lg-tab-item-heading>Tab 3</lg-tab-item-heading>
-              <lg-tab-item-content>
-                <div class="lg-tabs__content-section">
-                  <p>
-                    Phasellus enim sem, dignissim nec ullamcorper id, euismod in magna.
-                    Sed at egestas urna. Donec at nibh eros. Pellentesque at nunc in elit
-                    egestas pellentesque a quis nunc. Praesent commodo risus in metus
-                    tincidunt pretium. Nulla vehicula sem lectus, ac eleifend velit
-                    pellentesque a. Proin et varius ante, nec venenatis nulla.
-                    Pellentesque pharetra risus lorem, et congue ligula interdum volutpat.
-                    Donec ut iaculis metus. Nunc rutrum vestibulum ex nec condimentum.
-                    Pellentesque nec volutpat quam. Etiam tortor arcu, eleifend ut ante
-                    id, ullamcorper tristique libero. Praesent imperdiet, orci in
-                    tincidunt aliquet, quam sapien convallis nunc, non maximus dui lacus
-                    sed lacus. Suspendisse ut tortor dui.
-                  </p>
-                </div>
-              </lg-tab-item-content>
-            </lg-tab-item>
-            <lg-tab-item>
-              <lg-tab-item-heading>Tab 4</lg-tab-item-heading>
-              <lg-tab-item-content>
-                <div class="lg-tabs__content-section">
-                  <p>
-                    Pellentesque finibus ac libero ac rutrum. Morbi faucibus, dui et
-                    efficitur posuere, urna ipsum vehicula dui, in placerat risus felis et
-                    lectus. Vivamus imperdiet nulla nisi, eget varius mi cursus nec.
-                    Suspendisse quis risus eleifend, pretium lectus eget, condimentum leo.
-                    Aliquam faucibus at mi sit amet volutpat. Nunc nec orci sapien. Duis
-                    augue quam, bibendum in enim a, ultricies luctus mi. Aliquam eleifend
-                    magna est, eget sagittis massa malesuada quis. Maecenas mattis tortor
-                    sit amet mi sagittis, vitae aliquam dui rhoncus. Aenean sed erat eu
-                    ante ornare sollicitudin in et est.
-                  </p>
-                </div>
-              </lg-tab-item-content>
-            </lg-tab-item>
-          </lg-tabs>
+          <lg-card-content>
+            <lg-tabs [headingLevel]="2" label="Annuities">
+              <lg-tab-item>
+                <lg-tab-item-heading>Tab 1</lg-tab-item-heading>
+                <lg-tab-item-content>
+                  <div class="lg-tabs__content-section">
+                    <p>
+                      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi posuere
+                      nec leo nec hendrerit. Pellentesque eu lacinia tortor. Donec urna
+                      felis, dictum et faucibus et, interdum ut urna. Nam eleifend eleifend
+                      mi vel blandit. Morbi rutrum a odio in pharetra. Quisque vitae lacus
+                      efficitur, elementum mauris in, porta nisl. Praesent porttitor
+                      accumsan ante, sed efficitur nunc placerat in. Etiam non lorem leo.
+                      Aenean magna lacus, iaculis at velit non, congue commodo dui.
+                      Pellentesque tempus elementum tempor.
+                    </p>
+                  </div>
+                </lg-tab-item-content>
+              </lg-tab-item>
+              <lg-tab-item>
+                <lg-tab-item-heading>Tab 2</lg-tab-item-heading>
+                <lg-tab-item-content>
+                  <div class="lg-tabs__content-section">
+                    <p>
+                      Maecenas laoreet tristique ligula, mollis ullamcorper est faucibus
+                      eget. Aenean quis placerat arcu. Sed efficitur odio nunc, id facilisis
+                      orci accumsan eget. Nunc feugiat elit eget imperdiet faucibus.
+                      Maecenas faucibus sit amet nisi a ultricies. Donec id scelerisque
+                      ante, eget tempus dui. Fusce semper ex eu lorem pellentesque
+                      tristique. Proin non augue quis orci lobortis rhoncus. Aliquam quis
+                      luctus ipsum. Maecenas vulputate porta mauris, id lacinia turpis
+                      feugiat sed. Aenean et commodo elit, a dignissim massa. Fusce
+                      facilisis laoreet orci quis vehicula. Curabitur eu lacus vitae libero
+                      laoreet hendrerit at quis magna.
+                    </p>
+                  </div>
+                </lg-tab-item-content>
+              </lg-tab-item>
+              <lg-tab-item>
+                <lg-tab-item-heading>Tab 3</lg-tab-item-heading>
+                <lg-tab-item-content>
+                  <div class="lg-tabs__content-section">
+                    <p>
+                      Phasellus enim sem, dignissim nec ullamcorper id, euismod in magna.
+                      Sed at egestas urna. Donec at nibh eros. Pellentesque at nunc in elit
+                      egestas pellentesque a quis nunc. Praesent commodo risus in metus
+                      tincidunt pretium. Nulla vehicula sem lectus, ac eleifend velit
+                      pellentesque a. Proin et varius ante, nec venenatis nulla.
+                      Pellentesque pharetra risus lorem, et congue ligula interdum volutpat.
+                      Donec ut iaculis metus. Nunc rutrum vestibulum ex nec condimentum.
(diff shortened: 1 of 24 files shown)


thatguynamedandy:
Hi @AlexandruHoisan I've been chatting to @alexcanning about this as cards are a tricky one that we haven't fully defined from a design side, we maybe should have discussed this one a bit more coding, or if it's that I missed those discussions I do apologise 🙏 😬

I feel like the journey card is a bit specific to it's purpose, from the code it doesn't seem different enough from the normal card. My thoughts are that at the basic level we should have `card`, `card-header`, `card-body or card-content`, `card-footer or card-actions`. This taking a definite nod from [material ui](https://material.angular.io/components/card/examples) but we should feel free to deviate.

These base components should give us enough flexibility for a lot of situations, the form journey being one. We can then have slightly more specific sub components to create more opinionated things like the [product card](https://legal-and-general-canopy.netlify.app/?path=/story/components-card--product), I may be wrong but I don't think the form journey is necessarily warrants it's own components. 

Now if we do all agree on that we would need to make some changes to the existing card implementation.
Notably the horizontal rule goes full width, meaning we would need to move the padding from the card to the sub components, this wouldn't be a breaking change api wise, but it could lead to some visual differences, so we'd need to tread carefully. We should maybe jump on a call to chat this one over?

projects/canopy/src/lib/card/card.component.ts
```diff
@@ -5,28 +5,46 @@ import {
...
 } from '@angular/core';
 
 import { LgCardContentComponent } from './card-content/card-content.component';
 import { LgCardHeaderComponent } from './card-header/card-header.component';
+import { LgCardBottomAreaComponent } from './card-bottom-area/card-bottom-area.component';
+import { LgCardTopAreaComponent } from './card-top-area/card-top-area.component';
 
 @Component({
-  selector: 'lg-card',
+  selector: 'lg-card, lg-card-form-journey',
```
thatguynamedandy, line 17: I think if we are to provide a form-journey variant we would probably do it with a variant property instead of a distinct component, however it may be that we can achieve it already with the grid and margin/padding directives.

AlexandruHoisan:
> Hi @AlexandruHoisan I've been chatting to @alexcanning about this as cards are a tricky one that we haven't fully defined from a design side, we maybe should have discussed this one a bit more coding, or if it's that I missed those discussions I do apologise 🙏 😬
> 
> I feel like the journey card is a bit specific to it's purpose, from the code it doesn't seem different enough from the normal card. My thoughts are that at the basic level we should have `card`, `card-header`, `card-body or card-content`, `card-footer or card-actions`. This taking a definite nod from [material ui](https://material.angular.io/components/card/examples) but we should feel free to deviate.
> 
> These base components should give us enough flexibility for a lot of situations, the form journey being one. We can then have slightly more specific sub components to create more opinionated things like the [product card](https://legal-and-general-canopy.netlify.app/?path=/story/components-card--product), I may be wrong but I don't think the form journey is necessarily warrants it's own components.
> 
> Now if we do all agree on that we would need to make some changes to the existing card implementation.
> Notably the horizontal rule goes full width, meaning we would need to move the padding from the card to the sub components, this wouldn't be a breaking change api wise, but it could lead to some visual differences, so we'd need to tread carefully. We should maybe jump on a call to chat this one over?

FYI-for other reviewers:
Waiting to have a meeting on this one before further modifications are done.

elenagarrone:
> Hi @AlexandruHoisan I've been chatting to @alexcanning about this as cards are a tricky one that we haven't fully defined from a design side, we maybe should have discussed this one a bit more coding, or if it's that I missed those discussions I do apologise 🙏 😬
> 
> I feel like the journey card is a bit specific to it's purpose, from the code it doesn't seem different enough from the normal card. My thoughts are that at the basic level we should have `card`, `card-header`, `card-body or card-content`, `card-footer or card-actions`. This taking a definite nod from [material ui](https://material.angular.io/components/card/examples) but we should feel free to deviate.
> 
> These base components should give us enough flexibility for a lot of situations, the form journey being one. We can then have slightly more specific sub components to create more opinionated things like the [product card](https://legal-and-general-canopy.netlify.app/?path=/story/components-card--product), I may be wrong but I don't think the form journey is necessarily warrants it's own components.
> 
> Now if we do all agree on that we would need to make some changes to the existing card implementation.
> Notably the horizontal rule goes full width, meaning we would need to move the padding from the card to the sub components, this wouldn't be a breaking change api wise, but it could lead to some visual differences, so we'd need to tread carefully. We should maybe jump on a call to chat this one over?

I have to agree with @thatguynamedandy. I don't think this needs its own components - I would be prone to make the changes to the existing card instead like Andy suggested.

projects/canopy/src/lib/card/card.component.spec.ts
```diff
@@ -42,50 +41,32 @@ describe('LgCardComponent', () => {
     expect(el.getAttribute('class')).toContain('lg-card');
   });
 
-  describe('when there is lg-card-header', () => {
+  describe('when there is lg-card-header, lg-card-content and lg-card-footer', () => {
```
owensgit, line 44: I think we should be testing the order of the components here as well, to make sure they appear as header first, content second and footer last.

With that in mind, is it worth having a suite of tests for when there's different combinations? E.g. a card without header and footer, a card with just a header and content? Not sure on that one 🤔

projects/canopy/src/lib/card/card.module.ts
```diff
@@ -11,13 +11,17 @@ import { LgCardHeaderComponent } from './card-header/card-header.component';
 import { LgCardSubtitleComponent } from './card-subtitle/card-subtitle.component';
 import { LgCardTitleComponent } from './card-title/card-title.component';
 import { LgCardComponent } from './card.component';
+import * as iconSet from '../icon/icons.interface';
```
owensgit, line 14: Maybe we should just import the chevron icon and not the whole set to help with tree shaking?

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -250,6 +250,53 @@
             </lg-card-header>
           </lg-card>
 
+          <div lgContainer>
+            <div lgRow>
+              <div lgCol="12" lgColLg="8" lgColLgOffset="2" lgColMd="10" lgColMdOffset="1">
+                <lg-card lgPadding="none">
```
owensgit, line 256: Part of me is wondering if there's a precarious balance of margin/padding directives going on here, and it relies on the developer getting it just right, or ensuring that they copy and paste this example exactly. I'm not sure if you discussed this on the call and apologies if you have, but I do wonder it it's worth considering having a card variant for the Form Journey, which sets the classes on itself and it's children to make sure it displays with all the correct margin and padding. 

It could even apply the border on the header/footer, in case there are scenarios where we don't want a border (e.g. `--with-border`), although that could be [YAGNI](https://www.wikiwand.com/en/You_aren%27t_gonna_need_it). 

What do you think? Again apologies if I'm treating old ground at this late stage!

owensgit (commented):
Looks great, thanks @AlexandruHoisan, just a couple of comments and one question about the approach. Happy to jump on a call anytime.

projects/canopy/src/lib/card/card.component.scss
```diff
@@ -1,3 +1,5 @@
+@import '../../styles/mixins';
```
thatguynamedandy, line 1: Did you need this include? I couldn't see it referenced anywhere

projects/canopy/src/lib/card/card.component.ts
```diff
@@ -25,11 +25,8 @@ export class LgCardComponent implements AfterContentInit {
...
-  cardContent: QueryList<LgCardContentComponent>;
+  @ContentChild(LgCardFooterComponent)
+  lgCardFooterComponent: LgCardFooterComponent;
 
-  ngAfterContentInit() {
-    if (this.lgCardHeaderComponent && this.lgCardContentComponent) {
-      this.lgCardHeaderComponent.hasContent = true;
-    }
-  }
+  cardContent: QueryList<LgCardContentComponent>;
```
thatguynamedandy, line 31: I might be wrong but I suspect none of these content children are used anymore and could all go away.

projects/canopy/src/lib/card/card.component.spec.ts
thatguynamedandy: I noticed that in the main card.component.html the template is just 
```html
<ng-content></ng-content>
```
It might be good to make it stricter...
```html
<ng-content select="lg-card-header"></ng-content>
<ng-content select="lg-card-content"></ng-content>
<ng-content select="lg-card-footer"></ng-content>
```
It's unlikely that anyone will render them in the wrong order, but feels like good practice to enforce it.

projects/canopy/src/lib/card/card.module.ts
thatguynamedandy: I wondered if actually this should go into the consuming application and not this module. It ties the card module quite specifically to the use case of there being a back button in the header. There will be lots of scenarios where people won't need the chevron, but they'll be getting it loaded anyway.

projects/canopy-test-app/src/app/app.component.html
thatguynamedandy: For me the form journey variant still felt like one step too far into a specific use case, I may be wrong on this as @alexcanning  has pointed out it will be a common pattern. However I wouldn't want us to build up a catalogue of variants that we have to support.

We still haven't really looked at the card design holistically, it could be that they should all have the same paddings and margins but we just don't know it yet. Might be worth a real quick check in on that point only and see what effect it would have on the other card examples.

thatguynamedandy (commented):
Great work @AlexandruHoisan, Cards are starting to look a bit tidier, I haven't quite finished this review but figured I would post it now as you may well start earlier on Monday 👍

projects/canopy/src/lib/card/card.notes.ts
```diff
@@ -112,6 +175,9 @@ Generate the markup as show in the example below, no current modifiers.
     <div class="lg-card-content">
         Your content
     </div>
+    <div class="lg-card-footer">
+        Your content
```
thatguynamedandy, line 179:
```suggestion
        Your footer
```
Minor, but I guess the original 'Your content' was written to match the fact that it was the content component.

projects/canopy/src/lib/card/card.component.spec.ts
AlexandruHoisan: Made the card.component template stricter as Andy suggested and also added some test for other scenarios, e.g. a card without header and footer, a card with just a header and content and a card with just the content component. Not sure here if it's not a bit too much testing, as maybe just the one test for the scenario where all of them are rendered is enough.

projects/canopy/src/lib/card/card.module.ts
AlexandruHoisan: Took this out of the card.module as I also think this should go in the consuming app. The same is done with the Quick actions component icons and I also think we should keep consistency.

projects/canopy/src/lib/card/card.component.scss
AlexandruHoisan: Nope, removed it.

projects/canopy/src/lib/card/card.component.ts
AlexandruHoisan: Yep, that is correct. 
Done.

projects/canopy/src/lib/card/card.notes.ts
AlexandruHoisan: Done.

projects/canopy-test-app/src/app/app.component.html
AlexandruHoisan: Not sure what's the best course of action here. Without the variant, making use of canopy Grid, we can get a variety of cards out of the box, on the other hand the variant seems to be more clean or maybe it looks like that to me. I am good to go with both approaches, both of them giving great flexibility to the user. Not sure what you mean by "quick check in on that point only and see what effect it would have on the other card examples." as this is how I've built it once, with a variant, and I did not modify the other card examples.
I'm good with either implementation so waiting for thoughts on this one. :)

thatguynamedandy: I'm a bit stuck too, What I was wondering is..

1. We make these the default spacings for the card-header: `<lg-card-header lgPadding="sm" lgPaddingBottom="xs" lgMarginBottom="lg">`(I moved the marginTop from the lgContainer)
2. If the other storybook examples weren't badly affected then we proceed with that. The other examples with card headers have never had any formal design review so I'm not too worried about visual changes to them.

I think one we have a sensible margin and padding setup that we keep consistent that should make the cards a bit easier. Then we could have a form journey variant, but I wonder if it would actually need to be a component in it's own right as it will probably need all the grid markup to be responsive, we could defer that one for now to ease this in. Another opinion would be handy I think on the form journey question.

AlexandruHoisan: After a short discussion with Andy, I moved all the added spacing via margins/paddings directives just to the <lg-card-header> and the other cards did not suffer any changes from a design point of view.
Regarding the variant, we decided not to have one at the moment, but can be revisited in the future if and when we get further designs on the other cards like: standard, product and so on.

owensgit: Sounds good to me 🙂

owensgit (dismissed):

owensgit (approved):

projects/canopy/src/lib/card/card.notes.ts
```diff
@@ -30,28 +30,86 @@ or for a Product card:
...
-        <span><span class="lg-font-size-3">£</span>230.20</span>
-      </lg-card-principle-data-point-value>
-      <lg-card-principle-data-point-date>
-        as of 01 Jan 2020
-      </lg-card-principle-data-point-date>
-    </lg-card-principle-data-point>
-  </lg-card-header>
+  <lg-card-content>
+    <div lgRow>
+      <div lgCol=12 lgColMd=6>
```
elenagarrone, line 35: Just a small one but the columns values should have `"` e.g.:
```
<div lgCol="12" lgColMd="6">
```

There are few instances of this in here.

projects/canopy/src/lib/card/card.stories.ts
```diff
@@ -1,19 +1,100 @@
...
+  @Output() formSubmit: EventEmitter<void> = new EventEmitter();
+
+  form: FormGroup;
+
+  constructor(public fb: FormBuilder) {
+    this.form = this.fb.group({ accountNumber: { value: '', disabled: false } });
+    this.form.valueChanges.subscribe((val) => this.inputChange.emit(val));
+  }
+
+  onSubmit(event) {
```
elenagarrone, line 72: Should have a `void` return type

projects/canopy/src/lib/card/card.notes.ts
```diff
@@ -30,28 +30,86 @@ or for a Product card:
...
+          </lg-breadcrumb>
+        </lg-card-header>
+        <lg-card-content>
+          <div lgContainer>
+            <div lgRow>
+              <div lgCol="12" lgColMd="10" lgColMdOffset="1">
+                <lg-card-title headingLevel="4">
+                {{ title }}
+                </lg-card-title>
+                <lg-separator></lg-separator>
```
elenagarrone, line 83: This component has a `role="separator"` which in my opinion is going to be wrong in most cases as most of the time we'll use it just for styling. To get around that I think we should add a `aria-hidden="true"` attribute to this instance so that screen readers will skip it. This is because in this case the heading doesn't have to be "separated" from the content it refers to.

I'll open a discussion to talk about the separator. For the moment the attribute in this specific case will be fine.

projects/canopy/src/lib/card/card.notes.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/card/card.stories.ts
AlexandruHoisan: Done.

projects/canopy/src/lib/card/card.notes.ts
AlexandruHoisan: Done.

elenagarrone (approved):

owensgit (approved):

PR #120, opened 2020-12-15
fix(checkbox-group): uncheck child toggles when parent form is reset

paul-request:
Calling `form.reset()` from the parent form failed to uncheck the checked child toggle components.
So I have updated the `value` setter to uncheck all children when an empty array is passed in.

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.ts
+++ b/projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.ts
@@ -102,21 +102,22 @@ export class LgCheckboxGroupComponent implements ControlValueAccessor {
     this._validationElement = element;
   }
 
-  _value: Array<string> = null;
+  private _value: Array<string> = [];
   @Input()
   get value() {
     return this._value;
   }
-  set value(value) {
+  set value(value: Array<string>) {
     this._value = value;
     this.onChange(value);
-    if (this.checkboxes) {
-      this.checkboxes.forEach((checkbox) => {
-        if (value.includes(checkbox.value.toString())) {
-          checkbox.checked = true;
-        }
-      });
+
+    if (!this.checkboxes) {
+      return;
     }
+
+    this.checkboxes.forEach(
+      (checkbox) => (checkbox.checked = value.includes(checkbox.value.toString())),
+    );
   }
 
   @Input()
--- a/projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.spec.ts
+++ b/projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.spec.ts
@@ -138,6 +138,17 @@ describe('LgCheckboxGroupComponent', () => {
     expect(checkedOption.componentInstance.value).toContain('red');
   });
 
+  it('unchecks the selected checkbox buttons when an empty array value is provided', () => {
+    groupInstance.value = ['red'];
+    fixture.detectChanges();
+    groupInstance.value = [];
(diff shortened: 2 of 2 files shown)


projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.ts
```diff
@@ -102,20 +102,25 @@ export class LgCheckboxGroupComponent implements ControlValueAccessor {
...
     this._value = value;
     this.onChange(value);
+
     if (this.checkboxes) {
-      this.checkboxes.forEach((checkbox) => {
-        if (value.includes(checkbox.value.toString())) {
-          checkbox.checked = true;
-        }
-      });
+      if (value.length > 0) {
```
thatguynamedandy, line 115: I find the double 'if' statement makes this a bit harder to understand, I wasn't sure if this would do the same, or does it not account for certain situations?

```js
if (this.checkboxes) {
  this.checkboxes.forEach((checkbox) => {
    checkbox.checked = value.includes(checkbox.value.toString()) ? true : false;
  })
}
```

aaronBery: alternatively invert the if statements as a rule of thumb can reduce nesting

i.e.
```
if (!this.checkboxes) {
  return;
}
```

tools such as resharper (albeit not typescript) are known for promoting this pattern https://www.jetbrains.com/help/resharper/InvertIf.html

thatguynamedandy: yeah I like an early return, I think if the first approach doesn't work for some reason this would be a good option. It would be interesting though to see if the first approach doesn't work as it may imply something else odd is happening.

paul-request: Yeah good point, I think it wasn't working before because of some other change. Have returned earlier and simplified the if statements @thatguynamedandy @aaronBery

projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.ts
```diff
@@ -102,21 +102,23 @@ export class LgCheckboxGroupComponent implements ControlValueAccessor {
     this._validationElement = element;
   }
 
-  _value: Array<string> = null;
+  _value: Array<string> = [];
```
aaronBery, line 105: we can make this private

paul-request: done

projects/canopy/src/lib/forms/checkbox-group/checkbox-group.component.ts
```diff
@@ -102,21 +102,23 @@ export class LgCheckboxGroupComponent implements ControlValueAccessor {
...
-        }
-      });
+
+    if (!this.checkboxes) {
+      return;
     }
+
+    this.checkboxes.forEach(
+      (checkbox) =>
+        (checkbox.checked = value.includes(checkbox.value.toString()) ? true : false),
```
aaronBery, line 120: do we need the ternary expression here?

`(method) Array<string>.includes(searchElement: string, fromIndex?: number): boolean`

paul-request: ah good spot, yes we can simplify that

aaronBery (approved):

thatguynamedandy (approved):

PR #123, opened 2020-12-17
ci(semantic-release): update config to create pre-releases

elenagarrone:
# Description

After creating the `next` branch with the Angular upgrade I've noticed that the version was set to `4.0.0` instead of `4.0.0-next.1`. 
The documentation says that if we set `"prerelease": true` this will take the name of the branch which did not happen in our case 🤔  I have updated the config to be more specific `"prerelease": "next"` hoping that this fixes the issue.
More info here: https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#prerelease

I've also added the `channel` property which will allow users to install using the `@next` tag.

# Checklist:

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,6 +1,9 @@
 {
   "plugins": [
-    "branches": [ "master", {"name": "next", "prerelease": true} ],
+    "branches": [
+      "master",
+      {"name": "next", "channel": "next", "prerelease": "next"}
+    ],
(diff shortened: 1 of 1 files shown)


thatguynamedandy (approved):
Looks like it should work, this is pretty new to me too so a good experiment.

paul-request (approved):

PR #136, opened 2020-12-21
Remove references to non-existent css variables

### LINKED ISSUE #80: Non-existent CSS variables
Over time we seem to have accrued quite a few references to css variables that don't exist. I stumbled upon this when testing IE11 and noticing the console warnings from the css vars polyfill (there might be an easier way), but here is the current list..

Screenshot 2020-11-25 at 10 45 07

It may be worth researching each case individually, particularly the `line-height-sm` one just in case we regressed visually without realising. Visual testing is coming at some stage soon.

thatguynamedandy:
# Description

Somewhere along the way we removed some css variables but didn't update all of the components that used them. We haven't had any issues raised from a styling point of view so for now I just removed all of the references as opposed to researching where they were individually removed. I can't foresee any unintended consequences. 

We could potentially do some work to prevent this happening again, but for now I've just done the minimal amount to clear the bug out of the way.

Fixes #80

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
+++ b/projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
@@ -12,7 +12,6 @@
   position: relative;
   text-align: left;
   width: 100%;
-  line-height: var(--line-height-body);
 
   &:focus {
     @include lg-inner-focus-outline(var(--default-focus-color));
--- a/projects/canopy/src/lib/button/button.component.scss
+++ b/projects/canopy/src/lib/button/button.component.scss
@@ -56,11 +56,6 @@
     cursor: not-allowed;
   }
 
-  &--rounded {
-    border-radius: var(--btn-rounded-radius);
-    min-width: auto;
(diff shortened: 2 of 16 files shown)


owensgit (commented):
Thanks for doing this clean up Andy. I noticed there's a merge commit on this branch, should we look to remove that to keep our Git history clean?

elenagarrone:
> Thanks for doing this clean up Andy. I noticed there's a merge commit on this branch, should we look to remove that to keep our Git history clean?

@owensgit the merge commit is added when updating with master. It is shown in the history of this branch but will not be in the master history. I know it sounds impossible but it's true :D

owensgit (approved):

thatguynamedandy:
> > Thanks for doing this clean up Andy. I noticed there's a merge commit on this branch, should we look to remove that to keep our Git history clean?
> 
> @owensgit the merge commit is added when updating with master. It is shown in the history of this branch but will not be in the master history. I know it sounds impossible but it's true :D

I've always found that quite confusing too. I think there are some articles about the rebase merge on GitHub as I know it does a few different things.

PR #146, opened 2021-01-07
feat(table): add configurable fullSizeBp setting

pauleustice:
# Description

The LgTable currently always uses the non-mobile layout at and above the `md` breakpoint. However, if the table is placed in a container that does not fill the width of the screen then it can be desirable to show the mobile layout at larger screen widths.

This PR adds a new input which allows the consumer to specify at which breakpoint the full size layout should be displayed: 

`[fullSizeBp]="'sm'"` or `[fullSizeBp]="'md'"` or `[fullSizeBp]="'lg'"`

It defaults to using `md` if not provided, so should not introduce a breaking change. I also improved the LgTable storybook Knobs to make them easier to use.

## Requirements

Storybook link: https://deploy-preview-146--legal-and-general-canopy.netlify.app/

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
+++ b/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
@@ -1,13 +1,9 @@
 @import '../../../styles/mixins.scss';
+@import '../table.breakpoints';
 
 .lg-table-cell {
   display: flex;
   align-items: baseline;
-
-  @include lg-breakpoint('md') {
-    display: table-cell;
-    padding: var(--space-sm);
-  }
 
   &--toggle-cell {
     order: 1;
@@ -20,25 +16,11 @@
     .lg-table-cell__content {
       margin-left: 0;
     }
-
-    @include lg-breakpoint('md') {
-      min-width: var(--table-toggle-width);
-    }
   }
 }
 
 .lg-table-cell__content {
   margin-left: auto;
-
-  @include lg-breakpoint('md') {
-    text-align: left;
-  }
-
-  &--align-end {
-    @include lg-breakpoint('md') {
-      text-align: right;
-    }
-  }
 }
 
 .lg-table-cell--expandable-content {
@@ -49,8 +31,40 @@
 
 .lg-table-cell__label {
   font-weight: var(--font-weight-bold);
+}
 
-  @include lg-breakpoint('md') {
-    display: none;
+@each $columns-breakpoint in $columns-breakpoints {
+  .lg-table--columns-#{$columns-breakpoint} {
+    .lg-table-cell {
+      @include lg-breakpoint($columns-breakpoint) {
+        display: table-cell;
+        padding: var(--space-sm);
+      }
+
+      &--toggle-cell {
+        @include lg-breakpoint($columns-breakpoint) {
+          min-width: var(--table-toggle-width);
+          padding: 0;
+        }
+      }
+
+      &__content {
+        @include lg-breakpoint($columns-breakpoint) {
+          text-align: left;
+        }
+
+        &--align-end {
+          @include lg-breakpoint($columns-breakpoint) {
+            text-align: right;
+          }
+        }
+      }
+
+      &__label {
+        @include lg-breakpoint($columns-breakpoint) {
+          display: none;
+        }
+      }
+    }
   }
 }
--- a/projects/canopy/src/lib/table/table-head-cell/table-head-cell.component.scss
+++ b/projects/canopy/src/lib/table/table-head-cell/table-head-cell.component.scss
@@ -1,4 +1,5 @@
 @import '../../../styles/mixins.scss';
+@import '../table.breakpoints';
 
 .lg-table-head-cell {
   display: block;
@@ -7,12 +8,18 @@
   border-bottom: solid var(--table-header-border-width) var(--table-header-border-color);
   padding: var(--space-xs) 0;
 
-  @include lg-breakpoint('md') {
-    padding: var(--space-sm);
-    display: table-cell;
-  }
-
   &--align-end {
     text-align: right;
+  }
+}
+
+@each $columns-breakpoint in $columns-breakpoints {
+  .lg-table--columns-#{$columns-breakpoint} {
+    .lg-table-head-cell {
+      @include lg-breakpoint($columns-breakpoint) {
+        padding: var(--space-sm);
+        display: table-cell;
+      }
+    }
   }
 }
--- a/projects/canopy/src/lib/table/table.interface.ts
+++ b/projects/canopy/src/lib/table/table.interface.ts
@@ -8,3 +8,9 @@ export interface TableColumn {
   align: AlignmentOptions;
   showLabel: boolean;
 }
+
+export enum TableColumnLayoutBreakpoints {
+  Small = 'sm',
+  Medium = 'md',
+  Large = 'lg',
+}
--- a/projects/canopy/src/lib/table/table.notes.ts
+++ b/projects/canopy/src/lib/table/table.notes.ts
@@ -71,6 +71,14 @@ and in your HTML:
 
 ## Inputs
 
+### LgTableComponent
+A component that acts as a standard table.
+
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| \`\`showColumnsAt\`\` | Sets the minimum screen width from which the column layout is displayed. Accepts \`sm\`, \`md\` or \`lg\` | string | 'md' | No |
+
+
 ### LgTableBodyComponent
 A component that acts as a standard table body.
 
--- a/projects/canopy/src/lib/table/table.notes.ts
+++ b/projects/canopy/src/lib/table/table.notes.ts
@@ -101,7 +101,7 @@ A component that acts as a standard table header cell.
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
 | \`\`align\`\` | Alignment option for the header and its respective column | \`\`start\`\`, \`\`end\`\` | n/a | No |
-| \`\`showLabel\`\` | Whether to show label for this header in each row on mobile | boolean | true | No |
+| \`\`showLabel\`\` | Whether to show label for this header in each row in non-column layout | boolean | true | No |
 
 ### LgTableRowComponent
 A component that acts as a standard table row.
--- a/projects/canopy/src/lib/table/table/table.component.ts
+++ b/projects/canopy/src/lib/table/table/table.component.ts
@@ -1,15 +1,19 @@
 import {
+  AfterContentChecked,
   ChangeDetectionStrategy,
   Component,
-  HostBinding,
-  ViewEncapsulation,
-  AfterContentChecked,
   ContentChild,
+  ElementRef,
+  HostBinding,
+  Input,
+  OnInit,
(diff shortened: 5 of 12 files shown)


projects/canopy/src/lib/table/table-cell/table-cell.component.scss
```diff
@@ -1,14 +1,11 @@
 @import '../../../styles/mixins.scss';
 
+$full-size-breakpoints: 'sm', 'md', 'lg' !default;
```
pauleustice, line 3: I considered whether instances of these should be consolidated into a single scss file within `lib/table` but couldn't find a precedent for a similar approach. Please let me know if you think I should do this.

elenagarrone (dismissed):
Nice one @pauleustice 👍  

For everyone else, I used `stack` as an input for the radio buttons component but Paul rightly pointed out that that's not really mobile friendly. I'll be updating the radio component in a separate PR so that it's consistent with this approach.

elenagarrone (dismissed):

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
elenagarrone: I think it could probably go in an even more global place so that we can use the same breakpoints everywhere?

projects/canopy/src/lib/table/table-head-cell/table-head-cell.component.scss
```diff
@@ -1,18 +1,26 @@
 @import '../../../styles/mixins.scss';
 
+$full-size-breakpoints: 'sm', 'md', 'lg' !default;
```
aaronBery, line 3: does this need to be re-declared throughout the tables sub components?

aaronBery:
looks good and works well. just a few minor comments/questions @pauleustice  :-)

projects/canopy/src/lib/table/table-head-cell/table-head-cell.component.scss
pauleustice: Agreed, it doesn't. Please see the comment I left: https://github.com/Legal-and-General/canopy/pull/146#discussion_r553318520

Just need to know where Canopy peeps would prefer I put it :)

projects/canopy/src/lib/table/table/table.component.ts
```diff
@@ -32,12 +35,35 @@ export class LgTableComponent implements AfterContentChecked {
     return this.isExpandable;
   }
 
+  _fullSizeBp: TableFullSizeBreakpoints;
```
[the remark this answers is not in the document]
pauleustice: Resolved

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
pauleustice: @elenagarrone If there are other similar variables a global approach makes sense, otherwise I think within /lib/table would be better

elenagarrone: The radio also uses something similar`$stack-breakpoints: 'sm', 'md', 'lg' !default;`. 
Either you put `$full-size-breakpoints` in `/lib/table` and then once I modify the radio button I move it or you just move it now to a more global place and then I'll modify the radio variable name in a separate PR - together with some more bits to make it mobile friendly.

I don't mind which approach to be honest.

projects/canopy/src/lib/table/table.interface.ts
```diff
@@ -8,3 +8,9 @@ export interface TableColumn {
   align: AlignmentOptions;
   showLabel: boolean;
 }
+
+export enum TableFullSizeBreakpoints {
```
thatguynamedandy, line 12: We have an ongoing discussion around enums vs union types that we haven't really resolved yet. I wouldn't hold this PR up for it but any input you might have into it would be great.

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -71,6 +71,14 @@ and in your HTML:
 
 ## Inputs
 
+### LgTableComponent
+A component that acts as a standard table.
+
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| \`\`fullSizeBp\`\` | Sets the minimum screen width from which the full size table layout is displayed. Accepts \`sm\`, \`md\` or \`lg\` | string | 'md' | No |
```
thatguynamedandy, line 79: I'm not sure about the name of the property, I think it's the '**Bp**' that gets me as it might not be that clear what it stands for at first glance. I think this is only the second time we have used a property with a media breakpoint in it to define component behaviour (the radio stack being the other one), so we don't really have a convention yet.

Annoyingly I'm not sure of an alternative, from the design side the naming conventions are 'wide' and 'narrow', but I don't know if that's useful. 

`showWide="md"`
`fullSizeBreakpoint ="md"`
`wideAt="md"`
`wideBreakpoint="md"`

'wide' instead of `fullSize` is slightly more terse especially when used in the css, and it aligns with designs, but is it descriptive enough. Open to another opinion on that I think.

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
thatguynamedandy: Should they be the same? I'm just wondering if it's chance that on this occasion they happen to have the same properties.

> Nice one @pauleustice 👍
> 
> For everyone else, I used `stack` as an input for the radio buttons component but Paul rightly pointed out that that's not really mobile friendly. I'll be updating the radio component in a separate PR so that it's consistent with this approach.

@elenagarrone are you able to elaborate on that a bit? Do you mean you'll be changing the name of the input or is there more to it? I didn't quite pick up why it's not very mobile friendly.

thatguynamedandy (commented):
Makes sense, I think it's a good addition, just a couple of minor comments before I thumb it :-)

thatguynamedandy:
> Nice one @pauleustice 👍
> 
> For everyone else, I used `stack` as an input for the radio buttons component but Paul rightly pointed out that that's not really mobile friendly. I'll be updating the radio component in a separate PR so that it's consistent with this approach.

@elenagarrone are you able to elaborate on that a bit? Do you mean you'll be changing the name of the input or is there more to it? I didn't quite pick up why it's not very mobile friendly.

pauleustice:
> > Nice one @pauleustice 👍
> > For everyone else, I used `stack` as an input for the radio buttons component but Paul rightly pointed out that that's not really mobile friendly. I'll be updating the radio component in a separate PR so that it's consistent with this approach.
> 
> @elenagarrone are you able to elaborate on that a bit? Do you mean you'll be changing the name of the input or is there more to it? I didn't quite pick up why it's not very mobile friendly.

@thatguynamedandy It's not mobile friendly as it is not applying the stacking (which should be the mobile/default layout of the component) by default. So the work would involve reversing the way the styles are applied, effectively.

projects/canopy/src/lib/table/table.interface.ts
pauleustice: I think there are good reasons not to use enums (such as them being compiled), but I do like the syntax. Objectively, with all things considered, we should probably switch to union types IMO. For now though, this at least meets the precedent used in the radio group stack

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
pauleustice: @thatguynamedandy I think they should be kept separate, personally.

It does raise an interesting point that I considered when assessing the approach to take with this component. This method of using existing breakpoint names/sizes keeps things standardised (and in turn matches the grid layout and other media query-related styles), but on the other hand it doesn't allow the consumer to specify precise values based on their needs, and only supports one value.

My initial thinking before Elena told me about the existing radio stack approach was to make use of a `ResizeObserver` that listened to breakpoints specified within the input, but this was a much larger piece of work to consider. It would likely need a Canopy resize observer module that certain components (table and radio group at present) would depend on. Furthermore it isn't supported by IE.

It could have looked something like `<table lgTable [fullWidthAt]="{ sm: true, md: false }">` which would be nice. Maybe to consider in future?

projects/canopy/src/lib/table/table.notes.ts
pauleustice: @thatguynamedandy I'm wary of 'wide' since it implies the width of the table rather than the layout used. But, I'm having a hard time coming up with better names. I prefer `fullSizeAt` or `fullSizeBreakpoint` but they still don't seem great.

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
pauleustice: Moved to `lib/table/table.breakpoints.scss`

projects/canopy/src/lib/table/table.notes.ts
pauleustice: @thatguynamedandy @aaronBery How about `scaleUpAt="md"`?

projects/canopy/src/lib/table/table/table.component.ts
```diff
@@ -57,6 +72,26 @@ export class LgTableComponent implements AfterContentChecked {
...
+      this.hostElement.nativeElement,
+      `lg-table--columns-${this._columnsBreakpoint}`,
+    );
+
+    this.renderer.addClass(
+      this.hostElement.nativeElement,
+      `lg-table--columns-${columnsBreakpoint}`,
+    );
+
+    this._columnsBreakpoint = columnsBreakpoint;
```
thatguynamedandy, line 92: Sorry to be fussy but I wondered if this line should go in the `setter`. I found it odd that the `getter` returned the private variable but the `setter` didn't set it. Which does mean the ngOnInit would need to change as well 😬

I think that pattern that we normally use the one below in the alert component. I haven't considered it too much before but I guess there is probably some additional functional calls

https://github.com/Legal-and-General/canopy/blob/master/projects/canopy/src/lib/alert/alert.component.ts

projects/canopy/src/lib/table/table/table.component.ts
```diff
@@ -32,12 +36,23 @@ export class LgTableComponent implements AfterContentChecked {
     return this.isExpandable;
   }
 
+  private _columnsBreakpoint: TableColumnLayoutBreakpoints;
```
thatguynamedandy, line 39: could the private variable shadow the name of the Input?
```suggestion
  private _showColumnsAt: TableColumnLayoutBreakpoints;
```

aaronBery:
Sorry @pauleustice , just testing this again and I've found a slight inconsistency.

When in non-column mode on large mobile and the author label deselected the author label is not visible

Screenshot 2021-01-08 at 16 20 19

however when in column mode still on large mobile and the display author label still deselected the author label is now visible

Screenshot 2021-01-08 at 16 19 32

projects/canopy/src/lib/table/table/table.component.ts
pauleustice: Fixed

projects/canopy/src/lib/table/table/table.component.ts
pauleustice: Fixed

thatguynamedandy (approved):

aaronBery (approved):
could potentially squash the commits and use the commit message from the first commit. functionality and code looks good :-)

Issue #35, opened 2020-10-14
Upgrade to storybook 6


aaronBery:
This was initially flagged in https://github.com/Legal-and-General/canopy/pull/28 which references a handy [migration guide](https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5) on medium.

There are breaking changes associated with the upgrade and also decisions on best practice to be made e.g. which story format to adopt

elenagarrone:
Also, you will have to consider the dsm storybook npm package. I quickly had a look at this when doing the dependencies upgrades a while ago and it didn't seem that straight forward. It might have changed in the past month 🤞

thatguynamedandy:
I'm not happy with that DSM package and the amount of dependencies it has, which potentially hold us back on updates. A new version did get published two weeks ago, but from what I can see the release notes aren't published.
https://www.npmjs.com/package/@invisionapp/dsm-storybook

I wouldn't be opposed to looking at alternatives, to me it would seem simpler if we just iframed the storybook page into the DSM.

thatguynamedandy:
I had a few goes at playing around with this recently, there are some pretty big changes. 

I made better progress by removing the existing `.storybook' directory and then just running a single story. The `Alert` component was a good candidate for simplicity. From what I could see though, content projection prevents us from using the simplest approach.

I think Storybook 6 might reduce lots of the documentation we need to do. We could consider removing the html only support from the docs.

thatguynamedandy:
@paul-request what were your findings in the end? did you have any further ideas on how to tackle it?

paul-request:
So I didn't get as far as I would have liked, but I have managed to upgrade and get it running with an Alert component.

These are the main things to take away I think:

- Automatic documentation is achieved by using Compodoc internally, which has turned out to be massively disappointing. It doesn't do much more than add descriptions. So actually we would still be required to add documentation for local variables and functions manually.
- Compodoc also adds private variables into the docs and controls, so these would have to be ignored explicitly using a comment (not ideal).
- Knobs is replaced by Controls, which has a new API to use. Seems fine, but there is an outstanding bug where changing the controls in the docs tab doesn't update the example.  (https://github.com/storybookjs/storybook/issues/11908)

Still todo:
- There is a new feature called [DocsPage](https://storybook.js.org/docs/angular/writing-docs/introduction), which I need to look into that further. At first glance, it looks like we can just use it by creating a markdown file, instead of exporting strings of markdown from ts.
[...]

PR #165, opened 2021-01-18
docs(storybook): removes storybook docs from dependabot updates

thatguynamedandy:
Temporarily removing @storybook scope modules from the dependabot updates to reduce the amount of noise. Once we
have migrated to storybook 6 we should revert this change..

re #35


elenagarrone (dismissed):

owensgit (approved):
Good idea Andy 👍

.github/dependabot.yml
```diff
@@ -23,3 +23,6 @@ updates:
       prefix: 'npm'
       include: 'scope'
     versioning-strategy: increase
+    ignored_updates:
```
elenagarrone, line 26: Are you sure that's right?
In our project we use:

```
ignore:
      - dependency-name: ''
```

thatguynamedandy:
Just as I did this I noticed a storybook 5.3.x upgrade which I assumed wasn't going to happen. I might see if I can rewrite it to only ignore 6.x.x version

.github/dependabot.yml
elenagarrone: Here are the GitHub docs: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#ignore

elenagarrone (commented):
Please see my comment

.github/dependabot.yml
thatguynamedandy: good question, I think I followed the docs correctly..
https://dependabot.com/docs/config-file/#ignored_updates

elenagarrone: I'm pretty sure the Dependabot config (and docs) are the same as the Dependabot config/docs for GitHub. 
As far as I know the Github dependabot as less configs.

elenagarrone (approved):
👍

owensgit (approved):

PR #167, opened 2021-01-19
feat(breadcrumb-item): add input to show items on sm screen

### LINKED ISSUE #15: Simplify breadcrumb icon naming
In the breadcrumb component the container for icons has the class `lg-breadcrumb-item__container-hide-icons`  where `container-hide-icons` is an element on the `lg-breadcrumb-item`.

As the component has been scrutinised through usage it has become apparent the name is confusing and arguably sounds closer to a modifier than an element.

A simpler name could be `lg-breadcrumb-item__content-icons` suggesting the icons are set by the user (using content projection) distinguishing it from the backward and forward icons generated by the component.

aaronBery:
BREAKING CHANGE: `showOnSmScreens` attribute needs to be applied on breadcrumb items to enable them to be shown on small screen devices

# Description

The current breadcrumb implementation does not allow the user to select which breadcrumb items should be visible on small screen devices automatically selecting what it considers to be the correct ones but this does not necessarily offer the flexibility an open source component library should provide.

In addition to this the current `hideIcons` mechanism offers little and can be removed as it only targets the first crumb item in a 2 item breadcrumb

`      crumb.hideIcons = totalCrumbCount === 2 && !index;
`

The reason being, custom icons (those created via content projection) are typically only inserted on the first item in the breadcrumb (e.g. the home icon) and (as per our examples) we only show the penultimate breadcrumb item on small screens. It is simpler and more flexible to blanket hide all icons on small screens and (when/if the requirement arises) create an input of the form `showContentIconOnSmScreens` to allow the user to select which icons they want to show on small screens.

# Screenshot
2 items tablet the content icon is visible
Screenshot 2021-01-19 at 12 46 12

2 items sm screen the content icon is hidden
Screenshot 2021-01-19 at 12 45 40

3 items tablet the content icon is visible
Screenshot 2021-01-19 at 12 46 00

3 items sm screen the content icon is hidden
Screenshot 2021-01-19 at 12 45 49

Fixes #15 

## Requirements

The existing components should behave in the same way as they currently do once the `showOnSmScreens` attribute is added to the relevant breadcrumb item. This applies to the breadcrumb component, card component, hero component and the test app.

Storybook link: 
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/components-breadcrumb--two-items
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/components-breadcrumb--three-items
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/components-breadcrumb--asynchronous-loading
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/components-card--form-journey
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/components-hero--product-details
https://deploy-preview-167--legal-and-general-canopy.netlify.app/?path=/story/test-page--standard

# Checklist:

### DIFF
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -13,7 +13,7 @@
                   Home
                 </a>
               </lg-breadcrumb-item>
-              <lg-breadcrumb-item [showItemAt]="breadcrumbBreakpoints.Small">
+              <lg-breadcrumb-item [showItemAt]="breadcrumbItemBreakpoints.Small">
                 <a href="#">Products</a>
               </lg-breadcrumb-item>
               <lg-breadcrumb-item>
@@ -288,7 +288,7 @@
                   <lg-card lgPadding="none">
                     <lg-card-header lgPadding="sm" lgPaddingBottom="xs" lgMarginBottom="lg">
                       <lg-breadcrumb lgMarginBottom="none">
-                        <lg-breadcrumb-item [showItemAt]="breadcrumbBreakpoints.Small">
+                        <lg-breadcrumb-item [showItemAt]="breadcrumbItemBreakpoints.Small">
                           <a href="#">
                             <lg-icon [name]="'chevron-left'"></lg-icon>
                             Back
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -13,7 +13,7 @@
                   Home
                 </a>
               </lg-breadcrumb-item>
-              <lg-breadcrumb-item showOnSmScreens="true">
+              <lg-breadcrumb-item [showItemAt]="breadcrumbBreakpoints.Small">
                 <a href="#">Products</a>
               </lg-breadcrumb-item>
               <lg-breadcrumb-item>
@@ -288,7 +288,7 @@
                   <lg-card lgPadding="none">
                     <lg-card-header lgPadding="sm" lgPaddingBottom="xs" lgMarginBottom="lg">
                       <lg-breadcrumb lgMarginBottom="none">
-                        <lg-breadcrumb-item showOnSmScreens="true">
+                        <lg-breadcrumb-item [showItemAt]="breadcrumbBreakpoints.Small">
                           <a href="#">
                             <lg-icon [name]="'chevron-left'"></lg-icon>
                             Back
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -13,7 +13,7 @@
                   Home
                 </a>
               </lg-breadcrumb-item>
-              <lg-breadcrumb-item>
+              <lg-breadcrumb-item showOnSmScreens="true">
                 <a href="#">Products</a>
               </lg-breadcrumb-item>
               <lg-breadcrumb-item>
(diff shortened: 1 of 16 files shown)


projects/canopy-test-app/src/app/app.component.html
```diff
@@ -13,7 +13,7 @@
                   Home
                 </a>
               </lg-breadcrumb-item>
-              <lg-breadcrumb-item>
+              <lg-breadcrumb-item showOnSmScreens="true">
```
thatguynamedandy, line 16: Just in terms of naming convention we recently adopted this approach for a table property with the intention of keeping the same convention on stacked radio buttons (cc @elenagarrone 😁)...

```
showColumnsAt | Sets the minimum screen width from which the column layout is displayed. Accepts sm, md or lg
```
I wondered wether the same convention could map to this?

e.g.
```
showItemAt | Sets the minimum screen width that the breadcrumb item is displayed. Accepts sm, md or lg
```

thatguynamedandy:
There is almost scope for a utility directive (similar to lgMargin / lgPadding) to make this re-usable across multiple components. A bit like the ones in [bootstrap](https://getbootstrap.com/docs/4.0/utilities/display/). 

e.g. `lgShowAt="lg", lgHideAt="md"`

projects/canopy-test-app/src/app/app.component.html
aaronBery: I initially set out to follow the 'showAt' pattern the table introduced but I pivoted for a few reasons. One was to focus on simplifying what we already had whilst weeding out what was redundant. The second reason was I wasn't sure if it would be overkill (as the original breadcrumb UX was very much focussed on small screen verses non small screen behaviour). That being said if there is an appetite for having a higher degree of control and avoiding further breaking changes I would be happy to look into implementing this

thatguynamedandy: Maybe seek a second opinion, I'm mostly coming from a DevEx point point of view, for consistency on the properties. The extra control and breaking change avoidance would be a bonus.

owensgit: I would advocate using the `showItemAt` approach, I think the consistency and flexibility would help us in the long run. It doesn't feel much like YAGNI, because it's achieving the same result in a more flexible way and it's creating consistently too. Not having a breaking change is also a nice to have! 

In terms of the original UX guidelines, perhaps we can address that clearly in the notes - maybe a couple of lines explaining what the guidelines are and that to achieve this, you should use `showItemAt="md"`, with an a code example. Maybe something like this:

owensgit (commented):
Thanks Aaron, nice to see some simplification! I've added my thoughts to the question around using `showItemAt` :)

owensgit:
> There is almost scope for a utility directive (similar to lgMargin / lgPadding) to make this re-usable across multiple components. A bit like the ones in [bootstrap](https://getbootstrap.com/docs/4.0/utilities/display/).
> 
> e.g. `lgShowAt="lg", lgHideAt="md"`

I think this is a great idea, I can imagine that being very useful on occasion where you just need to show/hide a component at certain breakpoints, but where it's tricky to do it via CSS (esp. with components added via content projection).

aaronBery:
I'm going to close this as adding responsive options to mirror actual use cases unearthed more work that needs to be considered for the icon mangement

PR #168, opened 2021-01-19
feat(paginator): migrate paginator module from @lgim/canopy

p-m-p:
# Description

Have been meaning to put this in as a draft PR after our initial design discussion for a while but only just getting around to it now, apologies for that.

I believe James was looking at the paginator design but I've not seen anything so the component is not currently styled. I also need to add the correct story format and even re-familiarise myself with the code so will look to make those changes plus any feedback gained from review.

Thanks!

## Requirements

# Checklist:


projects/canopy/src/lib/paginator/paginator.component.html
```diff
@@ -0,0 +1,52 @@
...
+  <select lgSelect [ngModel]="pageSize" (ngModelChange)="handleSelectPageSize($event)">
+    <option *ngFor="let size of pageSizes" [value]="size">{{ size }}</option>
+  </select>
+</lg-select-field>
+
+<span class="lg-paginator__status"
+  >{{ currentPageStart }} - {{ currentPageEnd }} of {{ length }}</span
+>
+
+<div class="lg-paginator__controls">
```
elenagarrone, line 13: From an accessibility point of you I think it would be good for this to be marked as a `navigation` and to have the buttons within a list.
Also, one comment about `aira-label` is that some screen readers don't read it so it would be better to have the label of the button within the button e.g.: `<button ...><span class="lg-visually-hidden">Show first page</span></button>`.

I think this is a good article that explains how to make a paginator accessible: https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination

I've noticed this version has only the "arrows" and no page numbers? should we add the capability of having numbers too? e.g.:

Screenshot 2021-01-19 at 17 16 46

projects/canopy/src/lib/paginator/paginator.component.scss
```diff
@@ -0,0 +1,35 @@
+.lg-paginator {
+  align-items: center;
+  display: flex;
+  justify-content: flex-end;
+
+  &__page-size {
+    align-items: center;
+    display: flex !important;
+    margin: 0 var(--space-xs) 0 0 !important;
```
elenagarrone, line 9: We try to avoid `!important`, was there a reason why you added it? can it be solved with specificity?

projects/canopy/src/lib/paginator/paginator.component.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    it('moves to last page', () => {
+      component.length = 100;
+      component.pageSize = 10;
+      component.pageIndex = 0;
+      fixture.detectChanges();
+      fixture.debugElement.query(By.css('.lg-paginator__btn-last')).nativeElement.click();
+      fixture.detectChanges();
+      expect(status.innerHTML).toMatch(`91 - 100 of 100`);
```
elenagarrone, line 78: Tiny one but I think these template quote could be changed to a regular string

projects/canopy/src/lib/paginator/paginator.component.ts
```diff
@@ -0,0 +1,282 @@
...
+   * @param cdr injected change detector ref
+   * @param lgIconRegistry injected canopy icon registry
+   * @param defaults default paginator options
+   */
+  constructor(
+    private cdr: ChangeDetectorRef,
+    lgIconRegistry: LgIconRegistry,
+    @Optional() @Inject(LG_PAGINATOR_DEFAULT_OPTIONS) defaults?: PaginationOptions,
+  ) {
+    lgIconRegistry.registerIcons([
```
elenagarrone, line 184: We usually register the icons in the component module. I don't think it changes much but maybe for consistency I would move it.

projects/canopy/src/lib/paginator/paginator.component.html
p-m-p: Thanks for that, I'll have a read.

Regarding page numbers and how it works I just modeled the component on the basic functionality for mat-paginator. I don't have much of an opinion on how it should function so open to design input.

projects/canopy/src/lib/paginator/paginator.component.scss
p-m-p: I'll have to have a look, I think most of this CSS will be junk once there is a design for the component so I probably just bodged it in

projects/canopy/src/lib/paginator/paginator.component.ts
```diff
@@ -0,0 +1,272 @@
...
+  id = `lg-paginator-${nextId++}`;
+
+  /**
+   * The lg-paginator class name added to the host element.
+   */
+  @HostBinding('class.lg-paginator')
+  class = true;
+
+  currentPageStart = 1;
+  currentPageEnd = 1;
```
owensgit, line 164: I wonder if `currentPageStart` and `currentPageEnd` could be Inputs to this component? I could imagine a scenario where you'd want to start off at a particular page, for example if the host application put the page in the URL and wants to make it possible for the user to copy and paste the URL and keep their page position. Or if there was a list of FAQs, or something like that, and the host application needed to deep-link to a specific page.

projects/canopy/src/lib/paginator/paginator.component.html
```diff
@@ -0,0 +1,52 @@
+<lg-select-field class="lg-paginator__page-size" *ngIf="!hidePageSize">
```
owensgit, line 1: I'm wondering if the page size select field should be a separate component? I guess it depends on design somewhat, but could there be a scenerio where we'd want to show this selection at the top of the page and the actual pagination buttons at the bottom? Maybe it's worth waiting to see what the designs are and what the actual use-case is within our app.

projects/canopy/src/lib/paginator/paginator.component.html
```diff
@@ -0,0 +1,52 @@
...
+  <span class="lg-paginator__page-size-label">Items per page</span>
+
+  <select lgSelect [ngModel]="pageSize" (ngModelChange)="handleSelectPageSize($event)">
+    <option *ngFor="let size of pageSizes" [value]="size">{{ size }}</option>
+  </select>
+</lg-select-field>
+
+<span class="lg-paginator__status"
+  >{{ currentPageStart }} - {{ currentPageEnd }} of {{ length }}</span
+>
```
owensgit, line 11: Perhaps we could also have a `showPaginatorStatus` input? Not sure about this one, but there could be a scenario where we have a very simple 2-page pagination where this might be clutter. Again bit of a design question too. Perhaps also for accessibility, it should always be readable by the screenreader, regardless if shown visually.

projects/canopy/src/lib/paginator/paginator.component.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  /**
+   * The list of page sizes.
+   */
+  pageSizes?: Array<number>;
+
+  /**
+   * Flag to hide the page size selector.
+   */
+  hidePageSize?: boolean;
```
owensgit, line 59: For the most part in our existing components, we tend to use `showSomething`, instead of `hideSomething` and then default it to be `true` if desired. I guess in terms of consistency, it'd be nice to keep it that way. I can see below you're using `showFirstLastButtons`.

projects/canopy/src/lib/paginator/paginator.component.ts
```diff
@@ -0,0 +1,272 @@
...
+  /**
+   * Flag to show the first and last page buttons.
+   */
+  showFirstLastButtons?: boolean;
+}
+
+/**
+ * Injection token for setting default paginator options.
+ */
+export const LG_PAGINATOR_DEFAULT_OPTIONS = new InjectionToken<PaginationOptions>(
```
owensgit, line 70: I'm not familiar with this approach of providing default options, so forgive me if I'm missing something - but it's not a pattern we've used in Canopy so far. Just wondering if it's something we need, or is just part of the component patterns in @lgim/canopy that's made its way over. If it's a something we'd like in Canopy, then perhaps we could consider it as an enhancement to the other components? Again, forgive me if I'm missing something obvious.  Happy to go with the consensus on this 👍

owensgit (commented):
Thanks @p-m-p ! Just a few questions/comments from me, some of them are perhaps just a starting point for a conversation. I'm happy to jump on a call at some point if it helps.

projects/canopy/src/lib/paginator/paginator.component.ts
p-m-p: Sure, there is the `pageIndex` input to control the selected page externally. Hopefully you can see how that works in the knobs on the story.

projects/canopy/src/lib/paginator/paginator.component.ts
p-m-p: I modeled this on the mat-paginator options https://material.angular.io/components/paginator/api#MatPaginatorDefaultOptions but am happy to go with the consensus.

projects/canopy/src/lib/paginator/paginator.component.ts
p-m-p: It's pretty common approach, you can see https://material.angular.io/components/paginator/api#MAT_PAGINATOR_DEFAULT_OPTIONS for example in Material. It just gives the opportunity to provide default settings for all paginators in an application but isn't required.

projects/canopy/src/lib/paginator/paginator.component.html
p-m-p: Sure, I refrained from putting any options in to control labels etc until there was a design decision but it's definitely worth adding that.

projects/canopy/src/lib/paginator/paginator.component.ts
owensgit: Ah I see, so `pageIndex` can set the selected page, and `currentPageStart` and `currentPageEnd` are just internal variables used in the logic?

projects/canopy/src/lib/paginator/paginator.component.html
elenagarrone: @alexcanning any opinion on this?

projects/canopy/src/lib/paginator/paginator.component.ts
owensgit: Thanks, Material has been be a very valuable source of inspiration to us too! 🙂 

In my view, it'd be good to be consistent with naming and go with `showSomething` (with a default as `true` if needed) rather than `hideSomething`. But also happy to go with consensus as well.

@thatguynamedandy  @elenagarrone  @aaronBery  @paul-request  - Do you guys have a view on this at all?

projects/canopy/src/lib/paginator/paginator.component.ts
owensgit: Thanks, it looks good. I guess I just haven't come across it yet. In my view, we should probably check in with the rest of the team, and see if it's a pattern we want to introduce to Canopy. We tend to be very keen on consistency across all components, so if we adopt it then I imagine we'd look to roll it out to other components too. 

It might be worth doing if it's an approach used in other libraries and something that future adopters of Canopy would find useful.

@thatguynamedandy I know you did a lot of research into Material in the early days of building Canopy, is this a pattern you thought about adopting at the time?

projects/canopy/src/lib/paginator/paginator.component.ts
p-m-p: Yes, I should probably encapsulate them but they are for internal use only.

p-m-p:
@andy-polhill @elenagarrone have you had any update on designs for this? I will address the outstanding points but I think there is a couple unanswered questions around use of default configuration providers and the initial configuration values.

projects/canopy/src/lib/paginator/paginator.component.ts
elenagarrone: Apologies, I completely missed the updates on this PR. I agree with @owensgit - for consistency I would use `showSomething` :)

projects/canopy/src/lib/paginator/paginator.component.ts
elenagarrone: I think we used it for the feature toggle service. It's quite a common pattern so I'm personally happy to keep it.

elenagarrone:
> @andy-polhill @elenagarrone have you had any update on designs for this? I will address the outstanding points but I think there is a couple unanswered questions around use of default configuration providers and the initial configuration values.

Apologies, I thought the designers got in contact with you to discuss the design for this 🤔  I'll chase it and let you know asap

elenagarrone (commented):
Looks really good and reads well accessibility wise 👍  I'd say the only thing I would update is the variable to use the pattern `showSomething`

projects/canopy/src/lib/paginator/paginator.component.ts
elenagarrone: Just thinking a bit more about this - maybe it would be better for this to be an input on the component in this case? Just wondering if maybe you have multiple references to that component in the same container module and you want two different default values the input might give a bit more flexibility.

p-m-p: There's inputs for each setting so you can override any defaults by using those or if you wanted you could provide the defaults at component level but seems like an unlikely use case?

PR #192, opened 2021-02-01
fix(table): update spacing styles of cells on small devices

elenagarrone:
# Description

We found a bug where on mobile long table cell labels were too close to the table cell value. In addition to that the table cell was wrapping so I've set the table cell content to never do that since the table cell value should always be short.

Before:
Screenshot 2021-02-01 at 10 47 54

After:
Screenshot 2021-02-01 at 10 48 30

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
+++ b/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
@@ -21,9 +21,14 @@
 
 .lg-table-cell__content {
   margin-left: auto;
+  white-space: nowrap;
 }
 
 .lg-table-cell--expandable-content {
+  .lg-table-cell__label {
+    padding-right: 0;
+  }
+
   .lg-table-cell__content {
     margin-left: 0;
   }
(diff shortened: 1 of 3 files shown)


thatguynamedandy (changes requested):
Code looks good, but I think it's created an issue on the `details` table example...

Screenshot 2021-02-01 at 10 54 01

elenagarrone:
> Code looks good, but I think it's created an issue on the `details` table example...
> 
> Screenshot 2021-02-01 at 10 54 01

@thatguynamedandy fixed, thank you!
Screenshot 2021-02-01 at 11 02 53

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
```diff
@@ -21,16 +21,26 @@
 
 .lg-table-cell__content {
   margin-left: auto;
+  white-space: nowrap;
+
+  .lg-table-expanded-detail {
```
thatguynamedandy, line 26: Should this selector live in here? I'm just wondering to be more future safe if it should live the scss file for the expanded detail component? (which might need creating).

Where possible I think we should try and stop the components being aware of one another.

elenagarrone: Yes, you're right 🤦  updated now

thatguynamedandy (approved):

owensgit (approved):

PR #211, opened 2021-02-11
Fix issue with Dependabot ignore list

elenagarrone:
# Description

The current `"@storybook/*"` should work but it doesn't. I've found another person with the same issue but no answer.
I've updated the current list to be more specific which hopefully should work.

In addition to that I've modified the `.gitignore` to cover the nested node modules.

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ storybook-static
 /bazel-out
 
 # dependencies
-/node_modules
+node_modules/
 /.dsm
 
 # profiling files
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,6 +13,7 @@ storybook-static
 
 # dependencies
 /node_modules
```
thatguynamedandy, line 15:
```suggestion
node_modules/
```
Would this do the same thing?

thatguynamedandy (commented):
One tiny but potential improvement.

.gitignore
elenagarrone: Yes! updated thanks 👍

owensgit (approved):
Nice one @elenagarrone 👍

andy-polhill (approved):

Issue #110, opened 2020-12-10
lgPadding directive doesn't support breakpoints


pauleustice:
The lgPadding (and related lgMargin) directives are really useful however as they doesn't support breakpoint-specific values, scss overrides are necessary to change non-desktop Canopy styles.

Perhaps the directives could take an object with values by breakpoint? e.g.

`<lg-card lgPadding="{ mobile: 'xxl' }">`

thatguynamedandy:
We had this issue in the old repo, which I think links to it. https://github.com/canopy-collective/canopy/issues/249

The proposal was this...

lgMargin and lgPadding are useful escape hatches to tweak components without adding individual classes. I think a good advancement would be able to get them to provide different spaces for difference screen sizes.

For example the below will add a medium top padding for all screen sizes..

lgMarginTop="md"

This is good but what we may want to do is have different padding for different sizes, a potential API would be something like the below..

lgMarginTop="{ 'sm': 'md', md: 'xxl', 'lg': 'xxxl' }"

This would allow us to tailor the margin/padding to the screen resolution.

owensgit:
I also like the idea of using an object to describe the breakpoints, whilst still allowing a simple string for an all-breakpoints solution, which would also avoid a breaking change.

thatguynamedandy:
Links to #49 if anyone picks it up, not a massive task but maybe one up from 'good first issue'. @pauleustice would this do what you need?

pauleustice:
Sorry both, I clearly need to tweak my notifications because I didn't see your replies! 🤦 

@thatguynamedandy Yes, that would cover it exactly. The only thing that may be confusing (and why I used `mobile` as a breakpoint key) is that we use the same sizing names for breakpoints and padding, so to me `{ sm: 'md' }` doesn't seem very intuitive.

pauleustice:
Also worth considering that the approach above is at odds with how the [https://legal-and-general-canopy.netlify.app/?path=/story/directives--grid](grid directives) work

PR #219, opened 2021-02-18
feat(promo-cards): Add promo card, promo card list components

ABucin:
# Description

Adds Promo Cards to Canopy - these are a new way of displaying promotional content on page, and as such they use the expressive font (Lyon).

This is implemented using a Promo Card List component and populating with 2 to 3 Promo Cards. Note that the promo cards themselves follow the same conventions as a standard card components - implementing a title, content, footer along with an image as separate components added via content projection slots. 

The Promo Card List handles the layout of the cards, and in future will support different layouts (probably via an Input), but for now it's simply stacked on mobile and columns on desktop.

A new variant of the separator has been added to implement the dotted line separator at the top of the list.

**Note**: There is no DSM component to link to yet, so could open an issue to add this later if needed.

## TODO
- [ ] add to the test app

## Screenshot
Screenshot 2021-02-17 at 15 02 39

## Requirements

As a new component, this should follow the same patterna and standards of existing components, and the notes should be simple for a developer to pick up and use in their project.

**Storybook link (Promo Cards):** https://deploy-preview-219--legal-and-general-canopy.netlify.app/?path=/story/components-promo-card--promo-card-list

**Storybook link (Updated separator)**: https://deploy-preview-219--legal-and-general-canopy.netlify.app/?path=/story/components-separator--standard

# Checklist:
- [] I have added the functionality to the [test app](./blob/master/docs/CONTRIBUTING.md#build-test-application)


projects/canopy/src/lib/canopy.module.ts
```diff
@@ -21,6 +21,7 @@ import { LgIconModule } from './icon/icon.module';
 import { LgPageModule } from './page/page.module';
 import { LgPipesModule } from './pipes/pipes.module';
 import { LgPrefixModule } from './prefix/prefix.module';
+import { LgPromoCardModule } from './promo-card';
```
andy-polhill, line 24:
```suggestion
import { LgPromoCardModule } from './promo-card.module';
```
There is an issue  in ng-packagr where you need to add the file extension to module imports.

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.html
```diff
@@ -0,0 +1,6 @@
+<lg-heading
+  [class]="'lg-font--expressive lg-promo-card-list-title__heading lg-font-size-5--strong'"
```
andy-polhill, line 2: I wonder if we need to add an expressive modifier to the `lg-heading` component (@owensgit)

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../../../styles/mixins';
+
+.lg-promo-card-list-title__heading {
+  margin-top: var(--space-md);
+  margin-bottom: var(--space-lg);
+
+  @include lg-breakpoint('lg') {
```
andy-polhill, line 7: This adds a bit more of a case for the work proposed in https://github.com/Legal-and-General/canopy/issues/110
Then we could do all of this with directives and have one less css file.

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list.component.html
```diff
@@ -0,0 +1,7 @@
+<lg-separator class="lg-promo-card-list__separator" variant="dotted"></lg-separator>
```
andy-polhill, line 1: I wondered if the separator should be part of this list component? Or if we should rely on people adding it manually as a sibling to the promo-card-list.

projects/canopy/src/lib/promo-card/promo-card.module.ts
```diff
@@ -0,0 +1,30 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { LgPromoCardComponent } from './promo-card/promo-card.component';
+import { LgButtonModule } from '../button';
```
andy-polhill, line 5: I'm not sure if these need `.module` adding or not, maybe this works now.

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, HostBinding, Input, ViewEncapsulation } from '@angular/core';
+
+import { HeadingLevel } from '../../../heading';
+
+@Component({
+  selector: 'lg-promo-card-list-title',
+  templateUrl: './promo-card-list-title.component.html',
+  styleUrls: ['./promo-card-list-title.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
andy-polhill, line 9:
```suggestion
  encapsulation: ViewEncapsulation.None,
  changeDetection: ChangeDetectionStrategy.OnPush,
```

projects/canopy/src/lib/separator/separator.component.scss
```diff
@@ -1,5 +1,12 @@
...
   display: block;
   margin: var(--space-lg) 0;
 }
+
+.lg-separator--solid {
+  border-top: solid var(--border-width) var(--separator-color);
+}
+
+.lg-separator--dotted {
+  border-top: dotted 0.25rem var(--color-charcoal);
```
andy-polhill, line 11: Might be worth putting the `0.25rem` in a css var, just so it's customisable/re-usable in the future.

projects/canopy/src/lib/promo-card/variables.scss
```diff
@@ -0,0 +1,2 @@
+// this has been added as a SASS variable as we cannot use CSS custom properties in media queries
+$promo-card-inline-breakpoint: 37.5rem;
```
andy-polhill, line 2: can this go in `components/_promo-card.scss`?

andy-polhill (commented):
Looking good, I only had time for a quick review as I'm not working tomorrow, but hopefully a bit of early feedback helps.

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.html
andy-polhill: Is this what we did in the standard cards? just wondering if we can avoid hardcoding classes.

projects/canopy/src/lib/promo-card/variables.scss
ABucin: I added it here so it's scoped to the `promo-card` module. I can move it to the `_promo-card.scss` if it's more manageable that way, though.

projects/canopy/src/lib/separator/separator.component.scss
ABucin: Agreed.

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.html
ABucin: The other way around would be to add a variant to the header (e.g. normal / expressive)

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list.component.html
ABucin: I think it should be added manually as some people may not want to have it there.

owensgit: Yes I agree, we could allow to be added via a content projection slot, similar to the title and cards below.

projects/canopy/src/lib/promo-card/variables.scss
owensgit: When chatting about this originally, I wasn't sure about having it here or in the variables file, as it's SASS var, not a CSS custom property, and I wasn't sure if users of Canopy would assume that you can change it because it's grouped up with the ones you can override for theming (without needing to compile any SASS). On the other hand, it's still a variable and we do have a pattern for storing vars somewhere - so maybe let's put in with the other ones to keep things consistent? I remember thinking that if it comes up in the PR we should probably move it 🙂

projects/canopy/src/lib/promo-card/promo-card.module.ts
owensgit: I think IDEs tend to omit the `.module`, wonder if we should recommend a config until `ng-packagr` fixes it

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.html
owensgit: > I wonder if we need to add an expressive modifier to the lg-heading component (@owensgit)

I was going to add a new modifier to the `lg-heading` as part of the Lyon work - e.g. `<lg-heading fontType="expressive">`, but then wanted to check in with the rest of the team as we already have a `[class]` input that allows you to add the modifier class directly along with the other ones. So it could be argued that a specified modifier is superfluous. @elenagarrone  @andy-polhill what do you think?

> Is this what we did in the standard cards? just wondering if we can avoid hardcoding classes.

Yes we did - in the standard cards we only allow the `headingLevel` to be changed:

```
<lg-heading [level]="headingLevel" class="lg-card-title__heading">
  <ng-content></ng-content>
</lg-heading>
```

We were also wondering about the hardcoding of classes - but thought it may be a bit of YAGI to make the font type customisable from `<lg-promo-card-list-title>`, and the original card title implementation swayed us :) 

However, we could add an input, e.g. `<lg-promo-card-list-title fontType="expressive">` which gets passed down the to the `lg-heading`, in a similar way to the heading level, and then also possibly add the `fontType` input to the `lg-heading` too.

Maybe we can have quick chat about this on Monday?

andy-polhill: Ah I see, I think the `class` property in lg-heading is a bit of an anomaly, it was quite an early approach. On the whole we try not to promote the adding of arbitrary classes, as it can promote people to take their own path and also create unexpected breaking changes. We could look at deprecating that property separate from this PR.

For now I think I'm ok with the classes. I have a slight niggle in that we could use the scss mixin approach instead of the hardcoded css classes, but we haven't hooked `expressive` into the mixins yet

```html
<lg-heading [level]="headingLevel" class="lg-hero-card-title__heading">
```

```scss
.lg-promo-card-list-title__heading {
  @include lg-font-size('5', 'strong');
}
```

projects/canopy/src/lib/promo-card/promo-card.notes.ts
```diff
@@ -0,0 +1,135 @@
...
+## Purpose
+
+Promo Cards allow you to arrange content in a card-like manner for promotional purposes. Promo Cards are organised by a
+Promo Card List and will be displayed in columns on desktops and stacked on mobile devices.
+
+### Content guidelines
+
+Each Promo Card should respect the following conventions:
+
+- Title - max 33 characters (with spaces)
```
andy-polhill, line 13: I wondered wether we should mention the max character count here? I don't dislike it, but we haven't done it elsewhere. This would maybe be more of a design guideline than a technical one.

projects/canopy/src/lib/promo-card/promo-card.notes.ts
```diff
@@ -0,0 +1,135 @@
...
+| \`\`lg-promo-card__img-wrapper\`\` | Adds styles to the promo card image wrapper |
+| \`\`lg-promo-card-image\`\` | Adds styles to the promo card image |
+| \`\`lg-promo-card-image__content\`\` | Adds styles to the promo card image content |
+| \`\`lg-promo-card__content-wrapper\`\` | Adds styles to the promo card content wrapper |
+| \`\`lg-promo-card__content-inner\`\` | Adds styles to the promo card inner content |
+| \`\`lg-promo-card-title\`\` | Adds styles to the promo card title |
+| \`\`lg-promo-card-title__heading\`\` | Adds styles to the promo card title inner heading |
+| \`\`lg-promo-card-content\`\` | Adds styles to the promo card content |
+| \`\`lg-promo-card-footer\`\` | Adds styles to the promo card footer |
+| \`\`lg-btn__text\`\` | Adds styles to the promo card footer CTA button |
```
andy-polhill, line 92: I think you can leave this particular line out as it's tracked in the button docs already. The more I see us writing the 'Using only the SCSS' docs, the more I think we need to automate it away somehow.

projects/canopy/src/lib/promo-card/promo-card.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import { notes } from './promo-card.notes';
+import { LgPromoCardModule } from './promo-card.module';
+import { LgGridModule } from '../grid';
+import { LgHeadingModule } from '../heading';
+import { LgButtonModule } from '../button';
+import { LgSeparatorModule } from '../separator';
+import { LgSpacingModule } from '../spacing';
+
+const cardListConfig = {
```
andy-polhill, line 12: Not sure what others think but we could hook this data up to knobs like we do for the table..
https://deploy-preview-219--legal-and-general-canopy.netlify.app/?path=/story/components-table--standard
It gives people the chance to try out different data in the component, useful for testing.

projects/canopy/src/lib/promo-card/promo-card/promo-card-image/promo-card-image.component.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.lg-promo-card-image__content {
+  width: 100%;
+  height: 20rem;
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-size: cover;
+
+  @media (min-width: $promo-card-inline-breakpoint) {
+    width: 21.5rem;
```
andy-polhill, line 12: Does this 21.5 need to stay in sync with the one here? https://github.com/Legal-and-General/canopy/pull/219/files#diff-2b27734b611e9da58de61c6ae166d208c3d986131b49342df51ba5ec9cdb56b3R8
Just wondering if these should become css variables?

ABucin: yes, I'll extract these to a variable.

andy-polhill (approved):

projects/canopy/src/lib/promo-card/promo-card.stories.ts
ABucin: Thus has been discussed and will be left as it is for the time being.

owensgit: I'm happy either way with this one, with a small inclination not to do it - for the Promo Cards it's a little different, as you're only supposed to have two or three cards in the list, and the copy and images have to meet the design requirements, so it's kinda of limited, whereas the table has a lot more possibilities (unlimited rows, empty cells etc). The table is also often populated with JSON data in real world use - whereas you could argue Promo Cards probably wouldn't be (although ironically their first outing in MA will be) 🙂 
Having said that, I'd be fine if it was done.

projects/canopy/src/lib/promo-card/promo-card.notes.ts
owensgit: Good question - we originally came up with this because Seb asked if we could enforce it in the code. We mentioned it's not a great idea to go down that route and instead rely on everyone following the guidelines. So talked about recording in it both designer and developer docs to spread the message wide. My gut feel is that it's okay to have it to keep in the dev docs, because Promo Cards are kind of an outlier - they have these quite constrictive guidelines - whereas other components are much more freeform. Happy to remove though, if it's the consensus 👍

projects/canopy/src/lib/promo-card/promo-card-list/promo-card-list-title/promo-card-list-title.component.html
owensgit: Good point about the mixin and hooking in the `expressive` font. I'll add a new issue for that and also a discussion for deprecating the `class` at some point.

owensgit: Also just created an issue for potentially deprecating the `class` input on the lg-heading: https://github.com/Legal-and-General/canopy/issues/225

owensgit (approved):

PR #229, opened 2021-02-26
fix(input): prevent empty aria-describedby if no hint is present

owensgit:
An `lg-input-field` that did not have a hint component was producing an empty `aria-describedby` on it's
child input component - which is reported as an accessibility error by Wave. To fix this, the
default `ariaDescribedBy` value on the input is set to null, which prevents an empty attribute, and
then only created when there is a value set by the hint, error, prefix or suffice components. The
`toggleIdInStringProperty` method on the `LgDomService` is also updated to only return a property if
a new element is provided and if the current property is not `null` - this also prevents a empty
string being set as the `aria-describedby` property when called before applicable elements exist.

## Requirements

- Any input field which is not implemented with a hint component should not have an empty `aria-describedby` attribute on the child input control.
- All other inputs which do have a hint component should have the correct `aria-describedby` attribute
- The `aria-describedby` attribute should update accordingly if there is a validation/error message, prefix or suffix.

Screenshot:

Screenshot 2021-02-25 at 21 48 33

# Checklist:


andy-polhill (approved):

projects/canopy/src/lib/utils/dom.service.ts
```diff
@@ -17,14 +17,22 @@ export class LgDomService {
...
-   *     // returns 'id-1 id-2'
+   *     toggleIdInStringProperty('red blue', { id: 'red' }, { id: 'green' })
+   *     // returns 'blue green'
    */
   toggleIdInStringProperty(
     property = '',
     oldElement: { id: string },
     newElement?: { id: string },
   ): string {
+    if (!newElement && property === null) {
```
paul-request, line 28: `property` can't be `null` can it? Look like it is defaulted to `''` in the function param

owensgit: This works because it allows `null` to be passed through and is then evaluated in the `if` - it doesn't turn `null` into an empty string.I think it only defaults to a empty string if `property` is `undefined`.

paul-request (approved):

PR #239, opened 2021-03-03
fix(input): fixes size of number input

andy-polhill:
# Description

A proposal for controlling input width, we have noticed that the current approach does not work for `type="number"` inputs.  Draft only at present, needs some tidying (@pauleustice, @owensgit )

✅ No breaking change
✅ Fixes input type="number"
❌ Not great CSS, would almost consider doing an inline style

## Requirements

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/input/input-field.component.scss
+++ b/projects/canopy/src/lib/forms/input/input-field.component.scss
@@ -115,4 +115,11 @@
   &:-webkit-autofill {
     -webkit-background-clip: text;
   }
+
+  // This is a bit ugly, but maybe a compromise
+  @for $var from 1 through 20 {
+    &--size-#{$var} {
+      width: #{$var + 2}rem;
+    }
+  }
 }
--- a/projects/canopy/src/lib/forms/input/input.stories.ts
+++ b/projects/canopy/src/lib/forms/input/input.stories.ts
@@ -27,6 +27,7 @@ interface KnobsConfig {
   hint?: string | null;
   icon?: string;
   iconButton?: boolean;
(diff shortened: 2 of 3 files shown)


projects/canopy/src/lib/forms/input/input-field.component.scss
```diff
@@ -115,4 +115,11 @@
   &:-webkit-autofill {
     -webkit-background-clip: text;
   }
+
+  // This is a bit ugly, but maybe a compromise
+  @for $var from 1 through 20 {
```
andy-polhill, line 120: we probably need to check the default case for when a size isn't added, and also look at the other end of the scale for when it exceeds the maximum in the iteration.
The `+ 2` is also a bit arbitrary at the moment.

owensgit:
Thanks for putting this together Andy, I like the overall approach. I'd say the CSS classes you've added are fine, but would also be happy with an inline style as well.

projects/canopy/src/lib/forms/input/input.stories.ts
```diff
@@ -27,6 +27,7 @@ interface KnobsConfig {
   hint?: string | null;
   icon?: string;
   iconButton?: boolean;
+  inputType?: 'number' | 'text' | 'tel'; // TODO add rest or make a string
```
owensgit, line 30: I would say add the remaining input types from the HTML spec rather than make a string, as it'll give us auto-completion and prevent typos.

andy-polhill:
@pauleustice is this one still relevant? I'm not sure wether to carve out some time to finish it off.

andy-polhill:
I'm going to close this down for now to keep the list clean. We can always resurrect if it becomes a high priority again.

PR #261, opened 2021-03-15
fix(data-point): layout issue in IE11 with flex shorthand

paul-request:
In IE11, `flex: 1` is internally treated as `flex: 1 1 0px` instead of `flex: 1 1 0` like other browsers which causes items to mis-align, as that means the item has a width of 0px rather than the width of it's content:

Screenshot 2021-03-15 at 17 23 14

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/data-point/data-point/data-point.component.scss
+++ b/projects/canopy/src/lib/data-point/data-point/data-point.component.scss
@@ -1,6 +1,6 @@
 .lg-data-point {
   display: flex;
-  flex: 1;
+  flex: 1 1 0;
   flex-direction: column;
   margin-bottom: var(--component-margin);
   margin-right: var(--space-sm);


elenagarrone:
@paul-request could you add a description of the issue and maybe a screenshot of how it looked before the fix?

paul-request:
Discussion point: is `flex-basis: 0` what is actually desired anyway? This means that the minimum width of the data point will be the size of it's content. With `flex-basis: auto` it will work it out based on the flex items. I think maybe `auto` would be the desired value here?

projects/canopy/src/lib/data-point/data-point/data-point.component.scss
```diff
@@ -1,6 +1,6 @@
 .lg-data-point {
   display: flex;
-  flex: 1;
```
andy-polhill, line 3: Good spot, I totally forgot this quirk, there might be some others floating around in there too.

andy-polhill (approved):

projects/canopy/src/lib/data-point/data-point/data-point.component.scss
paul-request: A quick scan shows that this is the only place we used the shorthand with the single value, so the other should be fine

owensgit (approved):

PR #262, opened 2021-03-16
fix(data-point): fix IE11 data point layout issue

paul-request:
# Description

My previous fix to change `flex: 1` to `flex: 1 1 0` worked when updating in dev tools, but when loaded  via the stylesheet it does not work as `flex-basis: 0` is treated an an invalid value as it requires a unit (like px, for example). So changing it to `auto` fixes the layout for IE11.

I suspect `flex-basis: auto` is what we wanted anyway, so that the datapoint have equal width.

# Checklist:


andy-polhill (approved):
IE11 fun times

elenagarrone (approved):

PR #281, opened 2021-03-25
fix(quick-action): use component CSS var for colour style

### LINKED ISSUE #280: Quick Action colour variable is not used correctly
The component variable `--quick-action-color` is not used in the component CSS. The component uses `--link-color`. This needs to be changed so the Quick Action colour is styled with `--quick-action-color`.

**Update:** following [a discussion on the PR](https://github.com/Legal-and-General/canopy/pull/281#discussion_r601368543), it was decided that we should remove `--quick-action-color` entirely, favouring the use of `--link-color`.

owensgit:
# Description

The component variable for Quick Action colour was not used, `--link-color` was used instead. This
changes the colour property to use `--quick-action-color`

fix #280 

## Requirements

The Quick Action colour should be easily changed using `--quick-action-color`.

Screenshot 2021-03-25 at 11 01 42

# Checklist:


projects/canopy/src/lib/quick-action/quick-action.component.scss
```diff
@@ -12,7 +12,7 @@
   border: 0;
   background-color: transparent;
   appearance: none;
-  color: var(--link-color);
+  color: var(--quick-action-color);
```
paul-request, line 15: Aren't these the same colour though?

```
:root {
  --quick-action-color: var(--link-color);
```

paul-request: Do we need a different variable in this case?

andy-polhill: I see what you mean, the extra variable could allow us to change the colour in the future (but [YAGNI](https://martinfowler.com/bliki/Yagni.html)). Annoyingly in our [semver rules it would be a breaking change](https://github.com/Legal-and-General/canopy/blob/master/docs/BREAKING_CHANGES.md) to remove that variable.

owensgit: I came across a use-case when applying a syndicate/theme. I wanted to change the Quick Action colour, but not necessarily make a global change to the `--link-color` at this stage, so it feels useful to be able to get per-component control if needed. It's also a pattern that has appeared in other components, whereby component-specific vars are just referencing global vars. So to me if feels worthwhile, happy to discuss in more detail though.

elenagarrone: I see your point @owensgit but assigning it straight to the `--link-color` will make developing a syndicated site a lot less hassle. Is your case only due to the fact that you don't want to change the other links globally?

paul-request: Should the quick action be a different colour to the global link color though? Shouldn't it just be the case that you can override the `--link-color`

owensgit: Thanks @elenagarrone @paul-request  - yes I'd usually just override `--link-color` but in this particular case I was looking to make an incremental change and reduce the scope of testing. Given that there was already a `--quick-action-color` I thought perhaps I could just change that, but then realised it wasn't hooked up. I also have seen that elsewhere in Canopy we have these component specific variables which reference global ones e.g. `--quick-action-color: var(--link-color)`. 

I do see what you mean though, do you think we should reduce our component specific variables in favour of using the global ones more directly in our components? Perhaps this warrants a proper discussion?

elenagarrone: Yes, I think that would be better going forward 🤔

owensgit: @andy-polhill  - just wondering if it would really be a breaking change to remove that variable? It's never actually been hooked up in the CSS, so it's effectively been doing nothing. So removing it would technically not break anything 🤔

andy-polhill: oh I think in that case it should be fine, it's just a `fix`

owensgit: @andy-polhill @elenagarrone @paul-request  - thanks for the above discussion, I've now removed the specific variable `--quick-action-link-color` completely - the commit now reads:

```
fix(quick-action): remove specific color variable

removes the specific color variable for the quick action, in favour of
using the global --link-color instead
```

Please take another look when you can :)

elenagarrone (approved):

paul-request (approved):

Issue #196, opened 2021-02-03
Deprecations & upcoming breaking changes (next)


owensgit:
ℹ️  &nbsp; Small breaking changes are more easily consumed when shipped together in a major version bump.

💡  &nbsp; When identifying such a change, such as removing a CSS var, add a deprecation notice and where possible provide deprecation warnings in the console.

💬   &nbsp; Add a comment to this issue to log current deprecations and upcoming minor breaking changes.

🚀  &nbsp; At an appropriate time, they can be implemented together in one major release of Canopy (with some nice release notes).

♻️  &nbsp; When released, this issue can be closed down and another one spun up for the next batch.

---

> In future, we may want to consider creating a `next` branch to ship these changes for early adoption
> If this approach works well, we may want to consider creating an issue template for it.

owensgit:
**Deprecated CSS vars**

- `--hero-bg-image`
- `--gradient-super-blue`

Marked with a CSS comment: `/* deprecated */`

Deprecated in: https://github.com/Legal-and-General/canopy/releases/tag/v5.2.0 
Original PR: https://github.com/Legal-and-General/canopy/pull/198

andy-polhill:
Once we are happy with #282 we can probably remove the above deprecations, and be proud of ourselves for avoiding one extra breaking change.

There are a couple of additional issues flagged as 'breaking change' if we have capacity to bundle them in.

Angular also have a next branch raised for Angular 12, might be worth keeping one eye on that.

elenagarrone:
@andy-polhill I'll have a look and add them to my pr.

owensgit:
Closing this as these deprecations were successfully released in one go with other breaking changes 🥳

owensgit:
Closing this as these deprecations were successfully released in one go with other breaking changes 🥳

owensgit:
**Deprecated SCSS import in the following files**

- projects/canopy/src/lib/alert/alert.component.scss
- projects/canopy/src/lib/details/details.component.scss
- projects/canopy/src/lib/forms/validation/validation.component.scss

These files currently import `variants.scss` to make the global variant classes available in these components.

These imports need to be removed and a breaking change needs to be flagged so projects know that they need to import the new variants.scss file to get the variant classes, such as `.lg-variant--info`.

PR #282, opened 2021-03-25
feat(links): update link styles with new design

elenagarrone:
# Description

Update all the links to have the new design. 

**Please when reviewing this PR could you please make sure you double check each component in storybook? I think I've covered everything but might be good to have a second pair of eyes on it.** 

This PR will be merged in the `next` branch first so that we can check if it disrupt the different applications.

Storybook link: https://deploy-preview-282--legal-and-general-canopy.netlify.app/
Design link: https://legalandgeneral.invisionapp.com/d/main#/console/19910472/416751151/preview
Screenshot:
Screenshot 2021-03-25 at 16 38 53

# Checklist:


projects/canopy/src/styles/variables/_main.scss
```diff
@@ -21,9 +21,11 @@
   /* Links */
   --link-color: var(--color-super-blue);
   --link-hover-color: var(--color-super-blue-dark);
-  --link-visited-color: var(--color-super-blue-darkest);
-  --link-focus-background-color: var(--color-black);
```
elenagarrone, line 25: I've updated `--link-focus-background-color` to `--link-focus-bg-color`. I don't think it's a breaking change because it was only called in one of the mixins that has been changed.

elenagarrone:
I should probably flag this as a breaking change even if we put it first into `next`.

projects/canopy/src/styles/variables/_main.scss
owensgit: Just wondering about the scenario where an adopter of Canopy might be overriding `--link-focus-background-color` for themeing/syndication? Would we consider that to be a breaking change? If so, we could deprecate the old one and mark it for removal in our [upcoming deprecations and breaking changes](https://github.com/Legal-and-General/canopy/issues/196).

projects/canopy/src/styles/mixins.scss
```diff
@@ -13,13 +13,72 @@
...
+    outline: 0.063rem solid $focus-bg-color;
+    outline-offset: 0;
+
+    &:hover {
+      box-shadow: none;
+    }
+  }
+}
+
+@mixin lg-revert-link-style() {
```
owensgit, line 64: I think it would be useful to provide a utility class for apps that don't use or import the SASS files. E.g. in our project we only import the main **canopy.css** file. Perhaps something like this?

```
.lg-revert-link-style {
  @include lg-revert-link-style;
}
```

owensgit (changes requested):
Looks great! Thanks Elena - just one request from me 👍

owensgit:
I just released that this might also warrant a breaking change in that any consuming apps will need to use `revert-link-style` to prevent styling clashes and miscellaneous bugs caused by the global styling of links. In our project we'll need to do this when we update Canopy after this has been released.

projects/canopy/src/styles/variables/_main.scss
elenagarrone: You mean keep using `--link-focus-background-color` but mark it as deprecated? 
I just thought because this is already a breaking change it would be easier to do it all in one go instead of creating another breaking change later on.

Happy to do whatever you and @andy-polhill think it's best.

elenagarrone:
@andy-polhill and @owensgit should I update this PR with the changes to the deprecations flagged here? https://github.com/Legal-and-General/canopy/issues/196

projects/canopy/src/styles/variables/_main.scss
owensgit: Ah yes, that's a good point :-) let's keep it in this PR then - but I think we should add a note in the commit msg in the breaking change notes for this too, so it appears in the changelog/release notes. Maybe something like this?

```
Renamed --link-focus-background-color to --link-focus-bg-color
```

As @andy-polhill mentioned in #196, we could also remove the deprecated colours from #196 as part of this commit, effectively bundling up those changes in one breaking change as well 👉   https://github.com/Legal-and-General/canopy/issues/196#issuecomment-808138461

owensgit:
> @andy-polhill and @owensgit should I update this PR with the changes to the deprecations flagged here? #196

Yes I think that would be great, sorry I mentioned that in another comment above before I saw this one 😆 
I think we would just to ensure we add notes to the commit msg for each variable deprecated or renamed 👍

andy-polhill (commented):
Looks great Elena, just a couple of comments around naming. I wondered if the commit message could be a bit more informative as well? Something like....

```
New themed link styles have been introduced.
In most scenarios these link styles should just work with pre-existing Canopy components.
Please do check for individual instances within your application.
If you need to revert to the original link style use the `lg-revert-link-style` mixin at the top of the selector.

e.g.
```scss
a {
  @include lg-revert-link-style
  
  /* ...other styles */
}
```

projects/canopy/src/styles/mixins.notes.ts
```diff
@@ -17,10 +17,30 @@ Sets the focus outline inside the element, by using multiple box shadows.
...
+
+\`\`$active-bg-color\`\`: the active background color of the element (default value: \`\`var(--link-active-bg-color)\`\`).
+
+\`\`$focus-color\`\`: the focus color color of the element (default value: \`\`var(--link-focus-color)\`\`).
+
+\`\`$focus-bg-color\`\`: the focus background color of the element (default value: \`\`var(--link-focus-bg-color)\`\`).
+
+Styles elements with the link style.
+
+### lg-revert-link-style
```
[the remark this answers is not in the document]
elenagarrone: I would go for `lg-unbranded-link`?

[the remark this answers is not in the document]
andy-polhill: ha ha that was my least favourite, but I think it's better, I just not sure if it's really branding. 🤔
`lg-unstyled-link`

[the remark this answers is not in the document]
elenagarrone: ok, I'll go with `lg-unstyled-link` 😄

elenagarrone:
@andy-polhill and @owensgit  I can't see the netlify link, any idea? 🤔

owensgit:
@elenagarrone just wondering if the commit message to remove the `LgFormsModule` could be a little more descriptive and should also include an example? I also noticed instead of `LgFormsModule` you had written `FormsModule` 🙂 

```
BREAKING CHANGE:
Remove `LgFormsModule` in favour of importing each individual form module separately.

e.g. if you are using `LgSelectDirective`, remove the import of `LgFormsModule`:
```js
imports: [LgFormsModule]
```
and import the `LgSelectDirective` module itself separately
```js
declarations: [LgSelectDirective]
```

```

**Review by owensgit — APPROVED** [review]
Great work @elenagarrone! And really nice to be able to bunch up the breaking changes like this 👍

File: projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss [hunk 4]
```diff
@@ -1,5 +1,19 @@
 @import '../../../styles/mixins';
 
+.lg-accordion__heading {
+  > h1,
+  > h2,
+  > h3,
+  > h4,
+  > h5,
+  h6 {
```
andy-polhill, line 9: Should there be a child combinator symbol `>` here too?

andy-polhill (approved):

PR #298, opened 2021-04-06
Variant directive

owensgit:
# Description

Implements `LgVariantDirective` which applies one of the five common colour variants such as "warning", "success" etc to any component. These variants are already used by current components like Alert and Validation, but it has become apparent that a directive is needed for times when we need these colour treatments elsewhere.

The rationale for this approach was agreed as part of this discussion: https://github.com/Legal-and-General/canopy/discussions/289

## Requirements

Please take a look on Storybook when deployed at the directive.

Storybook link: https://deploy-preview-298--legal-and-general-canopy.netlify.app/?path=/story/directives--variant

Design link: https://legalandgeneral.invisionapp.com/d/main/#/console/21231753/447990804/preview

Screenshot:

Screenshot 2021-04-07 at 21 18 47

# Checklist:


projects/canopy/src/lib/variant/variant.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+  selector: '[lgVariant]',
+})
+export class LgVariantDirective {
+  variantClass: string;
+  @Input()
+  set lgVariant(variant: Variant) {
+    this.variantClass = this.toggleClass(`lg-variant--${variant}`, this.variantClass);
+  }
+
+  toggleClass(newClass: string, oldClass: string): string {
```
paul-request, line 15: Does this need to be a function if it is only ever called from the input setter? I only ask as something about the name `toggleClass` and the fact it also returns the class name doesn't seem quite right. Sounds a bit nitpicky I know!

owensgit: Good point, definitely can see what you mean. Actually this function is copied from the `lgMargin` and `lgPadding` directives, which do use the function multiple times throughout the file. I was actually thinking that at some point, we should create a util method or service for this, but I wanted to open a discussion for that and agree on an approach, as we're going to be adding more of these util directives. So in this case, I thought just keeping the functionality the same as `lgMargin` and `lgPadding` would mean it would be easier to apply some common logic later.

elenagarrone: I think in this new version of Angular we can avoid doing this bit of code. I think there is an issue (or there was in canopy collective) to go through the codebase and update it.

projects/canopy/src/lib/variant/variant.notes.ts
```diff
@@ -0,0 +1,61 @@
+export const notes = `
+
+# Variant Directive
+
+## Purpose
+This directive allows you to add one of the five common colour variants to any Canopy component, such as \`\`info\`\` and \`\`success\`\`.
+
+They are already applied to existing Canopy components which use these variants out of the box, such [Alert](/?path=/story/components-alert--standard), [Detail](?path=/story/components-details--standard) and [Form Validation](?path=/story/components-form-validation--standard), but this directive allows you to use them anywhere where required.
```
elenagarrone, line 8: Should it be `such as Alert`? just wondering!

projects/canopy/src/lib/variant/variant.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+const variants = ['generic', 'info', 'success', 'warning', 'error'];
+
+@Component({
+  selector: 'lg-variant-story',
+  template: `
+    <lg-card [lgVariant]="variant" lgPaddingLeft="lg" lgPaddingRight="lg">
+      <lg-card-content>
+        <div lgContainer>
+          <div lgRow>
+            <div [lgCol]="12" [lgColMd]="8">
```
elenagarrone, line 25: These grid directives don't need the `[]`

projects/canopy/src/lib/variant/variant.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+              <p>
+                <lg-icon lgMarginRight="xs" name="warning-fill"></lg-icon>
+                <strong>The variant directive</strong>
+              </p>
+              <p>
+                This card has the <strong>{{ variant }}</strong> variant applied. Here is
+                some <a href="#">link text</a>.
+              </p>
+            </div>
+            <div [lgCol]="12" [lgColMd]="4">
```
elenagarrone, line 35: Same here.

projects/canopy/src/lib/variant/variant.directive.ts
owensgit: Ah cool, is there a new feature in Angular that does this bit of a logic for us?

projects/canopy/src/lib/variant/variant.notes.ts
owensgit: Fixed 👍

projects/canopy/src/lib/variant/variant.stories.ts
owensgit: Fixed 👍

projects/canopy/src/lib/variant/variant.stories.ts
owensgit: Fixed 👍

paul-request (approved):

projects/canopy/src/lib/variant/variant.directive.ts
elenagarrone: Yes, but I can't remember what it is :D

elenagarrone (approved):

PR #302, opened 2021-04-08
feat(skeleton): adds skeleton loader directive

andy-polhill:
# Description

A very very draft piece of work for some experimentation around loading spinners, the code is quite old so probably doesn't follow more recent best practice.

Sharing as I think @paul-request might be looking to formalise it.

Amongst the many things I haven't done on this...
- Checked if it's accessible
- Run an example with lazy loaded content.
- Created a nice animation effect

I do quite like the API though, where you just apply the directive to the text element and give it a numeric value for the expected length of the field. e.g. `lgSkeleton="5"`

https://deploy-preview-302--legal-and-general-canopy.netlify.app/?path=/story/components-skeleton--product

### DIFF
--- a/projects/canopy/src/lib/skeleton/skeleton.directive.ts
+++ b/projects/canopy/src/lib/skeleton/skeleton.directive.ts
@@ -1,0 +1,19 @@
+import { Directive, ElementRef, Renderer2, Input, HostBinding } from '@angular/core';
+
+@Directive({
+  selector: '[lgSkeleton]',
+})
+export class LgSkeletonDirective {
+  @Input('lgSkeleton') lgSkeletonWidth = 4;
+
+  @HostBinding('style.width')
+  get backgroundSize(): string {
+    return `${this.lgSkeletonWidth}em`;
+  }
+
+  constructor(private renderer: Renderer2, private hostElement: ElementRef) {
+    console.log('adds skeleton');
+    this.hostElement.nativeElement.innerHTML = 'loading...';
+    this.renderer.addClass(this.hostElement.nativeElement, 'lg-skeleton');
+  }
+}
--- a/projects/canopy/src/lib/skeleton/skeleton.module.ts
+++ b/projects/canopy/src/lib/skeleton/skeleton.module.ts
@@ -1,0 +1,11 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+import { LgSkeletonDirective } from './skeleton.directive';
+
+@NgModule({
+  imports: [CommonModule],
+  declarations: [LgSkeletonDirective],
+  exports: [LgSkeletonDirective],
+})
+export class LgSkeletonModule {}
--- a/projects/canopy/src/lib/skeleton/skeleton.stories.ts
+++ b/projects/canopy/src/lib/skeleton/skeleton.stories.ts
@@ -1,0 +1,78 @@
+import { text, withKnobs } from '@storybook/addon-knobs';
+import { moduleMetadata } from '@storybook/angular';
+
+import { LgCardModule } from '../card/card.module';
+import { LgGridModule } from '../grid/grid.module';
+
+import { LgPaddingModule } from '../spacing/padding/padding.module';
+import { LgSkeletonModule } from './skeleton.module';
+
+export default {
+  title: 'Components/Skeleton',
+  parameters: {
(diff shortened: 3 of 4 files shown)


projects/canopy/src/lib/skeleton/skeleton.directive.ts
```diff
@@ -0,0 +1,19 @@
+import { Directive, ElementRef, Renderer2, Input, HostBinding } from '@angular/core';
+
+@Directive({
+  selector: '[lgSkeleton]',
+})
+export class LgSkeletonDirective {
+  @Input('lgSkeleton') lgSkeletonWidth = 4;
```
elenagarrone, line 7: I think we should also have a way of setting the `height` as well. 
For example, this was the design of one of our skeletons:
Screenshot 2021-04-09 at 09 46 26

projects/canopy/src/lib/skeleton/skeleton.directive.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  @HostBinding('style.width')
+  get backgroundSize(): string {
+    return `${this.lgSkeletonWidth}em`;
```
elenagarrone, line 11: The issue I'm seeing with this is that you cannot have it added conditionally e.g. `[lgSkeleton]="loading ? '8' : null"`. You could check for `null` values here though.

owensgit:
Nice, I like the API too, simple and easy to understand. 

Looking at the story, I can see there are two cards, one with the directive applied and no content, and one with content. Is the intention that the "real" card will be hidden until loaded? 

If so, I wonder if it would better if the "real" card had the directive applied, and then just removed when it's ready?

I suppose the former would be flexible for different layouts and not displaying static content 🤔 

I'm also wondering about using the relative container and absolutely positioning the cards, presumably to display one and then the other? Should we encapsulate that somehow - or would that be down to the developer to handle in their project?

paul-request:
> Nice, I like the API too, simple and easy to understand. As mentioned this will come in really handy for MA soon!
> 
> Looking at the story, I can see there are two cards, one with the directive applied and no content, and one with content. Is the intention that the "real" card will be hidden until loaded?
> 
> If so, I wonder if it would better if the "real" card had the directive applied to it, hiding the content, and then just removed when it's ready?
> 
> I suppose the former would be flexible for different layouts and not displaying static content 🤔
> 
> I'm also wondering about using the relative container and absolutely positioning the cards, presumably to display one and then the other? Should we encapsulate that somehow - or would that be down to the developer to handle in their project?

I just spoke with @andy-polhill this morning, this was just there to overlay to check how the card jumps about when the content is loaded

projects/canopy/src/lib/skeleton/skeleton.directive.ts
owensgit: It looks like the styles would take up the height of the element applied to it. Are you thinking there could be instances where you need to override the height? Maybe inline elements would need it more 🤔

projects/canopy/src/lib/skeleton/skeleton.directive.ts
owensgit: I agree - it would be good to be able to apply this to a component, and then remove it, revealing content underneath when the data has loaded - probably tied to a selector/observable. Unless we decide that not to allow this by design - kind of relates my comment below :) https://github.com/Legal-and-General/canopy/pull/302#issuecomment-816532018

projects/canopy/src/lib/skeleton/skeleton.directive.ts
elenagarrone: If I have and `h1` I wouldn't set an height on that element because it would be handled by the font-size of the text. If there is no text (because it's being fetched) there is no font to resize and the h1 would basically inexistent. I'll put a screenshot of what I mean because I'm not sure I explained it very well x)

With text:
Screenshot 2021-04-09 at 10 31 05

Without text:
Screenshot 2021-04-09 at 10 31 18

andy-polhill: This is what I was trying to set automatically using the pseudo `:after` element, it was an experiment more than anything. Will be interesting to see if @paul-request managed to take it anywhere.

andy-polhill:
Closing this down for now, hopefully @paul-request got what he needed from it.

PR #305, opened 2021-04-09
feat(responsive): add lgShowAt and lgHideAt directives

owensgit:
# Description

Introduces two new directives to control responsive behaviour of components: 

`lgShowAt` - only display an element after a certain breakpoint

`lgHideAt` - hide an element after a certain breakpoint

Example:

```
<lg-card lgShowAt="md">Now you see me...</lg-card>
```

## Requirements

Both directives should work at the specified breakpoints. 

Storybook link: 

Design link: N/A

Screenshot:

Screenshot 2021-04-09 at 11 01 01

# Checklist:


projects/canopy/src/lib/repsonsive/hide-at/hide-at.directive.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { Component, DebugElement, Input } from '@angular/core';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
+import { LgHideAtDirective } from './hide-at.directive';
+
+@Component({
+  template: ` <div [lgHideAt]="lgHideAt ? lgHideAt : undefined">How you see me...</div> `,
```
elenagarrone, line 8: I think this could be written as `[lgHideAt]="lgHideAt || null"`.

I would use `null` instead of `undefined` because that's the pattern used when handling attributes in angular.

The problem with directives is that they don't handle logic like attributes, so doing:
```
lgHideAt="null"
```
will not translate in the directive not being added like for a regular attribute.

Instead you should add the logic to say if the `responsiveUtilClass` is defined then add the new class otherwise do nothing.

If you check your code as it is now you most probably will see that if you set the `lgHideAt` as `undefined` or `null` the code will still run and add `lg-hide-at--undefined` (which works simply because that class doesn't exist).

projects/canopy/src/lib/repsonsive/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './show-at/index';
```
elenagarrone, line 1: I'm wondering if these should be part of the grid module and same folder since they are strictly related.

projects/canopy/src/styles/responsive.scss
```diff
@@ -0,0 +1,17 @@
+@import 'mixins';
```
elenagarrone, line 1: Same as the other comment, should this go in the grid?

elenagarrone (changes requested):

projects/canopy/src/lib/repsonsive/index.ts
owensgit: On one hand it does make sense to be part of the grid module, but then again - you might want to use these responsive directives without necessarily using the grid, in that sense maybe it's not a good to couple them together like that. Also, you may use the grid in your app, but never use these directives 🤔 

Alternatively, we could see them as another util directive and put them in the **/util-directives** folder. I was seeing them similar to Bootstrap's display utility classes. What do you think?

Would be great to get some more perspectives CC @andy-polhill  @paul-request  @aaronBery

projects/canopy/src/lib/repsonsive/hide-at/hide-at.directive.spec.ts
owensgit: Thanks for spotting that @elenagarrone - good call! 

I've update the code and added a test for it too, let me know what you think.

projects/canopy/src/lib/repsonsive/hide-at/hide-at.notes.ts
```diff
@@ -0,0 +1,54 @@
...
+# Hide At Directive
+
+## Purpose
+This directive allows for media quieries to be added to components, so they can be hidden at the desired breakpoint, without the need to write additional CSS.
+
+For example, you may need to hide a component on wider viewports, such as desktop.
+
+It utilises global CSS responsive utility classes to add the relevant classes.
+
+It has a brother: \`\`LghideAt\`\`.
```
elenagarrone, line 11: `LghideAt ` => `LgHideAt `

projects/canopy/src/lib/repsonsive/hide-at/hide-at.notes.ts
```diff
@@ -0,0 +1,54 @@
+export const notes = `
```
elenagarrone, line 1: Maybe we should write something in the grid notes to say these directives exist and add a link to these notes?

projects/canopy/src/lib/repsonsive/index.ts
elenagarrone: Yes, that's a good point. 
I think they should go into `/util-directives`.

projects/canopy/src/lib/repsonsive/repsonsive.module.ts
```diff
@@ -0,0 +1,10 @@
+import { NgModule } from '@angular/core';
```
elenagarrone, line 1: I'm not too sure about adding the responsive module, I would just keep them separate 🤔  
The two directives are useful but I don't see them being used too often in a codebase and most of the time you will use either one or the other in a module/component.

projects/canopy/src/styles/utils.scss
```diff
@@ -3,3 +3,24 @@
...
+  // 1024
+  'lg': 64rem,
+  // 1280
+  'xl': 80rem,
+  // 1440
+  'xxl': 90rem
+) !default;
+
+@each $name, $bp in $breakpoints {
+  .lg-show-at-#{$name} {
```
elenagarrone, line 21: What about the `.lg-hide-at-#{$name}`?

elenagarrone: Oh I'm a bit confused why this is here and in the `responsive.scss`? is that a mistake or am i missing something?

projects/canopy/src/lib/canopy.module.ts
```diff
@@ -36,6 +37,7 @@ const modules = [
   LgAlertModule,
   LgBreadcrumbModule,
   LgButtonModule,
+  LgBrandIconModule,
```
andy-polhill, line 40: Always bonus points for alphabetical ordering 🥳

projects/canopy/src/lib/repsonsive/index.ts
andy-polhill: I can see them being quite multi purpose, personally I'm happy for them to sit in their own module.

projects/canopy/src/lib/repsonsive/show-at/show-at.notes.ts
```diff
@@ -0,0 +1,54 @@
+export const notes = `
+# Show At Directive
+
+## Purpose
+This directive allows for mobile-first media quieries to be added to components, so they can be shown at the desired breakpoint, without the need to write additional CSS.
```
andy-polhill, line 5:
```suggestion
This utility directive allows for mobile-first media queries to be added to components, so they can be shown at the desired breakpoint, without the need to write additional CSS.
```

projects/canopy/src/lib/repsonsive/show-at/show-at.notes.ts
```diff
@@ -0,0 +1,54 @@
...
+
+For example, you may need to show a component only when the viewport is wide enough, e.g. desktop.
+
+It utilises global CSS responsive utility classes to add the relevant classes.
+
+It has a sister: \`\`LgHideAt\`\`.
```
andy-polhill, line 11:
```suggestion
It has a sibling directive: \`\`LgHideAt\`\`.
```

projects/canopy/src/styles/utils.scss
```diff
@@ -3,3 +3,24 @@
 .lg-visually-hidden {
   @include lg-visually-hidden;
 }
+
+$breakpoints: (
```
andy-polhill, line 7: Do these also still exist in the mixins.scss file? (which is arguably an odd place for them originally) I just wanted to make sure they weren't repeated.

andy-polhill (commented):
Nice work @owensgit just a few easy improvements / discussion points.

projects/canopy/src/lib/repsonsive/hide-at/hide-at.notes.ts
owensgit: Good idea 👍

projects/canopy/src/lib/repsonsive/repsonsive.module.ts
owensgit: I was a bit on the fence about this too but leaning a bit toward having them available as one module, partly because it's consistent with the `SpacingModule` and partly because it would be convenient to choose to have them imported together or separate. Having said that, I do kinda agree with your points too 🙂 

@andy-polhill  - what do you think?

projects/canopy/src/styles/utils.scss
owensgit: Oh whoops, was meant to remove that, thanks!

projects/canopy/src/lib/repsonsive/hide-at/hide-at.notes.ts
owensgit: Done ✅

projects/canopy/src/styles/utils.scss
owensgit: Ah yes they do - was meant to remove this stuff as it was an early thing - I've removed this bit

projects/canopy/src/lib/repsonsive/show-at/show-at.notes.ts
owensgit: Done ✅

projects/canopy/src/styles/utils.scss
owensgit: Removed ✅

projects/canopy/src/lib/repsonsive/hide-at/hide-at.notes.ts
owensgit: Added some notes here: https://github.com/Legal-and-General/canopy/pull/305/files#diff-9d96f01f73d60d459b090218b5baf1d53eb217acc3a66a143ca28476dad8ab71

projects/canopy/src/lib/repsonsive/repsonsive.module.ts
elenagarrone: Mmm I'm actually more keen to also remove the spacing module :D i personally only always add each module separately 🤔

andy-polhill: I think I'm inclined to trend in the same direction as @elenagarrone on this one. I think the aggregating modules are maybe a bit of an anti-pattern.

andy-polhill (approved):

projects/canopy/src/lib/repsonsive/repsonsive.module.ts
owensgit: I can see the logic in that :) I'll remove the responsive module and keep them separate 👍

projects/canopy/src/lib/shared/breakpoints.interface.ts
```diff
@@ -0,0 +1 @@
+export type Breakpoints = 'sm' | 'md' | 'lg' | 'xl' | 'xxl';
```
owensgit, line 1: @elenagarrone @andy-polhill - I've removed the responsive module which combined `showAt` and `hideAt`, but wanted to put this interface somewhere - it's used by both directives. First thought was this new `src/lib/shared` directory. What do you think?

projects/canopy/src/lib/repsonsive/repsonsive.module.ts
owensgit: This is done now ✅

projects/canopy/src/lib/shared/breakpoints.interface.ts
andy-polhill: makes sense 👍  works for me.

projects/canopy/src/lib/hide-at/hide-at.directive.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { Breakpoints } from '../shared/breakpoints.interface';
+
+@Directive({
+  selector: '[lgHideAt]',
+})
+export class LgHideAtDirective {
+  responsiveUtilClass: string;
+  @Input()
+  set lgHideAt(breakpoint: Breakpoints) {
```
andy-polhill, line 11: You could trial [this approach](https://github.com/Legal-and-General/canopy/discussions/303#discussioncomment-598758). I realise that it would be a shift in approach and I wouldn't hold PR up for it.

projects/canopy/src/lib/hide-at/hide-at.notes.ts
```diff
@@ -0,0 +1,54 @@
+export const notes = `
+# Hide At Directive
+
+## Purpose
+This directive allows for media quieries to be added to components, so they can be hidden at the desired breakpoint, without the need to write additional CSS.
+
+For example, you may need to hide a component on wider viewports, such as desktop.
+
+It utilises global CSS responsive utility classes to add the relevant classes.
```
andy-polhill, line 9:
```suggestion
It uses global CSS responsive utility classes to add the relevant classes.
```
v minor

projects/canopy/src/styles/grid.notes.ts
```diff
@@ -27,4 +27,8 @@ If IE11 support is required you will need to polyfill CSS variable functionality
 This can be done via [css-vars-ponyfill](https://www.npmjs.com/package/css-vars-ponyfill) or similar.
 
 A set of [Angular directives](/?path=/story/directives--grid) are also supplied which can be used to add grid classes to native HTML and Canopy elements.
+
+There are also two util directives available which allow you to show/hide components at different breakpoints:
```
andy-polhill, line 31:
```suggestion
There are also two utility directives available which allow you to show/hide components at different breakpoints:
```
v minor again

andy-polhill (commented):
Some super minor comments, the only thing holding back from a ✅  is the [question regarding module imports](https://github.com/Legal-and-General/canopy/pull/305#discussion_r613969881).

projects/canopy/src/lib/canopy.module.ts
```diff
@@ -17,13 +17,15 @@ import { LgGridModule } from './grid/grid.module';
...
 import { LgHeroModule } from './hero/hero.module';
+import { LgHideAtModule } from './hide-at';
 import { LgIconModule } from './icon/icon.module';
 import { LgPageModule } from './page/page.module';
 import { LgPipesModule } from './pipes/pipes.module';
 import { LgPrefixModule } from './prefix/prefix.module';
 import { LgPromoCardModule } from './promo-card/promo-card.module';
 import { LgQuickActionModule } from './quick-action/quick-action.module';
 import { LgSeparatorModule } from './separator/separator.module';
+import { LgShowAtModule } from './show-at';
```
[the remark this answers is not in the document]
owensgit: Ah yes, thanks for the reminder, I guess we have to be careful with the auto imports until we know it's resolved.

I've update these new ones in the meantime ✅

projects/canopy/src/lib/hide-at/hide-at.directive.ts
owensgit: Thanks, yes I'd prefer to get this one in if that's okay. That discussion was born from working on this branch where I've started to wonder if we could improve it 👍

projects/canopy/src/lib/hide-at/hide-at.notes.ts
owensgit: Done ✅

projects/canopy/src/styles/grid.notes.ts
owensgit: Done ✅

andy-polhill (approved):

elenagarrone (approved):
Apologies it took me so long to review! Looks good 👍

PR #313, opened 2021-04-12
feat(side-nav): add side-nav menu

ABucin:
# Description

This is a draft PR of the side navigation component, which is an offshoot of the tab navigation component. The functionality is similar (still allows you to switch between sections), but the behaviour is a bit different as the nav menu is on the left of the content on desktop and below it on mobile. In contrast to regular tabs, side-nav tabs have a title and an optional description.

This draft PR lacks the following:

1. complete HTML in the story
2. complete code in the notes
3. functionality added to test app

## Requirements

Storybook link: https://deploy-preview-313--legal-and-general-canopy.netlify.app/?path=/story/components-side-nav--standard

Design: https://legalandgeneral.invisionapp.com/overview/Account-Details-cknlj6bn71jia01ym13rjlqh8/screens

Screenshots: 

### Desktop

Screenshot 2021-04-12 at 16 47 43

### Tablet

Screenshot 2021-04-12 at 16 51 13

### Mobile phone

Screenshot 2021-04-12 at 16 50 27

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/canopy.module.ts
+++ b/projects/canopy/src/lib/canopy.module.ts
@@ -24,6 +24,7 @@ import { LgPrefixModule } from './prefix/prefix.module';
 import { LgPromoCardModule } from './promo-card/promo-card.module';
 import { LgQuickActionModule } from './quick-action/quick-action.module';
 import { LgSeparatorModule } from './separator/separator.module';
+import { LgSideNavModule } from './side-nav/side-nav.module';
 import { LgSpacingModule } from './spacing/spacing.module';
 import { LgSpinnerModule } from './spinner/spinner.module';
 import { LgSuffixModule } from './suffix/suffix.module';
@@ -55,6 +56,7 @@ const modules = [
   LgPromoCardModule,
   LgQuickActionModule,
   LgSeparatorModule,
+  LgSideNavModule,
   LgSpacingModule,
   LgSpinnerModule,
   LgSuffixModule,
--- a/projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.html
+++ b/projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.html
@@ -1,0 +1,9 @@
+<div class="lg-side-nav-bar-item__container">
+  <div class="lg-side-nav-bar-item__title lg-font-size-1--strong">{{ title }}</div>
+  <div *ngIf="description" class="lg-side-nav-bar-item__description lg-font-size-0-8">
+    {{ description }}
+  </div>
+</div>
+<div class="lg-side-nav-bar-item__icon-container">
+  <lg-icon name="chevron-right" class="lg-font-size-0-8"></lg-icon>
+</div>
--- a/projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.ts
+++ b/projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.ts
@@ -1,0 +1,14 @@
+import { Component, HostBinding, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+  selector: 'lg-side-nav-bar-item',
+  templateUrl: './side-nav-bar-item.component.html',
+  styleUrls: ['./side-nav-bar-item.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LgSideNavBarItemComponent {
+  @HostBinding('class.lg-side-nav-bar-item') class = true;
+
+  @Input() title: string;
+  @Input() description: string;
+}
--- a/projects/canopy/src/lib/side-nav/side-nav-bar/side-nav-bar.component.ts
+++ b/projects/canopy/src/lib/side-nav/side-nav-bar/side-nav-bar.component.ts
@@ -1,0 +1,116 @@
+import {
+  AfterContentChecked,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  ContentChildren,
+  forwardRef,
+  HostBinding,
+  HostListener,
+  Input,
+  OnDestroy,
+  QueryList,
+  ViewEncapsulation,
+} from '@angular/core';
+
+import { merge, Subject } from 'rxjs';
+import { takeUntil } from 'rxjs/operators';
+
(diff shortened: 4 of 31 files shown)


projects/canopy/src/lib/canopy.module.ts
```diff
@@ -24,6 +24,7 @@ import { LgPrefixModule } from './prefix/prefix.module';
 import { LgPromoCardModule } from './promo-card/promo-card.module';
 import { LgQuickActionModule } from './quick-action/quick-action.module';
 import { LgSeparatorModule } from './separator/separator.module';
+import { LgSideNavModule } from './side-nav/side-nav.module';
```
andy-polhill, line 27: There is a bit of an open question about names, but we can potentially ignore that for now.

projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="lg-side-nav-bar-item__container">
+  <div class="lg-side-nav-bar-item__title lg-font-size-1--strong">{{ title }}</div>
```
andy-polhill, line 2: Initially I thought this should be a header for a11y reasons, but I notice that isn't the case for the current tabs, so maybe not. @elenagarrone?

projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.ts
```diff
@@ -0,0 +1,14 @@
+import { Component, HostBinding, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+  selector: 'lg-side-nav-bar-item',
+  templateUrl: './side-nav-bar-item.component.html',
+  styleUrls: ['./side-nav-bar-item.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
andy-polhill, line 7: Most of the components should probably be switched to `changeDetection: ChangeDetectionStrategy.OnPush`

projects/canopy/src/lib/side-nav/side-nav-bar/side-nav-bar.component.ts
```diff
@@ -0,0 +1,114 @@
...
+import { LgSideNavBarLinkDirective } from '../side-nav-bar-item/side-nav-bar-link.directive';
+import { isKeyDown, isKeyLeft, isKeyRight, isKeyUp } from '../../utils/keyboard-keys';
+
+@Component({
+  selector: 'lg-side-nav-bar',
+  templateUrl: './side-nav-bar.component.html',
+  styleUrls: ['./side-nav-bar.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LgSideNavBarComponent implements AfterContentChecked, OnDestroy {
```
andy-polhill, line 27: There is obviously quite a lot of repetition here between this and the [tab-nav-bar](https://github.com/Legal-and-General/canopy/blob/master/projects/canopy/src/lib/tabs/tab-nav-bar/tab-nav-bar.component.ts) component. I wonder if there is a way of creating a more generic approach that works for both, `side-nav` and `tab-nav` are almost variants of the same thing.

```ts
selector: 'lg-side-nav-bar, 'lg-side-nav-bar'',
```

We did set a precedent on the [radio component](https://github.com/Legal-and-General/canopy/blob/master/projects/canopy/src/lib/forms/radio/radio-button.component.ts) where it has the ability to to behave differently, I wonder if this is similar.

projects/canopy/src/lib/side-nav/side-nav.model.ts
```diff
@@ -0,0 +1,4 @@
+export interface SideNavBarItem {
```
andy-polhill, line 1: We normally adopt the convention where the file would be `side-nav.interface.ts` and contain all the interfaces, types etc for that module.

andy-polhill (commented):
Looking good @ABucin I only skimmed it for now, I think the only contentious comment would be the one around reducing the repetition.

projects/canopy/src/lib/side-nav/side-nav.model.ts
ABucin: Fixed

projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.ts
ABucin: Fixed.

projects/canopy/src/lib/side-nav/side-nav-bar/side-nav-bar.component.ts
ABucin: I think the main thing that sets them apart is the fact that the sidebar content + nav are oriented differently based on screen width (horizontal on desktop, vertical on mobile). Not sure if this is easily achievable with the current tab implementation.

andy-polhill: I wondered if there is an opportunity to make the component code common, but have separate css files for the side / tab navigation approaches? I think if there was to be lots of code branching in the component this might get ugly, but if most of it is presentational then it could make it more DRY. Happy to offer a bit of support to help on it if needed?

ABucin: Sure, I think it's best we have a three amigos first with you, me, and Owen so we can outline the plan for this.

andy-polhill: Sounds good, are you ok to schedule something in? Shout if you can't find a time and I can re-arrange things. @owensgit is this all ok from your side?

projects/canopy/src/lib/side-nav/side-nav-bar-item/side-nav-bar-item.component.html
elenagarrone: Yes, I would definitely have it as a heading

elenagarrone: And i would also add an `aria-describedby` to the description connecting it to the title

elenagarrone: Ignore my comments above as I'm happy with what we've talked about on the call.

PR #314, opened 2021-04-12
Fix some issues with the alert and the variants

elenagarrone:
# Description

Add back the margin to the alert icon (removed by mistake https://github.com/Legal-and-General/canopy/commit/3f0e52c800e7f53e011fdc245b80653b47fd0094) and fixes the breaking change introduced in `5.8.0` where a new `variants.scss` was introduced which requires adding the import in every project using `scss` files. For the moment we will import the file in each component that uses it but flagged as deprecated. The breaking change will be addressed in the `next` branch.

Storybook link: https://deploy-preview-314--legal-and-general-canopy.netlify.app/

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/details/details.component.scss
+++ b/projects/canopy/src/lib/details/details.component.scss
@@ -1,4 +1,9 @@
 @import '../../styles/mixins';
+/*
+ The below import will be removed in the next breaking change. It is currently included to ensure
+ the same scss entities are exported. @deprecated
+*/
+@import '../../styles/variants';
 
 .lg-details {
   background: var(--details-bg-color);
--- a/projects/canopy/src/lib/forms/validation/validation.component.scss
+++ b/projects/canopy/src/lib/forms/validation/validation.component.scss
@@ -1,4 +1,9 @@
-@import '../../../styles/mixins.scss';
+@import '../../../styles/mixins';
+/*
+ The below import will be removed in the next breaking change. It is currently included to ensure
+ the same scss entities are exported. @deprecated
+*/
+@import '../../../styles/variants';
 
 .lg-validation {
   display: flex;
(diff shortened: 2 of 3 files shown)


projects/canopy/src/lib/details/details.component.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../styles/mixins';
+@import '../../styles/variants.scss';
```
andy-polhill, line 2: It's rare that I condone comments, but I think this could do with one. Without knowing the reasoning it looks like this is a redundant import (does it need the file extension?).

Do we include it for now but tag it as an eventual breaking change? With the view to bringing in that import another way? (finding a nice way might be tricky)

elenagarrone: Sure, I can add a comment and remove the `scss` extension.
My plan is once this gets merged i'll rebase `next` and revert these changes so that it can go in as a breaking change with all the rest.

projects/canopy/src/lib/forms/validation/validation.component.scss
```diff
@@ -1,4 +1,6 @@
-@import '../../../styles/mixins.scss';
+@import '../../../styles/mixins';
+// The below import will be removed in the next breaking change
```
andy-polhill, line 2:
```suggestion
/*
 The below import will be removed in the next breaking change. It is currently included to ensure
 the same scss entities are exported. @deprecated
*/
```
Not the best wording but a bit more context might help the next person, added the @deprecated tag too just to help track it.

andy-polhill (dismissed):

projects/canopy/src/lib/forms/validation/validation.component.scss
elenagarrone: Updated 👍

andy-polhill (approved):

owensgit (approved):

PR #334, opened 2021-04-26
feat(skeleton): adds skeleton loader directive 💀

paul-request:
The initial goal was to create a directive that would automatically add a skeleton loading state to an element/component. In order to know when content has changed, I used a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to listen for changes to the host element's content.

This works well to some extent, but has the following limitations:

1. It doesn't work for components where content is projected into child components as Angular injects a bunch of stuff which makes it difficult to tell what is content and what is not. (Perhaps some logic could be added to strip out comments, but still seems like it is hard to know for sure).
2. Uses CSS pseudo to add an `&nbsp;` to the host element, so that it has dimensions, which means by default it can mirror the component/element dimensions. Without this, the consumer would have to use the inputs to manually specify the dimensions of every skeleton state.
3. I've added a couple of cards and a data point to demo it in the stories, but it may be challenging getting it to work for all  other components.

I have tried to use it on a more complex component that makes use of multiple Canopy child components, but because of the limitation of only working on items with direct projected content, I ended up creating some static HTML to mirror the components and used the directive on that to achieve skeleton loading states instead. Something built purely using Canopy components, may however, be better suited to use this as is.

# Checklist:


projects/canopy/src/lib/skeleton/skeleton.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+  };
+}
+
+@Component({
+  selector: 'async-skeleton-loading-card',
+  template: `
+    <lg-card>
+      <lg-card-header>
+        <lg-card-title headingLevel="4">
+          <span lgSkeleton lgSkeletonWidth="20">{{ data?.title }}</span>
```
paul-request, line 32: This is an example where the `lg-card-title` component projects the content into another child component, which won't work. So I have instead wrapped the content in a span and added the directive to that.

projects/canopy/src/styles/styles.scss
```diff
@@ -7,3 +7,62 @@
...
+  max-width: 100%;
+}
+
+.lg-skeleton--right {
+  align-self: flex-end;
+  text-indent: 9999em;
+}
+
+.lg-skeleton:before {
+  content: '\200B';
```
paul-request, line 33: This is the `&nbsp;` I mentioned that gives the element dimensions so that the consumer wouldn't have to manually specify height/width.

projects/canopy/src/styles/styles.scss
```diff
@@ -7,3 +7,62 @@
...
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  transform: translateX(-100%);
+  background-size: 100% 100%;
+  background-repeat: no-repeat;
+  background-color: transparent;
+  background-image: linear-gradient(
```
paul-request, line 46: @alexcanning have you ever looked at designing a colour/animation for a potential skeleton loading state before? This is just one I come up with for demo purposes but can be replaced with something more on-brand if available?

projects/canopy/src/lib/skeleton/skeleton.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.changes = new MutationObserver(this.observeChanges);
+
+    this.changes.observe(this.hostElement.nativeElement, {
+      subtree: true,
+      characterData: true,
+    });
+  }
+
+  observeChanges = (mutations: Array<MutationRecord>) => {
+    mutations.forEach((mutation: MutationRecord) => {
```
andy-polhill, line 46: I wondered if we could set a boolean property for `hasContent` or similar. Then the behaviour for removing the classes could be done in the host binding as well. Could be a bit less code?
```ts
mutations.forEach((mutation: MutationRecord) => {
  this.hasContent = mutation.target.textContent.trim() !== ''
}); 
```


```ts
get alignClass() {
  return this.lgSkeletonRightAlign && !this.hasContent;
}
```

projects/canopy/src/lib/skeleton/skeleton.stories.ts
andy-polhill: Might be an ok limitation if it means we can start using this now, do you think it's worth any further investigation? another pair of eyes?

andy-polhill (commented):
I'd be up for trying this out, I think the API is good, so if the implementation did have to change it would be ok.

andy-polhill:
Screenshot 2021-05-10 at 13 22 40
 I wonder if we should reduce the height a little bit? I guess on auto it takes the height of the tallest character?

andy-polhill:
API wise I also wondered if we need `lgSkeletonWidth` or if we could just have `lgSkeleton`. So a minimal usage example would just be.. `lgSkeleton="20"`

projects/canopy/src/lib/skeleton/skeleton.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+  private changes: MutationObserver;
+
+  @Input() lgSkeletonWidth: string = null;
+  @Input() lgSkeletonHeight: string = null;
+  @Input() lgSkeletonRightAlign = false;
+
+  @HostBinding('class.lg-skeleton') skeletonClass = true;
+  @HostBinding('class.lg-skeleton--right')
+  get alignClass() {
+    return this.lgSkeletonRightAlign;
```
andy-polhill, line 23: Not sure if this adds complexity but you might be able to assign this automatically.
```ts
  @HostBinding('class.lg-skeleton--right')
  get alignClass() {
    return window
      .getComputedStyle(this.hostElement.nativeElement)
      .getPropertyValue('text-align') === 'right';
```

projects/canopy/src/lib/skeleton/skeleton.stories.ts
paul-request: I think it's ok, I had an initial attempt at making a product card in our app, and because of the ways its built with container components, I think the easiest implementation is possibly just replicating the card structure as a separate component and using this directive to add the loading states to get the desired effect.

projects/canopy/src/lib/skeleton/skeleton.directive.ts
paul-request: Yeah, I like that. Will make that change

projects/canopy/src/lib/skeleton/skeleton.directive.ts
paul-request: Yeah I did think about it... but its not just `text-align`, could be a flex item that is right-aligned with margin etc. So this gives the consumer the flexibility to use any layout mechanism and still get the alignment to the right, if needed.

projects/canopy/src/lib/skeleton/skeleton.stories.ts
andy-polhill: I did have a little look to get my head around it more. There might be a way to target the class on a child node, but for now I feel like it might be a nice minimal solution. If the limitation poses a wider problem we could invest more effort in it.

paul-request:
> API wise I also wondered if we need `lgSkeletonWidth` or if we could just have `lgSkeleton`. So a minimal usage example would just be.. `lgSkeleton="20"`

I did consider that, but I actually thought the default API of `lgSkeleton` would be the more commonly used one, where the hight and width is just 100% of the parent dimensions. Then the consumer can use the optional height/width inputs to adjust if necessary. I guess we would need input from @alexcanning on that though 🤷

paul-request:
> Screenshot 2021-05-10 at 13 22 40
> I wonder if we should reduce the height a little bit? I guess on auto it takes the height of the tallest character?

I'm not sure, that text in the card is the value which does have a larger font size... so seems to make sense. If it were smaller, maybe it would jump around more. Its trivial to override with the `lgSkeletonHeight` input though 🤔

projects/canopy/src/lib/skeleton/skeleton.directive.ts
andy-polhill: think this one might need resolving still, if you think it's a go-er?

projects/canopy/src/lib/skeleton/skeleton.directive.ts
```diff
@@ -0,0 +1,62 @@
+import { Directive, ElementRef, Input, HostBinding, OnDestroy } from '@angular/core';
+
+@Directive({
+  selector: '[lgSkeleton]',
+})
+export class LgSkeletonDirective implements OnDestroy {
+  private changes: MutationObserver;
+  private hasContent = false;
+
+  @Input() lgSkeletonWidth: string = null;
```
elenagarrone, line 10: Could you not leave the default values unspecified as they are set to be falsey anyway?

elenagarrone (approved):

andy-polhill (approved):

Issue #49, opened 2020-10-28
Add specificity to utility directive styles


pauleustice:
When using LgPadding and LgMargin, I've found that the styles applied by the directives can be overridden as they have low specificity. Perhaps they could make use of `!important` to overcome this?

owensgit:
@pauleustice thanks for raising this. My instinctual reaction with `!important` is that it should be avoided. But perhaps this is case where it's actually appropriate 🤔 

Looking at the CSS, I can see the utility classes are specified inside the `body`:

```css
body {
  // utility classes require additional specificity
  ...
  @each $val in $spacing-list {
    .lg-margin__top--#{$val} {
      margin-top: space($val);
    }
  ...
  }
}
```

Alternatively, maybe they could be a level higher, and placed inside the `html` to give them even higher specificity.

Could you provide an example/scenario where they are being overridden, it'll be good to figure our why the `body` CSS above is not specific enough.
_(please note that we can't share specific details about our digital products in this repo as it's open source._ 

It'll also be good to get the views of the rest of the team CC @elenagarrone  @thatguynamedandy  @aaronBery.

thatguynamedandy:
Ah this is an interesting one, @pauleustice are they being overridden by other Canopy classes? or is it pre-existing css in the app that is overriding them? I think in a Canopy only app the current approach should work, but maybe not so well in a more hybrid situation.

Using `!important` is a good talking point, like @owensgit my instinct is to avoid like the plague as that is what is drilled into us, hence my slightly hacky use of `body` to buy a bit of extra specificity, but maybe using `!important` for a utility class is actually a valid use case as it should always win any specificity war, does anyone else have any thoughts / downsides?

pauleustice:
Hi both, sorry it's taken me so long to see your replies! The directive styles were being overridden by other scss within my component, not Canopy-related. I think that utility classes (where the user is explicitly stating that they want to apply a particular style) are the only real-world scenario where you _should_ use `!important`.

thatguynamedandy:
I'm in agreement with that @pauleustice, converting this issue to an enhancement, there is still time to put a case to the contrary if anyone disagrees.

aaronBery:
Not that we should need to validate our decisions with blog posts,but given the stigma around the use of the `!important` I thought it might be handy to point to a CSS tricks article by Chris Coyier where he describes using the`!important rule` in utility classes (as @pauleustice mentioned) as one of the valid uses cases.

[...]

PR #336, opened 2021-04-27
fix(button): ensure link styles don't override a link styled as button

owensgit:
# Description

~Th `lg-link` mixin used `!important` to apply colours to links inside
one of the `lg-variant` classes (such as info, success, error, etc).
However, this caused issues with the `lg-button` directive when used to
style a link as a button, because the global link styles use `lg-link`.
This is fixed by making `!important` an opt in paramter of the `lg-link`
mixin, so that the variant styles can still use it and apply `!important`,
but that won't affect any other uses of the `lg-link` mixin.~

**Edit:**

Following the discussion on the use of `!important` for the links inside `lg-variant`, we have decided to remove `!important` completely from the `lg-link` mixin.

This change also edits the button story, so that it displays both a normal button and a link styled as a button (separate commit).

Storybook link: https://deploy-preview-336--legal-and-general-canopy.netlify.app/?path=/story/components-button--primary-button

**Before:**
Screenshot 2021-04-27 at 14 57 37

**After:**
Screenshot 2021-04-27 at 14 57 04

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/button/button.stories.ts
+++ b/projects/canopy/src/lib/button/button.stories.ts
@@ -26,6 +26,7 @@ const contentGroupId = 'content';
 @Component({
   selector: 'lg-button-story',
   template: `
+    <p>Used on a <strong>button</strong> element</p>
     <button
       lg-button
       [disabled]="disabled"
@@ -36,9 +37,24 @@ const contentGroupId = 'content';
       [size]="size"
       [variant]="variant"
     >
-      <ng-content></ng-content>
+      {{ content }}
       <lg-icon *ngIf="showIcon || iconButton" [name]="icon"></lg-icon>
     </button>
+    <p>Used on a <strong>link</strong> element</p>
+    <a
+      lg-button
+      href="#"
+      [disabled]="disabled"
+      [fullWidth]="fullWidth"
+      [iconButton]="iconButton"
+      [iconPosition]="iconPosition"
+      [loading]="loading"
+      [size]="size"
+      [variant]="variant"
+    >
+      {{ content }}
+      <lg-icon *ngIf="showIcon || iconButton" [name]="icon"></lg-icon>
+    </a>
   `,
 })
 class ButtonStoryComponent {
@@ -51,6 +67,7 @@ class ButtonStoryComponent {
   @Input() showIcon: boolean;
   @Input() size: ButtonSize;
   @Input() variant: string;
+  @Input() content: string;
   icons = iconsArray;
   constructor(private registry: LgIconRegistry) {
     this.registry.registerIcons(this.icons);
@@ -95,8 +112,8 @@ const createBtnStory = (config: KnobsConfig) => ({
       [showIcon]="showIcon"
       [size]="size"
       [variant]="variant"
+      [content]="content"
       >
-      {{content}}
   </lg-button-story>
   `,
   props: {
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -24,34 +24,34 @@
 ) {
   $box-shadow-inset-width: -0.063rem;
 
-  color: $default-color !important;
+  color: $default-color;
   text-decoration: none;
   border-bottom: 0.125rem solid $default-color;
   padding: 0 0.125rem;
 
   &:hover {
-    color: $hover-color !important;
+    color: $hover-color;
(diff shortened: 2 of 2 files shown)


elenagarrone:
I'm wondering if maybe it should be the button to have the colours with `important` if it's a link?

In the current implementation I would probably try to find a better name for the variable? or maybe add a comment to explain what's the usage of that argument for people wanting to use the mixin?

owensgit:
@elenagarrone 

> I'm wondering if maybe it should be the button to have the colours with `important` if it's a link?

`!important` was originally added just because of  the `lg-variant` directive - I think just to follow what the margin/padding directives do - they use `!important` on their styles to ensure that they always work. At some point in the `next` branch, `!important` got added to the global link styles, which I don't think is correct anyway - and is what caused this issue with the button. So this approach essentially reverts back to the original setup whereby any links inside `lg-variant` only had `!important` applied

I suppose one question is, do we want `!important` on the `lg-variant` directive styles? If so, I think we need to keep this approach, because it only applies `!important` for link colours inside `lg-variant`.

If we don't want to use `!important` for `lg-variant`, then we could just remove it completely - it actually seems to work fine without it.

Either way, I don't think we need the button colours to use `!important` - again because the original use of `!important` was only for links inside of `lg-variant`.

My inclination is to remove `!important` entirely.

Hope that makes sense, happy to jump on a call to discuss 🙂

elenagarrone:
> @elenagarrone
> 
> > I'm wondering if maybe it should be the button to have the colours with `important` if it's a link?
> 
> `!important` was originally added just because of the `lg-variant` directive - I think just to follow what the margin/padding directives do - they use `!important` on their styles to ensure that they always work. At some point in the `next` branch, `!important` got added to the global link styles, which I don't think is correct anyway - and is what caused this issue with the button. So this approach essentially reverts back to the original setup whereby any links inside `lg-variant` had `!important` applied and nowhere else.
> 
> I suppose one question is, do we want `!important` on the `lg-variant` directive styles? If so, I think we need to keep this approach, or similar, because it only applies `!important` for link colours inside `lg-variant`.
> 
> If we don't want to use `!important` for `lg-variant`, then we could just remove it completely - it actually seems to work fine without it.
> 
> Either way, I don't think we need the button colours to use `!important` - again because the original use of `!important` was only for links inside of `lg-variant`.
> 
> My inclination is to remove `!important` entirely.
> 
> Hope that makes sense, happy to jump on a call to discuss 🙂

Yes, thanks for the explanation, I agree 👍 
I thought `!important` was added to the variants because it was breaking something?

owensgit:
> From memory we only started to add !important to deal with apps migrating to Canopy, to make sure that the Canopy styles won over #49 (@andy-polhill)

Yes that's right, we added `!important` to `lgMargin` and `lgPadding` because of that reason.  I then added it to `lgVariant` just by way of following a pattern. I think it was removed with the introduction of the `lg-link` mixin, but then added back in a subsequent fix. As mentioned below, I don't think we actually need it though.

> I thought !important was added to the variants because it was breaking something? (@elenagarrone)

I thought that too - which I why I created that mixin argument - but then I removed `!important` all together, tested it Storybook and in-situ in my app, it seems to work fine. @elenagarrone Do you remember if there were any issues at your end at all?

> However it does feel like we are falling into the trap of a specificity war. (@andy-polhill)

Yes I agree, I think given the above testing without `!important` shows that we can probably remove it in this instance, and then possibly look at removing them from `lgMargin` and `lgPadding` if they Angular now takes care of that for us.

elenagarrone:
> > I thought !important was added to the variants because it was breaking something? (@elenagarrone)
> 
> I thought that too - which I why I created the mixin argument in this PR - but then I removed `!important` all together, tested it Storybook and in-situ in my app, it seems to work fine. @elenagarrone Do you remember if there were any issues at your consuming app at all?

I thought it was because we had to override the generic classes setting colours on the components but now that I think about it I don't think it would matter, right? 

Did you not open this to fix any bug? or did you just noticed that the `!important` was missing?

owensgit:
@elenagarrone @andy-polhill @paul-request  - Thanks for the great discussion on this, as agreed, I've removed `!important` from the link styles entirely, even for `lg-variant`.

andy-polhill (approved):
Good fix @owensgit thanks for chasing this one down.

elenagarrone (approved):

PR #347, opened 2021-05-05
fix(variants): prevent lg-link from overiding lg-btn styles

owensgit:
## Description

The `lg-link` mixin is used to style links inside the variant CSS (e.g. for success, error
variants). However it was causing layout issues as it was clashing with the `.lg-btn` styles. This
is because the links inside varaint CSS were gaining higher specifity. Adding `a:not(.lg-btn)`
prevents this from happening.

**Before**
Screenshot 2021-04-30 at 09 55 11

**After**
Screenshot 2021-05-05 at 10 38 06

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/variant/variant.stories.ts
+++ b/projects/canopy/src/lib/variant/variant.stories.ts
@@ -25,9 +25,11 @@ const variants = ['generic', 'info', 'success', 'warning', 'error'];
           This card has the <strong>{{ variant }}</strong> variant applied. Here is some
           <a href="#">link text</a>.
         </p>
-        <button lg-button variant="outline-primary" lgMarginBottom="none">
-          Outline primary button
-        </button>
+        <button lg-button variant="outline-primary">Outline primary button</button>
+        <br />
+        <a href="#" lg-button variant="outline-primary" lgMarginBottom="none">
+          Outline primary link styled as button
+        </a>
       </lg-card-content>
     </lg-card>
     <lg-card>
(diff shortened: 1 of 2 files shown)


projects/canopy/src/styles/mixins.scss
```diff
@@ -144,7 +144,7 @@ $breakpoints: (
   background-color: var(--#{$variant}-bg-color) !important;
   color: var(--#{$variant}-color) !important;
 
-  a {
+  a:not(.lg-btn) {
```
owensgit, line 147: Note this is inside the variant CSS, so it becomes `.lg-variant--success a:not(.lg-btn)` for example

projects/canopy/src/lib/variant/variant.stories.ts
```diff
@@ -28,6 +28,11 @@ const variants = ['generic', 'info', 'success', 'warning', 'error'];
         <button lg-button variant="outline-primary" lgMarginBottom="none">
```
andy-polhill, line 28:
```suggestion
        <button lg-button variant="outline-primary">
```
I know it's a draft but eventually you could do this instead of the `<br />`

projects/canopy/src/styles/mixins.scss
andy-polhill: I think thats ok, I assume we would never directly apply the `variant` directive to the link anyway.

projects/canopy/src/lib/variant/variant.stories.ts
owensgit: Ah yes, was meaning to do that :) I still one of the `<br />` though as the buttons display inline by default. That's updated now.

andy-polhill (approved):

elenagarrone (approved):

PR #349, opened 2021-05-06
feat(shadow): adds shadow directive

paul-request:
Removes the default shadow from the card component

Fixes #277

Storybook link: https://deploy-preview-349--legal-and-general-canopy.netlify.app

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/shadow/shadow.directive.spec.ts
+++ b/projects/canopy/src/lib/shadow/shadow.directive.spec.ts
@@ -1,0 +1,17 @@
+import { By } from '@angular/platform-browser';
+
+import { MockBuilder, MockRender } from 'ng-mocks';
+
+import { LgShadowDirective } from './shadow.directive';
+
+describe('LgShadowDirective', () => {
+  beforeEach(() => MockBuilder(LgShadowDirective));
+
+  it('automatically sets the class', () => {
+    const fixture = MockRender(`<div lgShadow>Shadow</div>`);
+    const hostElem: HTMLDivElement = fixture.debugElement.query(
+      By.directive(LgShadowDirective),
+    ).nativeElement;
+    expect(hostElem.getAttribute('class')).toContain('lg-shadow');
+  });
+});
--- a/projects/canopy/src/lib/shadow/shadow.directive.ts
+++ b/projects/canopy/src/lib/shadow/shadow.directive.ts
@@ -1,0 +1,8 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+  selector: '[lgShadow]',
+})
+export class LgShadowDirective {
+  @HostBinding('class.lg-shadow') class = true;
+}
--- a/projects/canopy/src/lib/shadow/shadow.notes.ts
+++ b/projects/canopy/src/lib/shadow/shadow.notes.ts
@@ -1,0 +1,28 @@
+export const notes = `
+# Shadow Directive
+
+## Purpose
+This directive adds a box shadow to a component.
+
+## Usage
+
+Import the module in your application:
+
+~~~js
+@NgModule({
+  ...
+  imports: [LgShadowModule],
+})
+~~~
+
+And in your HTML...
+
+~~~html
+<lg-card lgShadow>
+  <lg-card-content>
+    This card has a shadow
+  </lg-card-content>
+</lg-card>
+~~~
+
+`;
--- a/projects/canopy/src/lib/card/card.component.scss
+++ b/projects/canopy/src/lib/card/card.component.scss
@@ -1,7 +1,6 @@
 .lg-card {
   display: block;
   background-color: var(--card-bg-color);
-  box-shadow: var(--drop-shadow);
   border-radius: var(--border-radius-lg);
   margin-bottom: var(--grid-gutter);
   padding: var(--component-padding);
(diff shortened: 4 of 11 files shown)


projects/canopy/src/lib/shadow/shadow.directive.ts
```diff
@@ -0,0 +1,10 @@
+import { Directive, HostBinding } from '@angular/core';
+
+@Directive({
+  selector: '[lgShadow]',
+})
+export class LgShadowDirective {
+  @HostBinding('class.lg-shadow') class = true;
+
+  constructor() {}
```
owensgit, line 9: I think we can remove the constructor

projects/canopy/src/lib/shadow/shadow.notes.ts
```diff
@@ -0,0 +1,28 @@
...
+  imports: [LgShadowModule],
+})
+~~~
+
+And in your HTML...
+
+~~~html
+<lg-card lgShadow>
+  <lg-card-content>
+    This card has a shaow
```
owensgit, line 23:
```suggestion
    This card has a shadow
```

owensgit (changes requested):
Nice one @paul-request 👍  just a couple of minor requests from me.

paul-request:
> Nice one @paul-request 👍 just a couple of minor suggestions from me.

Thanks @owensgit I've made those changes

owensgit (approved):

projects/canopy/src/lib/shadow/shadow.directive.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { By } from '@angular/platform-browser';
+
+import { MockBuilder, MockRender } from 'ng-mocks';
+
+import { LgShadowDirective } from './shadow.directive';
+
+fdescribe('LgShadowDirective', () => {
```
elenagarrone, line 7: `fdescribe` ;)

elenagarrone (changes requested):

projects/canopy/src/lib/shadow/shadow.directive.spec.ts
paul-request: haha, doh

elenagarrone:
@paul-request I think this might need to be flagged as a breaking change? Mostly for projects that are in the process of moving to Canopy. For example in our project we would have to either add the new directive to the `lg-card` or update the old cards in our codebase to remove the box-shadow property so that it matches the new styles.

paul-request:
> @paul-request I think this might need to be flagged as a breaking change? Mostly for projects that are in the process of moving to Canopy. For example in our project we would have to either add the new directive to the `lg-card` or update the old cards in our codebase to remove the box-shadow property so that it matches the new styles.

Yeah I wasn't sure on that. As it won't necessarily break anything, will jus look a little different

elenagarrone:
> > @paul-request I think this might need to be flagged as a breaking change? Mostly for projects that are in the process of moving to Canopy. For example in our project we would have to either add the new directive to the `lg-card` or update the old cards in our codebase to remove the box-shadow property so that it matches the new styles.
> 
> Yeah I wasn't sure on that. As it won't necessarily break anything, will jus look a little different

I know what you mean but I think I would still prefer it to be flagged as some projects might not notice and go live with a look that is not right :/ 
What does @owensgit and @andy-polhill think?

owensgit:
> > > @paul-request I think this might need to be flagged as a breaking change? Mostly for projects that are in the process of moving to Canopy. For example in our project we would have to either add the new directive to the `lg-card` or update the old cards in our codebase to remove the box-shadow property so that it matches the new styles.
> > 
> > 
> > Yeah I wasn't sure on that. As it won't necessarily break anything, will jus look a little different
> 
> I know what you mean but I think I would still prefer it to be flagged as some projects might not notice and go live with a look that is not right :/
> What does @owensgit and @andy-polhill think?

Good question - I guess this is the first time we've got a substantial visual change, which isn't an actual breaking change in the code. Perhaps we need to a call to chat about this and figure out how we want to flag these kinds of things, whether as a semver breaking change, or some other way? 🤔

andy-polhill:
It's a tricky one this one, I'm trying to consider the different scenarios.
1️⃣  You had a drop shadow but shouldn't have, we made it better.
2️⃣  You had a drop shadow and should have, we made it worse.

On the whole a visual change that doesn't affect code is a benefit of the design system, we can easily iterate the design across multiple places. There may be some instances here where we modify the visual styling outside of the intention, but we don't break functionality, so the impact is relatively low. we could...

1️⃣  Document the change a bit more in the commit message to just highlight the risk. This means it can make it in straight away and doesn't have a breaking change barrier, but could effect some styling.
2️⃣  Tag it as a breaking change and put it on `next` which would delay the merge slightly, the breaking change barrier is coming anyway
3️⃣  Tag it as breaking change and merge sooner, which would be safe but our major versions move quite fast.

Personally I would go for 1️⃣  because it helps with the flow of work and the risk is low, but I'm open to a consensus.

elenagarrone:
Ok, happy to go with 1️⃣  👍

elenagarrone (approved):

owensgit:
I'm happy with Option 1️⃣  too. For me a semver breaking change feels a bit heavy as I'd think that I would need to make code changes when upgrading, but that's necessarily the case.

I wonder if we should have a standard commit message template for this kind of thing? E.g.

```
VISUAL CHANGE: 
Card components no longer have a drop shadow by default.
Use `lgShadow` to apply drop shadow where required.
```

paul-request:
Ok, I will amend the commit message accordingly 👍

andy-polhill (approved):

owensgit (approved):

PR #350, opened 2021-05-07
fix(variants): add hover and active state style to variant buttons

AlexandruHoisan:
# Description

Add styles for hover and active state, to variant buttons.

Note FYI: After a discussion with Michelle from the design team, we decided on this temporary solution until we design and build the component that combines the current `Alert component`, the`Detail component`, the `Form Validation component` and the `Variant Directive` that is being modified here. Inside variants we may only use the `outline-primary` button.

Design link: [design link here](https://legalandgeneral.invisionapp.com/console/share/SB10JVR247UQ/613975239/play)

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/heading/heading.component.spec.ts
+++ b/projects/canopy/src/lib/heading/heading.component.spec.ts
@@ -3,7 +3,7 @@ import { By } from '@angular/platform-browser';
 
 import { LgHeadingComponent } from './heading.component';
 
-fdescribe('LgHeadingComponent', () => {
+describe('LgHeadingComponent', () => {
   let component: LgHeadingComponent;
   let fixture: ComponentFixture<LgHeadingComponent>;
 
--- a/projects/canopy/src/lib/variant/variant.stories.ts
+++ b/projects/canopy/src/lib/variant/variant.stories.ts
@@ -64,9 +64,9 @@ export default {
         ],
       }),
     ],
-  },
-  notes: {
-    markdown: notes,
+    notes: {
+      markdown: notes,
+    },
   },
 };
 
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -160,5 +160,58 @@ $breakpoints: (
     background-color: transparent !important;
     border-color: var(--#{$variant}-color) !important;
     color: var(--#{$variant}-color) !important;
+
+    @if ($variant == 'generic') {
+      &:hover {
+        background-color: var(--color-taupe-grey) !important;
+        border-color: var(--color-taupe-grey) !important;
+        color: var(--color-white) !important;
+      }
+
+      &:active {
+        background-color: var(--color-charcoal) !important;
+        color: var(--color-white) !important;
+      }
+    } @else if ($variant == 'info') {
+      &:hover {
+        background-color: var(--color-midnight-blue) !important;
+        color: var(--color-white) !important;
+      }
+
+      &:active {
+        background-color: var(--color-super-blue-darkest) !important;
+        color: var(--color-white) !important;
+      }
+    } @else if ($variant == 'success') {
+      &:hover {
+        background-color: var(--color-mexican-green) !important;
+        color: var(--color-white) !important;
+      }
+
+      &:active {
+        background-color: var(--color-leafy-green-dark) !important;
+        color: var(--color-white) !important;
+      }
+    } @else if ($variant == 'warning') {
+      &:hover {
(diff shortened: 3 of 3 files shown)


projects/canopy/src/lib/heading/heading.component.spec.ts
```diff
@@ -3,7 +3,7 @@ import { By } from '@angular/platform-browser';
 
 import { LgHeadingComponent } from './heading.component';
 
-fdescribe('LgHeadingComponent', () => {
+describe('LgHeadingComponent', () => {
```
elenagarrone, line 6: Well spotted 👍

projects/canopy/src/styles/mixins.scss
```diff
@@ -160,5 +160,23 @@ $breakpoints: (
...
     color: var(--#{$variant}-color) !important;
+
+    &:active,
+    &:active:hover {
+      background-color: var(--#{$variant}-link-active-hover-color) !important;
+      color: var(--color-white) !important;
+    }
+
+    @if ($variant == 'generic') {
+      &:hover {
```
elenagarrone, line 171: Can I just say that I don't like the hover state for the generic button? 🤣  why doesn't it follow the same pattern as the other ones?

AlexandruHoisan: totally understandable, I don't like it either :)). The reason for this is that we already have a `--generic-link-hover-color` defined and it is not the color we need for the generic hover state, from the designs. I could not change the existing `--generic-link-hover-color` as that one is used for the link in the generic variant. So this was my solution for using `--color-taupe-grey` as background for the button in the generic variant, without breaking the links. Hope that makes sense :).

owensgit: I wonder if it's worth just running this by design again, as there may have been a mistake in the designs? They may change the designs if it's more consistent with the other ones.

projects/canopy/src/styles/variables/_variants.scss
```diff
@@ -8,6 +8,7 @@
   --generic-link-active-bg-color: var(--color-white);
   --generic-link-focus-color: var(--color-white);
   --generic-link-focus-bg-color: var(--color-black);
+  --generic-link-active-hover-color: var(--color-charcoal);
```
owensgit, line 11: Same as my comment on the sibling PR:

https://github.com/Legal-and-General/canopy/pull/351#discussion_r628215597

projects/canopy/src/styles/mixins.scss
AlexandruHoisan: Will run this by design again and see what we should  do. :)

projects/canopy/src/styles/variables/_variants.scss
AlexandruHoisan: I do have a small question on this one. If I get rid of the variables here, then I have to add the `@if @else` in the mixins.scss, and there will be some boilerplate code, as in master. Is that ok or did you have something else in mind?  :)

projects/canopy/src/styles/mixins.scss
elenagarrone: Oh I think the issue is that the border color is supposed to be the same `--color-taupe-grey` on hover as the background.

projects/canopy/src/lib/variant/variant.stories.ts
```diff
@@ -64,9 +64,9 @@ export default {
         ],
       }),
     ],
-  },
-  notes: {
-    markdown: notes,
+    notes: {
```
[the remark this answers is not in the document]
AlexandruHoisan: The notes were not showing for Variants because they needed to be in the parameters object. :)

projects/canopy/src/styles/variables/_variants.scss
AlexandruHoisan: can jump on a quick call to talk this over id needed :)

projects/canopy/src/styles/mixins.scss
AlexandruHoisan: @owensgit Had a chat with Michelle this morning and she thinks this is ok for now. This will be looked at again when the component that combines "Variants", "Alerts" and "Details" is going to be designed and built.
@elenagarrone yep, will do that now.

Michelle said that if we need to discuss this further she will be happy to jump on a call with us :).

projects/canopy/src/styles/variables/_variants.scss
AlexandruHoisan: Done like in the other PR and the button states now uses variables of colours and not the link defined ones. :)

PR #352, opened 2021-05-07
refactor(side-nav): adds side nav cards

paul-request:
# Description

This PR adds the correct layout and styling to the Side Nav Bar component.

Storybook link: https://deploy-preview-352--legal-and-general-canopy.netlify.app

# Checklist:

- [x] The commit messages follow the [convention for this project](./blob/master/docs/CONTRIBUTING.md#conventional-commits)
- [x] I have provided an adequate amount of test coverage
- [ ] I have added the functionality to the [test app](./blob/master/docs/CONTRIBUTING.md#build-test-application)
- [x] I have provided a story in storybook to document the changes
- [x] I have provided documentation in the notes section of the story
- [x] I have added any new public feature modules to public-api.ts

PR #358, opened 2021-05-10
Revert "refactor(side-nav): adds side nav cards"

paul-request:
Reverts Legal-and-General/canopy#352

Reverting as it used the `refactor` commit which doesn't trigger a release 🤦 . Will re-raise with correct commit


owensgit (commented):
Just wondering what the rules are here as the commit doesn't seem to follow the commitlint rules:

```
Revert "refactor(side-nav): adds side nav cards" 
```

I think there's a revert commit. I wonder if we should change this commit message to be...

```
revert(side-nav)  adds side nav cards

Reverts XYZetc
```

Part of me is also wondering if we're able to edit the commit directly on the master branch, and then push it up? Maybe only @andy-polhill has the access to do that?

paul-request:
> Just wondering what the rules are here as the commit doesn't seem to follow the commitlint rules:
> 
> ```
> Revert "refactor(side-nav): adds side nav cards" 
> ```
> 
> I think there's a revert commit. I wonder if we should change this commit message to be...
> 
> ```
> revert(side-nav)  adds side nav cards
> 
> Reverts XYZetc
> ```
> 
> Part of me is also wondering if we're able to edit the commit directly on the master branch, and then push it up? Maybe only @andy-polhill has the access to do that?

Yeah, this PR was just created by GitHub using the revert commit button, so I guess it won't follow our commit lint rules

elenagarrone (approved):

paul-request:
I may not need this if this PR gets merged anyhow:

PR #362, opened 2021-05-11
feat(padding-directive): add responsive padding

### LINKED ISSUE #110: lgPadding directive doesn't support breakpoints

owensgit:
# Description

~Draft~ Implementation for responsive margin/padding util directives. ~Looking for feedback on both approach and API~.

Update: now moved out of draft.

Fixes #110

## API

The idea behind this approach is that it’s backwards compatible. You can pass in a standard spacing variant to `lgPadding` - and it will work the same as before. However you can pass in an object instead, which describes the responsive behaviour.

`<div [lgPadding]="{ sm: 'sm', md: 'xxl' }"></div>`

Example object:
```
{
  sm: "md",
  md: "xxl"
}
```

Each key is a mobile-first breakpoint, corresponding to our standard list of breakpoints in the `$breakpoints` SASS map: `sm`, `md`, `lg`, `xl`, `xxl`.

The corresponding value is one of the standard spacing variants. 

So the above example describes the following:
- At the `sm` breakpoint, the padding variant will be `md`
- At the `md` breakpoint, the padding variant will be `xxxl`

## Requirements

Take a look at the padding and margin directives in Storybook, it has a new card to work with the new responsive object, which has corresponding fields in the knobs tab.

**Storybook link: https://deploy-preview-362--legal-and-general-canopy.netlify.app/?path=/story/directives--padding**

Screenshot: 

Screenshot 2021-05-11 at 14 02 10

# Checklist:


elenagarrone:
@owensgit great work on this 👍  I like the implementation and I agree that the advantage of having the responsive directives outweighs the size increase of the files.

projects/canopy/src/lib/spacing/spacing.interface.ts
```diff
@@ -10,3 +10,11 @@ export type SpacingVariant =
   | 'xxl'
   | 'xxxl'
   | 'xxxxl';
+
+export interface ResponsiveSpacing {
+  sm?: SpacingVariant;
+  md?: SpacingVariant;
+  lg?: SpacingVariant;
+  xlg?: SpacingVariant;
```
dannyoz, line 18: Shouldn't these keys be `xl` / `xxl` to follow the naming conventions?

andy-polhill:
The size it adds to the css did surprise me a bit, like you say it's not too bad, but I do wonder if there is a way to keep that list small. I wondered what an inline css approach might look like? So instead of adding classes we add inline styles?

Normally inline styles are a bit smelly of course, but I wondered if it could be preferable for the size improvements 🤔. Or alternatively we could even build the [css styles dynamically](https://stackoverflow.com/questions/1720320/how-to-dynamically-create-css-class-in-javascript-and-apply) as we go.

projects/canopy/src/lib/spacing/spacing.interface.ts
andy-polhill: good call, it could probably use these?
https://github.com/Legal-and-General/canopy/blob/master/projects/canopy/src/lib/shared/breakpoints.interface.ts

projects/canopy/src/lib/spacing/padding/padding.directive.ts
```diff
@@ -14,69 +14,78 @@ import { SpacingVariant } from '../spacing.interface';
...
-  paddingRightClass: string;
+  paddingRightClasses: Array<string> = [];
   @Input()
-  set lgPaddingRight(padding: SpacingVariant) {
-    this.paddingRightClass = this.togglePaddingClass(
-      `lg-padding__right--${padding}`,
-      this.paddingRightClass,
-    );
+  set lgPaddingRight(padding: SpacingVariant | ResponsiveSpacing) {
+    const newClasses = this.createNewClasses(padding, 'lg-padding__right');
```
andy-polhill, line 27: Minor stylistic suggestion, but might not be to everyones tastes :-)
```suggestion
    this.paddingRightClasses = this.toggleClasses(
      this.createNewClasses(padding, 'lg-padding__right'),
      this.paddingRightClasses
    );
```

projects/canopy/src/lib/spacing/padding/padding.directive.ts
```diff
@@ -14,69 +14,78 @@ import { SpacingVariant } from '../spacing.interface';
...
-    );
+  set lgPadding(padding: SpacingVariant | ResponsiveSpacing) {
+    const newClasses = this.createNewClasses(padding, 'lg-padding');
+    this.paddingClasses = this.toggleClasses(newClasses, this.paddingClasses);
   }
 
-  togglePaddingClass(newClass: string, oldClass: string): string {
-    if (oldClass) {
-      this.renderer.removeClass(this.hostElement.nativeElement, oldClass);
+  createNewClasses(padding: SpacingVariant | ResponsiveSpacing, classPrefix: string) {
```
andy-polhill, line 64: There's something I don't quite like about this. I think it's the fact that `Spacing` is the first word on one, and the second on the other. I also wonder if `ResponsiveSpacing` should be more of a noun? e.g. `SpacingRules`. (not too sure)

projects/canopy/src/lib/spacing/padding/padding.directive.ts
```diff
@@ -14,69 +14,78 @@ import { SpacingVariant } from '../spacing.interface';
...
+    const newClasses = [];
+    if (!padding) {
+      return [];
     }
-    this.renderer.addClass(this.hostElement.nativeElement, newClass);
-    return newClass;
+    if (typeof padding === 'object') {
+      Object.keys(padding).forEach((key) => {
```
andy-polhill, line 70: Just a minor one, but would this be nicer?
```ts
if (typeof padding === 'object') {
  return Object.keys(padding).map(key => `${classPrefix}--${key}--${padding[key]}`);
} else {
  return [`${classPrefix}--${padding}`]
}
```

projects/canopy/src/lib/spacing/padding/padding.directive.ts
owensgit: How about `SpacingResponsive`? I quite like it - short and to the point. I've added that for now in my most recent commit, but can change it again.

owensgit:
As per [Andy's comment](https://github.com/Legal-and-General/canopy/pull/362#issuecomment-853699750), I had a go at **adding the CSS styles dynamically as we go**, rather than double the canopy.css filesize.

It's a new commit on this branch, so you can see the difference. 
 
**feat(padding-directive): implement dynamic styles**. 

Happy to drop this though if we feel the approach is too complex.

**How it works**
_Note this is very much a draft implementation_
A new service (`DynamicStylesService`) is responsible for adding new styles to a `<script>` tag in the `<head>` of the document. In the padding directive, we compile the styles needed and then add them in one by one, this can be abstracted for use in margin and others too. `DynamicStylesService` will only add styles that haven't previously been added. Currently it only works with all-round padding, and not the specific ones (e.g. `padding-left`) as I just wanted to get something up for now.

Benefit of this is that there's no extra weight on the canopy.css, and we only add the styles when they are required. There is some extra weight now on the padding directive, but only if choose to import it. The same method applies to the non-responsive styles - which means those classes can be removed too, further reducing the canopy.css file.

Let me know what you think!

projects/canopy/src/lib/spacing/padding/padding.directive.ts
```diff
@@ -14,70 +16,94 @@ import { SpacingVariant } from '../spacing.interface';
...
-    );
+  set lgPadding(padding: SpacingVariant | SpacingResponsive) {
+    const newClasses = this.createNewClasses(padding, 'padding');
+    this.paddingClasses = this.toggleClasses(newClasses, this.paddingClasses);
   }
 
-  togglePaddingClass(newClass: string, oldClass: string): string {
-    if (oldClass) {
-      this.renderer.removeClass(this.hostElement.nativeElement, oldClass);
+  createNewClasses(spacing: SpacingVariant | SpacingResponsive, spacingProp: string) {
```
owensgit, line 66: This is where the dynamic styles are added.

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { BreakpointValues } from '../shared/breakpoints.interface';
+
+@Injectable({
+  providedIn: 'root',
+})
+/**
+ * Dynamically add styles to a <style> tag in <head>
+ */
+export class DyanmicStyleService {
```
andy-polhill, line 11: I know it's a draft but the typo :-) `DyanmicStyleService => DynamicStyleService`

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    if (this.selectors.includes(selector)) {
+      return;
+    }
+    this.selectors.push(selector);
+    const styleString = `.${selector}{${style}!important}`;
+    this.styleTag.innerHTML += styleString;
+  }
+
+  addStyleToMediaQuery(
```
andy-polhill, line 45: I wondered about the two methods, not really focusing in implementation, but could one method be more intuitive? [Terminology wise](https://css-tricks.com/css-ruleset-terminology/) I wonder if it's a `rule` or `ruleset`

```ts
addRule(selector: string, declaration: string, breakpoint?: BreakpointValues)
```

andy-polhill: If you do manage to cut it down to one public method, you might be able to trim this repetition down to only be in that method...
```ts
if (!this.styleTag) {
  this.createStyleTag();
}
```

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    selector: string,
+    style: string,
+    breakpoint: BreakpointValues,
+  ): void {
+    if (!this.styleTag) {
+      this.createStyleTag();
+    }
+    if (!this.mediaQueries[breakpoint]) {
```
andy-polhill, line 53: This kind of checking is interesting, should we do it here, or in the method itself, which can determine whether it should do something or return early. Not sure if there is a name for that approach, but doing it in the method almost feels safer? Interested in opinions on that one

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    this.mediaQueries[breakpoint].push(selector);
+
+    const styleString = `.${selector}{${style}!important}`;
+
+    const innerHTML = this.styleTag.innerHTML;
+    const mqIndex = innerHTML.indexOf(`@media(min-width:${breakpoint}){`) + 1;
+    const htmlArr = [...innerHTML];
+    htmlArr.splice(mqIndex, 0, styleString);
+    this.styleTag.innerHTML = htmlArr.join('');
```
andy-polhill, line 68: I wonder whether changing the innerHTML on multiple occasions could trigger noticeable repaints 🤔
If so, could we just change it once?
Maybe even via an `addRules` method for multiple ones, to cut down the repaint?

projects/canopy/src/lib/spacing/padding/padding.directive.ts
```diff
@@ -16,62 +18,73 @@ import { ResponsiveSpacing, SpacingVariant } from '../spacing.interface';
...
+      Object.keys(spacing).forEach((key) => {
+        const className = `lg-${spacingProp}--${key}--${spacing[key]}`;
+        const spacingKey = spacing[key];
+        const styleRule = `${spacingProp}: ${SpacingValues[spacingKey]}`;
+        this.dynamicStyleService.addStyleToMediaQuery(
+          className,
+          styleRule,
+          BreakpointValues[key],
+        );
+        newClasses.push(className);
```
andy-polhill, line 81: Tempted to ditch the assignments for brevity, but that could be a style opinion.
```
this.dynamicStyleService.addStyleToMediaQuery(
  `lg-${spacingProp}--${key}--${spacing[key]}`,
   `${spacingProp}: ${SpacingValues[spacing[key]]},
   BreakpointValues[key],
);
```

andy-polhill (commented):
Looking good @owensgit I put in a few more considerations, which may or may not be useful.

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: I think I know what you mean, what method are you referring to?

projects/canopy/src/lib/spacing/padding/padding.directive.ts
owensgit: Changed to `SpacingResponsive`

projects/canopy/src/lib/spacing/padding/padding.directive.ts
owensgit: This has changed quite a bit and is now in the **spacing.service.ts** file

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: I've updated it now so that it only updates the innerHTML once after each set of rules or media queries are created, rather than each time. This is noticeable when the responsive spacing object is used, as it may contain two or three rules, with two or three media queries, which now are only added once they are all created. That brings it from 3 to 6 innerHTML changes to only 1.

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Both methods are now called `addRules` and `addRulesToMediaQuery`. I didn't manage to combine the methods, as each one does a fair amount of different logic, that I didn't want to nest into a lot of `if` statements. To me it feels easier to read this way, but I'd be open to discuss.

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Changed

projects/canopy/src/lib/spacing/padding/padding.directive.ts
owensgit: Done ✅

projects/canopy/src/lib/shared/breakpoints.interface.ts
```diff
@@ -1 +1,9 @@
 export type Breakpoints = 'sm' | 'md' | 'lg' | 'xl' | 'xxl';
+
+export enum BreakpointValues {
```
andy-polhill, line 3: It might be one for the future but I learnt a bit from this article..
https://blog.logrocket.com/why-you-should-use-css-env-9ee719ce0f24/

We may be able to reduce the replication of these variables across css and js.

andy-polhill: I just noticed the spacing variables replication too. I wouldn't be too opposed to these being replicated, but it might be nice to log an issue and reference it in the code for both the css and js, just so people know they need to update both.

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,97 @@
+import { Injectable } from '@angular/core';
+
+import { BreakpointValues } from '../shared/breakpoints.interface';
+
+interface Rule {
+  selector: string;
+  declaration: string;
+}
+interface RuleRepsonive {
```
andy-polhill, line 9:
```suggestion
interface RuleResponsive {
```

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,97 @@
...
+  private createMediaQuery(breakpoint: BreakpointValues): string {
+    if (this.mediaQueries[breakpoint]) {
+      return;
+    }
+    this.mediaQueries[breakpoint] = [];
+    return `@media(min-width:${breakpoint}){}`;
+  }
+
+  addRules(rules: Array<Rule>): void {
+    if (!this.styleTag) {
```
andy-polhill, line 41: quite minor, but the assumption being that if this service is loaded then it will be used, does that mean we could do this in the `constructor`?

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    rules.map((r) => {
+      if (this.selectors.includes(r.selector)) {
+        return;
+      }
+
+      this.selectors.push(r.selector);
+
+      const ruleString = `.${r.selector}{${r.declaration}!important}`;
+      styleTagCache += ruleString;
```
andy-polhill, line 56:
```suggestion
      styleTagCache += `.${r.selector}{${r.declaration}!important}`;
```
it's just styling preference, but you could leave the assignment.

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  addRules(rules: Array<Rule>): void {
+    if (!this.styleTag) {
+      this.addStyleTag();
+    }
+
+    // cache current styles to avoid multiple edits to the innerHTML
+    let styleTagCache = this.styleTag.innerHTML.slice();
+
+    rules.map((r) => {
```
andy-polhill, line 48:
```suggestion
    rules
      .filter((r) => !this.selectors.includes(r.selector))
      .map((r) => {
```
More of a style one again, but could you use a filter instead of the early return on line 49-50?

projects/canopy/src/lib/utils/dynamic-style.service.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (this.mediaQueries[r.breakpoint].includes(r.selector)) {
+        return;
+      }
+
+      // create record to avoid adding same rule twice
+      this.mediaQueries[r.breakpoint].push(r.selector);
+
+      const ruleString = `.${r.selector}{${r.declaration}!important}`;
+
+      // find the existing media query frm the style tag
```
andy-polhill, line 84:
```suggestion
      // find the existing media query from the style tag
```

projects/canopy/src/styles/spacing.scss
```diff
@@ -46,16 +48,18 @@ $spacing-list: (
   }
 }
 
-@each $val in $spacing-list {
-  .lg-margin--#{$val} {
-    margin: space($val) !important;
-  }
+// Default margin / padding
+// @each $val in $spacing-list {
```
andy-polhill, line 52: I think these might need to stay 🤔. Would it create much of an issue?
I think we said that these classes would come under semantic versioning.
https://github.com/Legal-and-General/canopy/blob/master/docs/BREAKING_CHANGES.md

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: 🤦‍♂️  Fixed ✅

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Ah good point, I agree - also cuts a little duplication. Moved it ✅

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Nice, yeah it's a little cleaner this way, done ✅

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Ah of course, thanks - added ✅

projects/canopy/src/lib/utils/dynamic-style.service.ts
owensgit: Fixed ✅

projects/canopy/src/styles/spacing.scss
owensgit: Ah good point, I wasn't sure about this, and was partly commenting them out here to check that the dynamic style service was definitely working. 

I've opened a discussion to cover this: https://github.com/Legal-and-General/canopy/discussions/509

projects/canopy/src/lib/shared/breakpoints.interface.ts
owensgit: Thanks Andy, I'd like propose that we add an issue and do this in another PR, perhaps on one of a dedicated Canopy day, IMO getting this feature in sooner outweighs the benefits of having this duplication and waiting to do the css env implementation. Happy to discuss though

elenagarrone (dismissed):

andy-polhill (approved):

andy-polhill (approved):

elenagarrone (approved):

PR #372, opened 2021-05-18
feat(scss): add font type parameter to lg-font-size mixin

### LINKED ISSUE #224: Add ability to set expressive font via lg-font-size mixin
Currently the only way to apply the expressive (Lyon) font is to use the CSS modifier: `lg-font--expressive`. 

We should also add the ability to apply the expressive font when using the `@lg-font-size` mixin, with the default being the productive (Robot) font.

Current mixin:

```
@mixin lg-font-size($size, $modifier: null) {
  font-size: var(--text-fs-#{$size}-size);
  line-height: var(--text-fs-#{$size}-line-height);

  @if $modifier {
    font-weight: var(--text-fs-#{$size}-weight--#{$modifier});
  } @else {
    font-weight: var(--text-fs-#{$size}-weight);
  }
}
```

Possible approach could be to include a third parameter: `@mixin lg-font-size($size, $modifier: null, $fontType: 'productive')`.

AlenNS:
# Description

Add ability to apply the expressive font when using the _@lg-font-size_ mixin by enhancing the mixin with the third parameter _type_ with the default value null. When no parameter provided for the type of font family there will be no overriding of the font family set from another class. If we provide _productive_ as the value for the type parameter it will override the font family with the productive (Roboto) font. If the _expressive_ value has been set as a type parameter, expressive (Lyon) font will be set.

Fixes #224 

## Requirements

No requirements are needed. If we don't provide type parameter to the lg-font-size mixin there will be no changes to the font-family.

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -22,7 +22,7 @@
   outline: var(--link-focus-outline) solid $bg-color;
 }
 
-@mixin lg-font-size($size, $modifier: null) {
+@mixin lg-font-size($size, $modifier: null, $type: null) {
   font-size: var(--text-fs-#{$size}-size);
   line-height: var(--text-fs-#{$size}-line-height);
 
@@ -30,6 +30,10 @@
     font-weight: var(--text-fs-#{$size}-weight--#{$modifier});
   } @else {
     font-weight: var(--text-fs-#{$size}-weight);
+  }
+
+  @if $type == 'expressive' {
+    font-family: var(--font-family-#{$type});
   }
 }
 


projects/canopy/src/styles/mixins.scss
```diff
@@ -31,6 +31,12 @@
   } @else {
     font-weight: var(--text-fs-#{$size}-weight);
   }
+
+  @if $type == 'expressive' {
+    font-family: var(--font-family-#{$type});
+  } @else if $type == 'productive' {
```
elenagarrone, line 37: Just wondering if `productive` should be the default value?
Maybe actually it could work only with this bit:
```
@if $type == 'expressive' {    font-family: var(--font-family-#{$type});  }
```
since the font-family by default it's set to Roboto? Would that work?

AlenNS: Well, that was my first thought, but I was thinking about the possibility where some other font would be used in the future. This way we would have an option to set Roboto with the productive parameter. Anyway, I would rather go toward fully generic solution where we would take the type parameter and provide it to the `var(--font-family-#{$type}` and set proper font family.

If you think it's better, I will remove productive parameter part. :)

owensgit: I agree with @elenagarrone, I don't think we need `if $type == 'productive'`, because it's the default font anyway which isset by the CSS var `--font-family`. So I think the logic could look like this:

```
@if $type == 'expressive' {    
  font-family: var(--font-family-#{$type});  
} @else {
  font-family: var(--font-family);
}
```

Future we could add more conditions if we add more fonts, and/or if the default font changed too.

owensgit (commented):
Great first contribution @AlenNS! Just one comment from me

owensgit:
One more thing, I think the commit message would be better if the scope was `(scss)`, I think it describes the change a little more precisely and matches with similar past commits:

```
* 811e404 - (tag: v4.0.2) fix(scss): removes unused css variables (4 months ago) <Andrew Polhill>
* a0cc55d - fix(scss): removes unused `text-sm` css variable (4 months ago) <Andrew Polhill>
* a932713 - feat(scss): adds global scss to bundled package (#25) (9 months ago) <Andrew Polhill>
```

projects/canopy/src/styles/mixins.scss
AlenNS: In this case we don't need _@else_ part since it will set the **Roboto** font which is already the default font. It will set **Lyon** font only if we provide _$type_ parameter as expressive and will not set font-family property for this mixin if we don't provide this value for the parameter (null or other value will skip this case).

owensgit:
Also just noticed - I think the commit message may be missing the issue number in the commit message:

```
feat(scss): add ability to set expressive font in font size mixin

fix #224
```

andy-polhill:
> Also just noticed - I think the commit message may be missing the issue number in the commit message:
> 
> ```
> feat(scss): add ability to set expressive font in font size mixin
> 
> fix #224
> ```

@AlenNS not sure if you picked up on this in the docs, but you can use `npm run commit` to build up the commit message. Excellent work,  it's great to have a new contributor on Canopy 🎉

andy-polhill (approved):

AlenNS:
> > Also just noticed - I think the commit message may be missing the issue number in the commit message:
> > ```
> > feat(scss): add ability to set expressive font in font size mixin
> > 
> > fix #224
> > ```
> 
> @AlenNS not sure if you picked up on this in the docs, but you can use `npm run commit` to build up the commit message. Excellent work, it's great to have a new contributor on Canopy 🎉

I will use it next time. 💯 
Thanks a lot. 👍 :)

paul-request (approved):

Issue #371, opened 2021-05-17
Button misaligned when using iconPosition="left"


NemirStephan:
When using iconPosition="left" on a button, the button alignment with other buttons is slightly off

iconPosition="right"
<img width="799" alt="Screenshot 2021-05-17 at 14 59 35" src="https://user-images.githubusercontent.com/22469863/118501790-fdcf7480-b720-11eb-80a4-dcfdffdf06ed.png">

iconPosition="left"
<img width="790" alt="Screenshot 2021-05-17 at 14 59 55" src="https://user-images.githubusercontent.com/22469863/118501807-0162fb80-b721-11eb-9403-21ece4e33b23.png">

PR #373, opened 2021-05-18
fix(scss): fix the button styling when icon position is left

AlenNS:
# Description

Remove _order_ property from the _.lg-icon_ class and set _flex-direction_ property to _row-reverse_ to display icon before the text.

Fixes # (371)

## Requirements

Set the button's iconPosition property to left next to another button/element.

Screenshot: 
2021-05-18 16_20_46-Test Page - Standard ⋅ Storybook

# Checklist:


elenagarrone (approved):
Nice one! The only thing I would suggest is changing the commit scope to be `button` instead of `scss` so that it's a bit more specific.

AlenNS:
> Nice one! The only thing I would suggest is changing the commit scope to be `button` instead of `scss` so that it's a bit more specific.

Thanks @elenagarrone I've changed the score. :)

projects/canopy/src/lib/button/button.component.scss
```diff
@@ -61,8 +61,9 @@
   }
 
   &--icon-left {
+    flex-direction: row-reverse;
```
owensgit, line 64: I suppose the only downside to this approach is that if we were to add a third item inside the button, it might have an unintended affect on the order. E.g. 

```
<button lg-button>
  Submit
  <lg-something>with 10% discount</lg-something>
  <lg-icon name=""></lg-icon>
</button>
```

Or is that a bit of [YAGNI](https://martinfowler.com/bliki/Yagni.html) creeping in? 🙂 

I'm not too worried about this, so would be happy to proceed with this if everyone else is.

owensgit:
I think the commit may also be missing the issue number in the body E.g.

```
fix(button): fix the button styling when icon position is left

fix #371
```

projects/canopy/src/lib/button/button.component.scss
andy-polhill: I think for now it's ok, but you raise a good point. If we ever had a button with an icon at the start and one at the end it would swap them round. However, the current component is only configured to ever have one icon. One to consider if we have a future pattern with multiple icons, but for this I think it's fine.

elenagarrone: I might be wrong but I think @alexcanning told me buttons will only ever have a single icon.

owensgit:
@AlenNS 

> I think the commit may also be missing the issue number in the body E.g.
> 
> ```
> fix(button): fix the button styling when icon position is left
> 
> fix #371
> ```

I think it's just this one outstanding 🙂 

The scope of your commit says `(scss)` but I think it should be `(button)` as it's a fix to a bug in the button component.

And we just need that issue number inside the commit body 👍

owensgit (changes requested):
Just one thing left from me regarding the commit message

AlenNS:
> Just one thing left from me regarding the commit message

@owensgit Updated commit message.

andy-polhill (approved):

owensgit (approved):

PR #375, opened 2021-05-21
build(dependabot): ignore all angular updates

owensgit:
# Description

As suggested in [this Dependabot PR](https://github.com/Legal-and-General/canopy/pull/366), this change should disable all Dependabot updates for Angular packages, as we do these manually at the same time.

Shout if you don't think this is the right thing to do.

# Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,6 +24,10 @@ updates:
       include: 'scope'
     versioning-strategy: increase
     ignore:
+      - dependency-name: "@angular/*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "@angular-devkit/*"
+        update-types: ["version-update:semver-major"]
       - dependency-name: "@storybook/addon-a11y"
         versions: ["6.x"]
       - dependency-name: "@storybook/addon-actions"


.github/dependabot.yml
```diff
@@ -24,6 +24,7 @@ updates:
       include: 'scope'
     versioning-strategy: increase
     ignore:
+      - dependency-name: "@angular/*"
```
andy-polhill, line 27: Can we just ignore major versions? Just in case there are any useful minor or patch fixes that go out.

.github/dependabot.yml
```diff
@@ -24,6 +24,7 @@ updates:
...
       - dependency-name: "@storybook/addon-a11y"
```
elenagarrone, line 28: Are there any other packages that are linked to the angular upgrade? e.g. I remember `ng-mocks` was following the versioning of angular too: https://www.npmjs.com/package/ng-mocks

owensgit: I'd be weary of ignoring other packages, just incase we forgot about them when upgrading Angular. What do you think?

.github/dependabot.yml
owensgit: Good shout! I've updated to only ignore major versions

.github/dependabot.yml
elenagarrone: We are doing the same in my project and we never forget to update whenever I update angular 🤔 
One thing I would definitely add is @angular-devkit

owensgit: Ah okay, shall we see how we go and only add `ng-mocks` if we find the Dependabot updates annoying?

I've added `@angular-devkit` 🙂

elenagarrone (approved):

paul-request (approved):

PR #386, opened 2021-05-27
feat: carousel component added

dannyoz:
# Description

This PR adds a new carousel component which will be used on OSS.  This MVP version currently only supports click / keyboard navigation.  Swipe gesture and autoplay support to be added in future sprints.

View on [Netlify](https://deploy-preview-386--legal-and-general-canopy.netlify.app/?path=/story/components-carousel--standard)

# Checklist:

- [x] I have added the functionality to the [test app](https://deploy-preview-386--legal-and-general-canopy.netlify.app/?path=/story/components-carousel--standard)

### DIFF
--- a/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.html
+++ b/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.html
@@ -1,0 +1,1 @@
+<ng-content></ng-content>
--- a/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
+++ b/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
@@ -1,0 +1,26 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ElementRef,
+  HostBinding,
+  ViewEncapsulation,
+} from '@angular/core';
+
+@Component({
+  selector: 'lg-carousel-item',
+  templateUrl: './carousel-item.component.html',
+  styleUrls: ['./carousel-item.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class LgCarouselItemComponent {
+  @HostBinding('class.lg-carousel-item') class = true;
+  @HostBinding('attr.role') role = 'tab-panel';
+  @HostBinding('attr.aria-selected') public selected = false;
+
+  constructor(public element: ElementRef) {}
+
+  get itemContent() {
+    return this.element.nativeElement.innerHTML;
+  }
+}
--- a/projects/canopy/src/lib/carousel/carousel.component.html
+++ b/projects/canopy/src/lib/carousel/carousel.component.html
@@ -1,0 +1,59 @@
+<section class="lg-carousel" [attr.aria-labelledby]="description">
+  <lg-heading [level]="headingLevel" class="lg-carousel__heading lg-visually-hidden">{{
+    description
+  }}</lg-heading>
+  <div
+    class="lg-carousel__active-content lg-visually-hidden"
+    aria-atomic="true"
+    aria-live="polite"
+    [innerHTML]="selectedItemContent"
+  ></div>
+  <div
+    class="lg-carousel__wrapper"
+    [style.width.%]="100 * carouselItemCount"
+    [style.left.%]="-100 * selectedItemIndex"
+    [style.transition]="'left ease ' + slideDuration / 1000 + 's'"
+  >
+    <ng-content></ng-content>
+  </div>
+
+  <div class="lg-carousel__navigation" role="tablist">
+    <button
+      role="tab"
+      class="lg-carousel__button"
+      (click)="previousCarouselItem()"
+      [disabled]="!loopMode && selectedItemIndex === 0"
+    >
+      <span class="lg-visually-hidden">Previous {{ description }} item</span>
+      <span class="lg-carousel__button__disk lg-carousel--centered">
+        <lg-icon name="chevron-left"></lg-icon>
+      </span>
+    </button>
+    <button
+      role="tab"
+      class="lg-carousel__button lg-carousel--centered"
+      (click)="selectCarouselItem(i)"
+      *ngFor="let item of carouselItems; let i = index"
+      [tabindex]="'0' ? i === selectedItemIndex : '-1'"
+      [attr.aria-selected]="i === selectedItemIndex"
+    >
+      <span class="lg-visually-hidden">{{ description }} {{ i + 1 }}</span>
+      <span class="lg-visually-hidden" *ngIf="i === selectedItemIndex">(selected)</span>
+      <span
+        class="lg-carousel__bullet"
+        [ngClass]="{ 'lg-carousel__bullet--active': i === selectedItemIndex }"
+      ></span>
+    </button>
+    <button
+      role="tab"
+      class="lg-carousel__button"
+      (click)="nextCarouselItem()"
+      [disabled]="!loopMode && selectedItemIndex === carouselItemCount - 1"
+    >
+      <span class="lg-visually-hidden">Next {{ description }} item</span>
+      <span class="lg-carousel__button__disk lg-carousel--centered">
+        <lg-icon name="chevron-right"></lg-icon>
+      </span>
+    </button>
+  </div>
+</section>
--- a/projects/canopy/src/lib/carousel/carousel.component.scss
+++ b/projects/canopy/src/lib/carousel/carousel.component.scss
@@ -1,0 +1,64 @@
+.lg-carousel {
+  overflow: hidden;
+
+  &__wrapper {
+    position: relative;
+    display: flex;
+    align-items: stretch;
+  }
+
+  &__navigation {
+    padding-top: var(--space-sm);
+    display: flex;
+    justify-content: center;
+  }
+
+  &--centered {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+  }
+
+  &__button {
+    padding: 0;
+    border: none;
+    background-color: transparent;
+    margin: 0;
+    width: var(--space-xxl);
+    height: var(--space-xxl);
+    cursor: pointer;
+
+    &:focus-visible {
+      outline: none;
+    }
+
+    &__disk {
+      width: var(--space-xxl);
+      height: var(--space-xxl);
+      border-radius: 100%;
+    }
+
+    &:hover:not([disabled]) .lg-carousel__button__disk {
+      background-color: var(--carousel-nav-background-color);
+    }
+
+    &[disabled] {
+      cursor: default;
+    }
+  }
+
+  &__bullet {
+    width: var(--space-xxs);
+    height: var(--space-xxs);
+    background-color: var(--carousel-bullet-color);
+    display: block;
+    border-radius: 100%;
+    transition: all var(--animation-fn) var(--animation-duration);
+
+    &--active {
+      width: var(--space-xs);
+      height: var(--space-xs);
+      background-color: var(--carousel-active-bullet-color);
+    }
+  }
+}
--- a/projects/canopy/src/lib/carousel/carousel.component.ts
+++ b/projects/canopy/src/lib/carousel/carousel.component.ts
@@ -1,0 +1,65 @@
+import {
+  AfterContentInit,
+  ChangeDetectionStrategy,
+  Component,
+  ContentChildren,
+  Input,
+  QueryList,
+  ViewEncapsulation,
+} from '@angular/core';
+
+import { HeadingLevel } from '../heading';
+import { LgCarouselItemComponent } from './carousel-item/carousel-item.component';
+
+@Component({
+  selector: 'lg-carousel',
+  templateUrl: './carousel.component.html',
+  styleUrls: ['./carousel.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class LgCarouselComponent implements AfterContentInit {
+  @Input() description: string;
+  @Input() headingLevel: HeadingLevel;
+  @Input() loopMode = false;
+  @Input() slideDuration = 500;
+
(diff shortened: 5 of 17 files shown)


markblandford (approved):
Looks really good 👍

AlenNS (approved):

AlenNS:
Great job @dannyoz

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="lg-carousel-item" [ngClass]="'lg-padding--' + paddingSize">
```
andy-polhill, line 1: One convention we try and follow on Canopy is to use the wrapping element that Angular creates. So in this instance we would probably do...

```ts
@HostBinding('class.lg-carousel-item') class = true;
```

And then your template would simply be
```html
<ng-content></ng-content>
```

Meaning the final markup is..
```html
<lg-carousel-item class="lg-carousel-item">
  content
</lg-carousel-item>
```
as opposed to the slightly heavier...
```html
<lg-carousel-item>
  <div class="lg-carousel-item">
    content
  </div>
</lg-carousel-item>
```

I also wondered about removing the padding options altogether, if we provide a sensible default this could be overridden using the [padding](https://deploy-preview-386--legal-and-general-canopy.netlify.app/?path=/story/directives--padding) or [margin](https://deploy-preview-386--legal-and-general-canopy.netlify.app/?path=/story/directives--margin) directives

projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -0,0 +1,55 @@
+<section class="lg-carousel">
+  <lg-heading [level]="headingLevel" class="lg-carousel__heading lg-visually-hidden">{{
+    carouselTitle
```
andy-polhill, line 3: I wonder if this should be content projected as opposed to a string? Content projection would keep it consistent with the other components and allows a bit more flexibility.

andy-polhill (commented):
Only managed to find time for a partial review, apologies, but should be a start.

andy-polhill:
Commit wise it should probably be `feat(carousel): component added`
Where `carousel` is the scope, if you run `npm run commit` it will help build the commit message for you.

projects/canopy/src/lib/carousel/carousel.component.html
dannyoz: I think this only needs to be a string as it will just populate the heading describing the carousel for customers with screen readers

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component, HostBinding } from '@angular/core';
+@Component({
+  selector: 'lg-carousel-item',
+  templateUrl: './carousel-item.component.html',
+  styleUrls: ['./carousel-item.component.scss'],
+})
+export class LgCarouselItemComponent {
+  @HostBinding('class') classes = 'lg-carousel-item lg-padding-md';
```
andy-polhill, line 8: I think at the moment the view encapsulation is stopping the padding styles from working. Generally the approach we take is below, this should set a good baseline. 
```suggestion
  @HostBinding('class.lg-carousel-item') class = true;
```
and then in the css...
```
.lg-carousel-item {
  height: 100%;
  padding: var(--component-padding)
}
```

I think with the current approach there could be a bit of a specificity issue where `lg-padding-md` might not be override-able by the padding directive.

Historically in Angular this `@HostBinding('class')` directly to class would cause issues. It would override classes added by additional directives (in our case utility ones like the Variant one). I think that might be fixed now, but for consistency the above might be nicer.

aaronBery:
Minor note, the link to the test app in the checklist should ideally go to the test app page in the deployed netlify instance :-)

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
aaronBery, line 8: We could probably have a unit test to validate the class is applied. We've done similar things on other components if you want to refer to them

projects/canopy/src/lib/carousel/carousel.component.html
dannyoz: Resolving this as we discussed this over a call

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,57 @@
...
+import { HeadingLevel } from '../heading';
+import { LgCarouselItemComponent } from './carousel-item/carousel-item.component';
+
+@Component({
+  selector: 'lg-carousel',
+  templateUrl: './carousel.component.html',
+  styleUrls: ['./carousel.component.scss'],
+})
+export class LgCarouselComponent implements AfterContentInit {
+  @Input() carouselTitle: string;
```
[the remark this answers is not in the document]
dannyoz: I actually changed this to `description` since that is effectively what it will used for - describing the carousel to customers with screen readers

projects/canopy/src/lib/carousel/carousel.stories.ts
```diff
@@ -0,0 +1,76 @@
...
+      }),
+    ],
+    notes: {
+      markdown: notes,
+    },
+  },
+};
+
+const carouselItems = `
+<lg-carousel-item style="background-color:#f9f9f9">
```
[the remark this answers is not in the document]
dannyoz: Agreed. Removed.

dannyoz:
> Semantically I wonder if the carousel should be a list. I would need to investigate this further but I realise it's quite fundamental to the component structure so I wanted to raise this as soon as possible.
> 
> Looking at guidelines on https://www.w3.org/WAI/tutorials/carousels/structure/ it states
> 
> 'carousels are typically best represented as unordered lists, using `<ul> and <li>`'
> 
> If we were not to use list elements I wonder if we need to use appropriate aria roles

@aaronBery I did refer to those guidelines when building the component but decided to go for divs as since the component uses transclusion for the slides, also because it's a recommendation rather than a stipulation. I will look at the aria roles though

projects/canopy/src/lib/carousel/carousel.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+    range: true,
+    min: 0,
+    max: 5000,
+    step: 100,
+  };
+  return {
+    headingLevel: select('headingLevel', [1, 2, 3, 4, 5, 6], 3),
+    description: text('description', 'Example carousel'),
+    slideDuration: number('slideDuration', 500, slideDurationOptions),
+    paddingSize: select(
```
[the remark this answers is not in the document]
dannyoz: Thats old i need to remove it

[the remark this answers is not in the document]
andy-polhill: oh good spot, I think that can be removed now, padding is set to a sensible default and people can use the padding/margin directives to provide more control.

projects/canopy/src/lib/carousel/carousel.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+
+function props() {
+  const slideDurationOptions = {
+    range: true,
+    min: 0,
+    max: 5000,
+    step: 100,
+  };
+  return {
+    headingLevel: select('headingLevel', [1, 2, 3, 4, 5, 6], 3),
```
aaronBery, line 63: Is it worth making the labels more readable i.e. 'Heading level' rather than 'headingLevel'

andy-polhill:
https://github.com/Legal-and-General/canopy/pull/386#issuecomment-850455264

A list might make sense, the cleanest way is probably to provide aria roles via HostBinding https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Listitem_role. The other method us probably to switch to an attribute selector, but I think the former has a nicer dev experience and is less prone to error.

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,57 @@
...
+})
+export class LgCarouselComponent implements AfterContentInit {
+  @Input() description: string;
+  @Input() headingLevel: HeadingLevel;
+  @Input() loopMode = false;
+  @Input() slideDuration = 500;
+
+  @ContentChildren(LgCarouselItemComponent, { read: ElementRef })
+  carouselItems = new QueryList<ElementRef>();
+  selectedElementRef: ElementRef;
```
aaronBery, line 26: does this need to be public?

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    &__disk {
+      width: var(--space-xxl);
+      height: var(--space-xxl);
+      border-radius: 100%;
+    }
+
+    &:hover:not([disabled]) {
+      .lg-carousel__button__disk {
```
aaronBery, line 42: not something that particularly bothers me but I know we try to avoid more than 3 levels of nesting in our sass

dannyoz: no problem, that's easily changed

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  @HostBinding('attr.role') role = 'tab-panel';
+
+  _selected = false;
+  @HostBinding('attr.aria-selected')
+  get selected() {
+    return this._selected;
+  }
+  set selected(selected) {
+    this._selected = selected;
+  }
```
andy-polhill, line 27: Sorry @dannyoz it just occurred to me that we aren't doing anything in the getter and setter here, could it just be 
```suggestion
  @HostBinding('attr.aria-selected')
  public selected = false;
```

projects/canopy/src/lib/carousel/carousel.component.ts
dannyoz: it breaks the ngFor in the  template if set to private

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  QueryList,
+} from '@angular/core';
+
+import { HeadingLevel } from '../heading';
+import { LgCarouselItemComponent } from './carousel-item/carousel-item.component';
+
+@Component({
+  selector: 'lg-carousel',
+  templateUrl: './carousel.component.html',
+  styleUrls: ['./carousel.component.scss'],
```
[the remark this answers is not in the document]
dannyoz: Done

projects/canopy/src/lib/carousel/carousel.stories.ts
dannyoz: Fixed

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -0,0 +1,62 @@
...
+
+    &[disabled] {
+      cursor: default;
+    }
+  }
+
+  &__bullet {
+    width: var(--space-xxs);
+    height: var(--space-xxs);
+    background-color: var(--color-charcoal);
```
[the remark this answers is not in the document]
dannyoz: Theming file added

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  &__bullet {
+    width: var(--space-xxs);
+    height: var(--space-xxs);
+    background-color: var(--carousel-bullet-color);
+    display: block;
+    border-radius: 100%;
+    transition: all --animation-fn --animation-duration;
```
andy-polhill, line 56:
```suggestion
    transition: all var(--animation-fn) var(--animation-duration);
```
Unless the current approach does work?

dannyoz: Oops good catch

andy-polhill:
Just one thing holding me off thumbing, if you wouldn't mind squashing all of the commits into just that one message...

```
feat(carousel): component added
```

Due to the way semantic release works we can `merge + squash`, generally we ship new components under one commit, and then updates to components can be more granular.

andy-polhill (approved):

projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -0,0 +1,59 @@
...
+      [disabled]="!loopMode && selectedItemIndex === 0"
+    >
+      <span class="lg-visually-hidden">Previous {{ description }} item</span>
+      <span class="lg-carousel__button__disk lg-carousel__centered">
+        <lg-icon name="chevron-left"></lg-icon>
+      </span>
+    </button>
+    <button
+      role="tab"
+      class="lg-carousel__button lg-carousel__centered"
```
owensgit, line 34: Wondering if `lg-carousel__centered` should be a modifier class? E.g. `lg-carousel__button--centered`

owensgit (commented):
Great addition, thanks Dan! Just a couple of minor questions from me

projects/canopy/src/lib/carousel/carousel.notes.ts
```diff
@@ -0,0 +1,42 @@
...
+  imports: [ ..., LgCarouselModule ],
+})
+~~~
+
+and in your HTML:
+
+~~~html
+<lg-carousel [description]="description" [headingLevel]="headingLevel" [loopMode]="true">
+  <lg-carousel-item>
+    <!-- Carousel item 1 content goes here -->
```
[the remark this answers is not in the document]
dannyoz: Good one,  I've changed it so the comments have been removed

projects/canopy/src/lib/carousel/carousel.component.html
dannyoz: Yes this should indeed be a modifier class

dannyoz:
> Great addition, thanks Dan! Just a couple of minor questions from me

Thanks @owensgit I made your suggested changes

andy-polhill (approved):

owensgit (approved):
Nice work @dannyoz!

PR #411, opened 2021-06-14
fix(typography): enable lyon font path setting

### LINKED ISSUE #409: Unable to include styles in non-angular project
The current settings in typography.scss assume a directory structure that some projects do not have (i.e. /assets/fonts).  This causes webpack builds to fail as it is difficult/impossible to store the Lyon fonts in a location that can be found. 

The simple solution to this is to allow the variable '$fonts-path-lyon' to be overridden when imported, in the same way as '$fonts-path' currently allows, whilst keeping the original default values in place to prevent breaking changes.

Hopkinsa:
# Description
The current settings in typography.scss assume a directory structure that some projects do not have (i.e. /assets/fonts). This causes webpack builds to fail as it is difficult/impossible to store the Lyon fonts in a location that can be found.

Fixes #409


elenagarrone:
@Hopkinsa I think something is wrong with the PR - I don't think you changes 1071 files 😅

Hopkinsa:
> @Hopkinsa I think something is wrong with the PR - I don't think you changes 1071 files 😅



> @Hopkinsa I think something is wrong with the PR - I don't think you changes 1071 files 😅
Something definitely wrong, thats not what I pushed!  Will have a look to see what's happened

Hopkinsa:
Incorrect files included in PR

Issue #331, opened 2021-04-26
Canopy Test App shows blank screen on startup


ABucin:
The demo app shows a white screen and the following error(s) in the dev console:

<img width="513" alt="Screenshot 2021-04-26 at 12 32 52" src="https://user-images.githubusercontent.com/4033816/116061608-bdce2200-a68b-11eb-82b0-ba907c86b6b3.png">

andy-polhill:
This is a recurrence of https://github.com/Legal-and-General/canopy/issues/79

andy-polhill:
I think there are two stages to this issue
- The way the icon registry is loaded into the test project means there are effectively two instances of it.
- There is an ExpressionChangedAfterItHasBeenCheckedError in the table component

pauleustice:
Closing as no longer an issue.

PR #416, opened 2021-06-17
feat(modal): add new component

elenagarrone:
# Description

Add the new modal component. 

Storybook link: https://deploy-preview-416--legal-and-general-canopy.netlify.app/?path=/story/components-modal--standard
Design link: https://legalandgeneral.invisionapp.com/console/Canopy-%E2%80%93-Modals-ckpico8se00ap010nhbdd3xlr/ckpico9g901ke01vyci0r1h5k
Screenshot:
Screenshot 2021-06-17 at 18 54 12

Screenshot 2021-06-17 at 18 56 59

# Checklist:


projects/canopy/src/lib/modal/modal-header/modal-header.component.html
```diff
@@ -0,0 +1,13 @@
+<lg-heading [level]="headingLevel" class="lg-modal-header__title">
```
owensgit, line 1: The class isn't needed here, right?

projects/canopy/src/lib/modal/modal-header/modal-header.component.scss
```diff
@@ -0,0 +1,42 @@
...
+  @include lg-breakpoint(md) {
+    padding: var(--space-md);
+  }
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
```
owensgit, line 18: Just a thought - I'm starting to wonder if having that class input on `lg-header` was actually quite useful 🤔 , as quite often in Canopy and when using Canopy, I find that I need override the heading styles for some reason. I can't remember the rationale in removing it now 😄

projects/canopy/src/lib/modal/modal.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  onModalClick(event: Event): void {
+    event.stopPropagation();
+  }
+
+  ngOnDestroy(): void {
+    this.modalService.remove(this.id);
+    this.subscriptions.forEach((subscription) => subscription.unsubscribe());
```
owensgit, line 98: Just wondering what you think of the `takeUntil(this.ngUnsubscribe)` pattern for unsubscribing to observables, rather than looping over observables to unsubscribe? Perhaps we could adopt that here? I think this implementation is fine though.

projects/canopy/src/lib/modal/modal-trigger/modal-trigger.directive.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor(private el: ElementRef, private modalService: LgModalService) {}
+
+  ngOnInit(): void {
+    this.subscription = this.modalService
+      .isOpen$(this.lgModalTrigger)
+      .pipe(
+        map((isOpen) => {
+          if (!isOpen && this.allowFocusOnModalTrigger) {
+            const triggerEl = this.el.nativeElement as HTMLElement;
```
andy-polhill, line 40: Minor, but can you drop the assignment?
```suggestion
            (this.el.nativeElement as HTMLElement).focus();
```

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -567,6 +567,17 @@
               </tbody>
             </table>
 
+            <lg-heading level="3">Modal</lg-heading>
+
+            <button lgModalTrigger="modal-story" lg-button type="button" variant="outline-primary">Open modal</button>
+            <lg-modal id="modal">
```
owensgit, line 573: I think the `id` here needs to match the `lgModalTrigger` value above?

projects/canopy/src/lib/modal/modal.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  ngAfterContentInit(): void {
+    this.modalHeaderComponent.id = `lg-modal-header-${this.id}`;
+    this.modalHeaderComponent.modalId = this.id;
+    this.modalBodyComponent.id = `lg-modal-body-${this.id}`;
+  }
+
+  @HostListener('keydown', ['$event']) onKeydown(event: KeyboardEvent): void {
+    if (event.key === keyName.KEY_ESCAPE && this.isOpen && this.closeOnOverlayClick) {
```
owensgit, line 78: Just wondering if `this.closeOnOverlayClick` needs to be `true` for this one? Is it an accessibility issue to not allow a user to dismiss it with the ESC key? Or is this for instances where the user has to click on a specific button, or perform a specific action to get past the modal?

projects/canopy/src/lib/modal/modal.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  @Input() closeOnOverlayClick = true;
+  @Output() open: EventEmitter<void> = new EventEmitter();
+  @Output() closed: EventEmitter<void> = new EventEmitter();
+  @ContentChild(forwardRef(() => LgModalHeaderComponent))
+  modalHeaderComponent: LgModalHeaderComponent;
+  @ContentChild(forwardRef(() => LgModalBodyComponent))
+  modalBodyComponent: LgModalBodyComponent;
+
+  isOpen: boolean;
+  subscriptions: Array<Subscription> = [];
```
andy-polhill, line 42: Just wondering about the array of subscriptions? it seems like there can only ever be one at the moment.

projects/canopy/src/lib/modal/modal.component.scss
```diff
@@ -0,0 +1,54 @@
...
+  animation-duration: var(--animation-duration);
+  animation-name: opacity;
+  left: 50%;
+  max-height: 100%;
+  overflow-y: auto;
+  position: fixed;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  width: 90%;
+  z-index: 1001;
```
owensgit, line 32: Perhaps we should store the z-index as a var in the main variables file? I'm thinking of when we have other items on the z-index, such as popovers/tooltips etc, we could ensure the modal is always the highest number. Could be something like `--zindex-modal`, then later the could be a `--index-tooltip`, which we would ensure is lower.

projects/canopy/src/lib/modal/modal.notes.ts
```diff
@@ -0,0 +1,89 @@
...
+  imports: [ ..., LgModalModule ],
+})
+~~~
+
+and in your HTML:
+
+~~~html
+<button lgModalTrigger="modal-story" lg-button type="button" [variant]="'outline-primary'">Open modal</button>
+
+<lg-modal id="modal-story" [closeOnOverlayClick]="closeOnOverlayClick" >
```
andy-polhill, line 21: I think I mentioned previously but I wonder if we could get a design/uk call on...
 - closeOnOverlayClick
 - showCloseButton

 It could be flexibility that we need, but if we don't, I wonder if we can drop them for simplicity and consistency.

projects/canopy/src/lib/modal/modal.component.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+      const event = new KeyboardEvent('keydown', {
+        key: keyName.KEY_UP,
+      });
+      component.onKeydown(event);
+
+      verify(modalServiceMock.close(id)).never();
+      expect().nothing();
+    });
+
+    it("shouldn't close the modal when the modal is already close", () => {
```
owensgit, line 133: Super minor 🙂 

```suggestion
    it("shouldn't close the modal when the modal is already closed", () => {
```

projects/canopy/src/lib/modal/modal.service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from '@angular/core';
+
+import { BehaviorSubject, Observable } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LgModalService {
+  private states: Map<string, BehaviorSubject<boolean>> = new Map();
```
andy-polhill, line 9: Oooh I still get excited when I see a `Map` or `Set` 🎉

package.json
```diff
@@ -35,6 +35,7 @@
   "private": false,
   "dependencies": {
     "@angular/animations": "^11.1.0",
+    "@angular/cdk": "^11.2.13",
```
andy-polhill, line 38: Just wondering if we should add a short one-liner to the contribution docs 🤔. Just so people know that the cdk resources are available to play with.

projects/canopy-test-app/src/app/app.component.html
```diff
@@ -567,6 +567,17 @@
               </tbody>
             </table>
 
+            <lg-heading level="3">Modal</lg-heading>
+
+            <button lgModalTrigger="modal-story" lg-button type="button" variant="outline-primary">Open modal</button>
```
andy-polhill, line 572:
```suggestion
            <button lgModalTrigger="modal" lg-button type="button" variant="outline-primary">Open modal</button>
```
Should it match the id of the modal? (I know the test page is bust #331 )

owensgit (commented):
Great work @elenagarrone! 🕺  Just a few questions/suggestions from me

projects/canopy/src/lib/modal/modal.module.ts
```diff
@@ -0,0 +1,51 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import { A11yModule } from '@angular/cdk/a11y';
```
andy-polhill, line 3: Is there a way to only bring in the `cdkFocusTrap`? or is it all linked with the other `a11y` code, just concerned it might bloat the build size otherwise

andy-polhill (commented):
Awesome work @elenagarrone, a few comments but nothing to major hopefully.

projects/canopy/src/lib/modal/modal.component.scss
andy-polhill: Good call, we should probably have some set z-indexes for certain tasks.

LIghtning do this.. https://www.lightningdesignsystem.com/design-tokens/#category-z-index

Might be some insight in here too..
https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/

projects/canopy/src/lib/modal/modal-header/modal-header.component.scss
elenagarrone: I think it was because it was allowing people to add their own classes 🤔  maybe we should add a hardcoded class like we do for all the other components instead. It will make these type of things a lot nicer to target.

projects/canopy/src/lib/modal/modal.component.ts
elenagarrone: To be honest I don't mind too much, it's just what i've been used to use. I just find it a bit more readable to do:
```
subscriptions.push(all the subscriptions)
```
but that's probably only because I'm used to it.
Maybe worth opening a discussion? it will have to be changed across the codebase anyway.

projects/canopy-test-app/src/app/app.component.html
elenagarrone: doh yes ;)

projects/canopy/src/lib/modal/modal.component.ts
elenagarrone: Sorry I'm not sure I completely understand.
It's a requirement that if the overlay is dismissible by clicking outside the modal a keyboard user should have the same functionality by clicking `ESC`. Same is for the opposite: if a specific action is required on the modal and we don't want the user to dismiss the modal with a click outside then the same should happen for the keyboard user.

Does any of the above clarifies your questions?

projects/canopy/src/lib/modal/modal.component.scss
elenagarrone: Yes, nice one 👍

projects/canopy/src/lib/canopy.module.ts
```diff
@@ -45,6 +45,7 @@ import {
   LgToggleModule,
   LgValidationModule,
 } from './forms';
+import { LgModalModule } from './modal/modal.module';
```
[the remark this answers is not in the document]
elenagarrone: I don't think se don't but I will 🙂  I'm assuming it would be in alphabetical order of the path 🤔

projects/canopy/src/lib/modal/modal.notes.ts
elenagarrone: I would definitely keep the `showCloseButton` because if you have a short modal that has a heading, one single line of text and the "close" and "cta" buttons the cross at the top is completely unnecessary and it actually makes it an annoyance for users using screen readers.

I've added `closeOnOverlayClick` because I remembered stakeholders asking for that exact functionality when I was working on a previous project. It's when a user has to make a decision.

@alexcanning do you have any views on both please?

projects/canopy/src/lib/modal/modal.module.ts
elenagarrone: Good point, I can just include `CdkTrapFocus`

projects/canopy/src/lib/modal/modal.component.ts
owensgit: Yes thanks, I also saw another comment below about the user making a decision on some modals, that makes sense as well 👍

projects/canopy/src/lib/modal/modal.component.ts
owensgit: Yes probably worth a discussion, not really something for this PR :)

projects/canopy/src/lib/modal/modal-header/modal-header.component.html
elenagarrone: Removed

projects/canopy-test-app/src/app/app.component.html
elenagarrone: Updated 👍

projects/canopy/src/lib/modal/modal.component.scss
elenagarrone: I've added them to the main variables file. As we add more we could split them in a separate file.

projects/canopy/src/lib/modal/modal.component.spec.ts
elenagarrone: Updated

projects/canopy/src/lib/canopy.module.ts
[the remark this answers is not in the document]
elenagarrone: Done

projects/canopy/src/lib/modal/modal-trigger/modal-trigger.directive.ts
elenagarrone: Updated 👍

projects/canopy/src/lib/modal/modal.component.ts
elenagarrone: Updated to use `subscription` 🙂

package.json
elenagarrone: I've added:
```
# Angular Material CDK
The `@angular/cdk` can be used as long as only the necessary functionality is imported. 
```

projects/canopy-test-app/src/app/app.component.html
elenagarrone: Done

projects/canopy/src/lib/modal/modal.module.ts
elenagarrone: Updated 👍

projects/canopy/src/lib/modal/modal.notes.ts
alexcanning: We've written some guidance around the close button which can be read here -https://landg.atlassian.net/wiki/spaces/CAN/pages/857617703/Modal+Dialogue. It basically says, always show the close button.

I'd also have thought that clicking outside of the modal would be expected behaviour - performing the same dismiss action as close.

elenagarrone: Perfect, I'll update that :)

projects/canopy/src/lib/modal/modal.module.ts
elenagarrone: I take this back, the build failed because the `cdkFocusTrap` is already imported by `A11yModule` so I will have to import all of it.

projects/canopy/src/styles/utils.scss
```diff
@@ -7,3 +7,18 @@
 .lg-unstyled-link {
   @include lg-unstyled-link;
 }
+
+.lg-footer-buttons {
```
andy-polhill, line 11: I wonder does this need a class ? I feel like the standard behaviour would always be to have this margin when two buttons or links are adjacent to one another.

There may be some exceptions in existing components but we could probably find them..
https://deploy-preview-416--legal-and-general-canopy.netlify.app/?path=/story/components-form-input--with-multiple-button-suffixes

elenagarrone: Do you mean to put this in the buttons styles?

elenagarrone: I thought about adding it to the button component but I feel like it's not the right thing to do. The button component handles the style for a single button and if we start adding all the different handling of multiple buttons it might get bloated.

elenagarrone: Discussing this with @andy-polhill he suggested to have a `lg-button-group` component so that we don't have to manually add the class and it's more built into the API.

andy-polhill: I wonder if it could be a simple component..
```html
<lg-button-group>
  <lg-button />
  <lg-button />
</lg-button-group>
```

My only concern with the class is that it isn't something we have used much before, a developer in an application would have to step out of the normal Canopy paradigms to add that class. There is also a risk they may not know it exists.

elenagarrone: Done 👍

projects/canopy/src/lib/modal/modal-trigger/modal-trigger.directive.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  constructor(private el: ElementRef, private modalService: LgModalService) {}
+
+  ngOnInit(): void {
+    this.subscription = this.modalService
+      .isOpen$(this.lgModalTrigger)
+      .pipe(
+        map((isOpen) => {
+          if (!isOpen && this.allowFocusOnModalTrigger) {
```
paul-request, line 39: You could remove the `if` and use a filter in the pipe. Not a blocker though

paul-request: Also, should the focus happen in the subscribe, rather than a map? 🤔

projects/canopy/src/lib/modal/modal.service.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { spy, verify } from 'ts-mockito';
+import { Subscription } from 'rxjs';
+
+import { LgModalService } from './';
+
+describe('LgModalService', () => {
+  let service: LgModalService;
+  let serviceSpy: LgModalService;
+  const id = 'test-1';
```
paul-request, line 11: I'm surprised prettier doesn't re-order these 🤔

projects/canopy/src/lib/modal/modal-trigger/modal-trigger.directive.ts
elenagarrone: yes, good point

projects/canopy/src/lib/modal/modal.service.spec.ts
elenagarrone: I'll do it :D

projects/canopy/src/lib/modal/modal-trigger/modal-trigger.directive.ts
elenagarrone: @paul-request updated

projects/canopy/src/lib/modal/modal.service.spec.ts
elenagarrone: Updated

paul-request (dismissed):

projects/canopy/src/lib/button/button-group/button-group.component.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../../styles/mixins.scss';
+
+.lg-button-group {
```
andy-polhill, line 3:
```suggestion
.lg-btn-group {
```
Should we abbreviate button again? for consistency?

paul-request (approved):

andy-polhill (approved):

PR #424, opened 2021-06-22
feat(auto play carousel): added auto play carousel feature

Nupur04:
# Description

Added optional autoPlay mode to Carousel component. This adds a timer with pause and play button to give control to the user. AutoplayDelay is a parameter that can be defined to set the timer.
 
Design link: https://legalandgeneral.invisionapp.com/overview/Canopy-Carousel-ckou4o2im00m9013db6ww6lfv/screens?share=&sortBy=1&sortOrder=1&viewLayout=2

Screenshot: https://www.awesomescreenshot.com/video/4219164?key=a2993ad195e64c47115859e396c26b28

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
+++ b/projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
@@ -1,0 +1,18 @@
+<button
+  class="lg-carousel-autoplay__btn lg-carousel-autoplay__btn--pause"
+  *ngIf="!pause.value"
+  (click)="pause.next(true)"
+  type="button"
+>
+  <span class="lg-visually-hidden">Stop Animation</span>
+  <lg-icon name="pause"></lg-icon>
+</button>
+<button
+  class="lg-carousel-autoplay__btn lg-carousel-autoplay__btn--play"
+  *ngIf="pause.value"
+  (click)="pause.next(false)"
+  type="button"
+>
+  <span class="lg-visually-hidden">Start Animation</span>
+  <lg-icon name="play"></lg-icon>
+</button>
--- a/projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
+++ b/projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
@@ -1,0 +1,16 @@
+.lg-carousel-autoplay {
+  position: absolute;
+  right: var(--space-sm);
+  z-index: var(--z-index-carousel-autoplay);
+  top: var(--space-sm);
+
+  &__btn {
+    background-color: var(--color-white);
+    border-radius: 50%;
+    border: none;
+    width: var(--space-xxl);
+    height: var(--space-xxl);
+    font-size: var(--text-fs-3-size);
+    cursor: pointer;
+  }
+}
--- a/projects/canopy/src/lib/carousel/auto-play/auto-play.component.spec.ts
+++ b/projects/canopy/src/lib/carousel/auto-play/auto-play.component.spec.ts
@@ -1,0 +1,66 @@
+import { DebugElement } from '@angular/core';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
+import { MockComponents } from 'ng-mocks';
+
+import { BehaviorSubject } from 'rxjs';
+
+import { LgIconComponent } from '../../icon';
+
+import { LgAutoplayComponent } from './auto-play.component';
+
+describe('LgButtonComponent', () => {
+  let component: LgAutoplayComponent;
+  let fixture: ComponentFixture<LgAutoplayComponent>;
+  let pause: DebugElement;
+  let play: DebugElement;
+
+  beforeEach(
+    waitForAsync(() => {
+      TestBed.configureTestingModule({
+        declarations: [LgAutoplayComponent, MockComponents(LgIconComponent)],
+      }).compileComponents();
+    }),
+  );
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(LgAutoplayComponent);
+    component = fixture.componentInstance;
+    component.pause = new BehaviorSubject<boolean>(false);
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should apply initial host binding classes and attributes', () => {
+    expect(
+      component.element.nativeElement.classList.contains('lg-carousel-autoplay'),
+    ).toBe(true);
+  });
+
+  it('should pause the autoPlay functionality', () => {
+    pause = fixture.debugElement.query(
+      By.css('.lg-carousel-autoplay__btn.lg-carousel-autoplay__btn--pause'),
+    );
+    const spy = spyOn(component.pause, 'next');
+
+    pause.nativeElement.click();
+    expect(spy).toHaveBeenCalledWith(true);
+  });
+
+  it('should play autoPlay functionality', () => {
+    component.pause = new BehaviorSubject<boolean>(true);
+    fixture.detectChanges();
+
+    play = fixture.debugElement.query(
+      By.css('.lg-carousel-autoplay__btn.lg-carousel-autoplay__btn--play'),
+    );
+    const spy = spyOn(component.pause, 'next');
+    play.nativeElement.click();
+
+    expect(spy).toHaveBeenCalledWith(false);
+  });
+});
--- a/projects/canopy/src/lib/carousel/carousel.component.html
+++ b/projects/canopy/src/lib/carousel/carousel.component.html
@@ -2,6 +2,7 @@
   <lg-heading [level]="headingLevel" class="lg-carousel__heading lg-visually-hidden">{{
     description
   }}</lg-heading>
+  <lg-auto-play *ngIf="autoPlay" [pause]="pause"></lg-auto-play>
   <div
     class="lg-carousel__active-content lg-visually-hidden"
     aria-atomic="true"
@@ -43,6 +44,18 @@
         class="lg-carousel__bullet"
         [ngClass]="{ 'lg-carousel__bullet--active': i === selectedItemIndex }"
       ></span>
+      <svg
+        class="lg-carousel__button-timer"
+        *ngIf="autoPlay && i === selectedItemIndex && !pause.value"
+      >
+        <circle
+          class="lg-carousel__button-timer__circle"
+          [style.animation-duration.s]="autoPlayDelay / 1000"
+          r="18"
+          cx="20"
+          cy="20"
+        ></circle>
+      </svg>
     </button>
     <button
       role="tab"
--- a/projects/canopy/src/lib/carousel/carousel.component.scss
+++ b/projects/canopy/src/lib/carousel/carousel.component.scss
@@ -1,5 +1,40 @@
+@import '../../styles/mixins.scss';
+
 .lg-carousel {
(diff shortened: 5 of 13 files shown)


projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -2,6 +2,31 @@
...
+
+  <div
+    *ngIf="autoPlay"
+    aria-live="polite"
+    aria-atomic="true"
+    class="lg-carousel__autoplay"
+  >
+    <button
+      class="lg-carousel__autoplay__button lg-carousel__pause"
+      *ngIf="!pause.value"
```
elenagarrone, line 14: I would add a `type=button` to these buttons otherwise the default is submit :)

projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -2,6 +2,31 @@
   <lg-heading [level]="headingLevel" class="lg-carousel__heading lg-visually-hidden">{{
     description
   }}</lg-heading>
+
+  <div
+    *ngIf="autoPlay"
+    aria-live="polite"
+    aria-atomic="true"
+    class="lg-carousel__autoplay"
```
elenagarrone, line 10: I'm wondering if this autoplay should be in its own component?

I would also change the classes to use a single element. For example I would have:
```
lg-carousel-autoplay
lg-carousel-autoplay__btn
lg-carousel-autoplay__btn--pause
lg-carousel-autoplay__btn--play
```

projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -2,6 +2,31 @@
...
+  >
+    <button
+      class="lg-carousel__autoplay__button lg-carousel__pause"
+      *ngIf="!pause.value"
+      (click)="pause.next(true)"
+    >
+      <span class="lg-visually-hidden">Stop Animation </span>
```
elenagarrone, line 17: Can we remove the space in both the visually hidden spans please? If you have added it for spacing you can use the `lgMargin` directive instead.

projects/canopy/src/lib/carousel/carousel.notes.ts
```diff
@@ -39,4 +39,6 @@ and in your HTML:
 | \`\`slideDuration\`\` | Duration in milliseconds of the transition between slides.                                               | number  |   500     |   No     |
 | \`\`description\`\`   | Text used to describe the carousel content for screen readers. This is visually hidden.                  | string  | undefined |   Yes    |
 | \`\`headingLevel\`\`  | The heading level of the description: \`\`1\`\`, \`\`2\`\`, \`\`3\`\`, \`\`4\`\`, \`\`5\`\`, \`\`6\`\`   | number  | undefined |   Yes    |
+| \`\`autoPlay\`\`      | Enables auto play mode when set to true.                                                                 | boolean |   false   |   No     |
+| \`\`autoPlayDelay\`\` | Delay time in miliseconds to switch to next slide when autoPlay mode is On.                              | number  |   2000    |   No     |
```
elenagarrone: I guess because the `autoPlay` input is a boolean I would say `Delay time in milliseconds to switch to next slide when autoPlay mode is set to true.`?
There is also a tiny typo in `miliseconds` -> `milliseconds `

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -1,5 +1,61 @@
...
+    &__button {
+      background-color: var(--color-white);
+      border-radius: 50%;
+      border: none;
+      width: var(--space-xxl);
+      height: var(--space-xxl);
+      font-size: var(--text-fs-3-size);
+      cursor: pointer;
+
+      &[keyboard-focus]:focus {
```
elenagarrone, line 22: Canopy shouldn't handle the `keyboard-focus`. I think we are moving away from having focus only on keyboard so this should be changed to allow focus normally.

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -1,5 +1,61 @@
+@import '../../styles/mixins.scss';
+
 .lg-carousel {
   overflow: hidden;
+  position: relative;
+
+  &__autoplay {
+    position: absolute;
+    right: var(--space-sm);
+    z-index: 1;
```
elenagarrone, line 10: Could you add this to the z-index list in the main variable file please (make sure you rebase).

projects/canopy/src/lib/carousel/carousel.component.scss
```diff
@@ -1,5 +1,61 @@
...
+      font-size: var(--text-fs-3-size);
+      cursor: pointer;
+
+      &[keyboard-focus]:focus {
+        @include lg-outer-focus-outline();
+      }
+    }
+  }
+
+  &__button-timer {
```
elenagarrone, line 28: Looks like the `play` button is not fully centred:
Screenshot 2021-06-22 at 15 44 33

elenagarrone (commented):
Thanks for doing this @Nupur04. Just some comments from me 🙂

projects/canopy/src/lib/carousel/carousel.component.html
Nupur04: I'll remove the space. Its not needed

projects/canopy/src/lib/carousel/carousel.component.scss
Nupur04: Yes, I just saw in some places we are have it and then I used it ;)

projects/canopy/src/lib/carousel/auto-play/auto-play.component.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(LgAutoplayComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
owensgit, line 17: Just wondering if we should add some tests here to cover the visibility of the buttons, based on the condition: `*ngIf="pause.value"` for example.

Perhaps we can also have a test to ensure the `pause` observable is changed on click?

projects/canopy/src/lib/carousel/carousel.component.scss
owensgit: Same for the pause icon too :)

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -58,8 +69,31 @@ export class LgCarouselComponent implements AfterContentInit {
     }
   }
 
+  setAutoPlayInterval(): void {
+    this.pausableTimer$ = defer(() => {
+      return interval(this.autoPlayDelay).pipe(
+        takeUntil(this.unsubscribe),
+        withLatestFrom(this.pause),
+        filter(([v, paused]) => !paused),
```
owensgit, line 77: Not sure what `v` is in this `filter` operator? Is the interval? Just wondering if it should be `interval` instead?

projects/canopy/src/styles/variables/_main.scss
```diff
@@ -49,4 +49,7 @@
 
   /* Separator */
   --separator-color: var(--color-platinum);
+
+  /* z-index */
+  --z-index-1: 1;
```
owensgit, line 54: I think we have a z-index section already, may have been merged after your started this work. If you rebase on master, you should see these two:

```
/* z-index */
--z-index-modal-overlay: 1000;
--z-index-modal: 1010;
```

So you could add this new one at the top of this list.

owensgit (commented):
Thanks @Nupur04, nice work! A few questions/comments from me

projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
```diff
@@ -0,0 +1,20 @@
+<div aria-live="polite" aria-atomic="true" class="lg-carousel-autoplay">
```
owensgit, line 1: One thing we tend to to with component wrappers is apply them directly to the host element, using `HostBinding`.

E.g.

```
@HostBinding('class.lg-carousel-autoplay') class = true;
@HostBinding('attr.aria-live') role = 'polite';
@HostBinding('attr.aria-atomic') role = 'true';
```

dannyoz:
Hi @elenagarrone and @owensgit You both commented on the play Icon not being aligned to the centre.  It actually is a centred canopy icon but looks like an optical illusion to me due to the shape of the svg fill.

We could add a left margin to the icon to shift it so it looks more centred, but adding css to fix the position of the icon feels like hacky bad practice to me.

We do have an Icon called play-spot which looks like the play is more centred but we don't have the pause equivalent. I think a better option than adding margin to the icon would be to leave it as it is for now but get a pause-spot designed and added and then switch to that version as a future enhancement.

What do you guys think?

Screenshot 2021-06-23 at 10 15 16

projects/canopy/src/lib/carousel/carousel.component.html
Nupur04: yes, moved it in component and updated the class names.

projects/canopy/src/lib/carousel/carousel.component.ts
Nupur04: Its not needed. Removed

owensgit:
> Hi @elenagarrone and @owensgit You both commented on the play Icon not being aligned to the centre. It actually is a centred canopy icon but looks like an optical illusion to me due to the shape of the svg fill.
> 
> We could add a left margin to the icon to shift it so it looks more centred, but adding css to fix the position of the icon feels like hacky bad practice to me.
> 
> We do have an Icon called play-spot which looks like the play is more centred but we don't have the pause equivalent. I think a better option than adding margin to the icon would be to leave it as it is for now but get a pause-spot designed and added and then switch to that version as a future enhancement.
> 
> What do you guys think?

Thanks for looking into this. I think that's a good idea, IMO we can live with a slight misalignment for now. Alternatively, we could see the CSS hack as a temporary solution until we have proper icons, but I wouldn't mind if we didn't. What do you think @elenagarrone?

Would you be okay to add a new issue to work on the pause-spot icon designed and incorporated so we don't forget?

owensgit (approved):

dannyoz:
> > We do have an Icon called play-spot which looks like the play is more centred but we don't have the pause equivalent. I think a better option than adding margin to the icon would be to leave it as it is for now but get a pause-spot designed and added and then switch to that version as a future enhancement.
> > What do you guys think?
> 
> Thanks for looking into this. I think that's a good idea, IMO we can live with a slight misalignment for now. Alternatively, we could see the CSS hack as a temporary solution until we have proper icons. What do you think @elenagarrone?
> 
> @dannyoz Would you be okay to add a new issue to work on the pause-spot icon designed and incorporated so we don't forget?

I can do that sure 👍

projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
```diff
@@ -0,0 +1,16 @@
+.lg-carousel-autoplay {
+  position: absolute;
+  right: var(--space-sm);
+  z-index: var(--z-index-autoPlay);
```
elenagarrone, line 4: I can't see this variable anywhere? the main.scss has `z-index` which I'm assuming has to be matching this?

Could you change the name to be `--z-index-carousel-autoplay` please?

owensgit (approved):

elenagarrone (approved):
Great work!

PR #437, opened 2021-06-28
feat(table): add stackedOnMobile option and long copy version of table

owensgit:
# Description

Adds new option to `lg-table-cell` called `stackSm`. If set, this option will stack the labels on top of the cell content, instead of side by side. E.g. 

```
<td lg-table-cell [stackSm]="true">
```

Also introduces a border to the top of the table.

A new story component is added to render a table with a large amount of copy in each cell, to catch these designs: https://legalandgeneral.invisionapp.com/console/share/JKZ6TDGSC4G/597020691/inspect

~Note this is still draft, will need to update notes and unit tests once the general approach is approved.~

Ready for full review 🙂 

### stackedOnMobile set to false (default)
Screenshot 2021-06-28 at 08 29 04

### stackedOnMobile set to true
Screenshot 2021-06-28 at 08 29 19

----

Design link: https://legalandgeneral.invisionapp.com/console/share/JKZ6TDGSC4G/597020691/inspect

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/table/table-cell/table-cell.component.html
+++ b/projects/canopy/src/lib/table/table-cell/table-cell.component.html
@@ -9,9 +9,10 @@
 </span>
 <div
   #content
+  class="lg-table-cell__content"
   [ngClass]="{
     'lg-table-cell__content--align-end': align === alignOptions.End,
-    'lg-table-cell__content': showLabel
+    'lg-table-cell__content--hidden-label': !showLabel
   }"
 >
   <ng-content></ng-content>
--- a/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
+++ b/projects/canopy/src/lib/table/table-cell/table-cell.component.scss
@@ -5,6 +5,26 @@
   display: flex;
   align-items: center;
 
+  &--stacked {
+    flex-direction: column;
+    align-items: stretch;
+
+    .lg-table-cell__label {
+      margin-bottom: var(--space-xxs);
+    }
+
+    .lg-table-cell__content {
+      margin-left: 0;
+      margin-bottom: var(--space-md);
+    }
+
+    &:last-child {
+      .lg-table-cell__content {
+        margin-bottom: 0;
+      }
+    }
+  }
+
   &--toggle-cell {
     order: 1;
     padding: 0;
@@ -21,7 +41,10 @@
 
 .lg-table-cell__content {
   margin-left: auto;
-  white-space: nowrap;
+}
+
+.lg-table-cell__content--hidden-label {
+  margin-left: 0;
 }
 
 .lg-table-cell--expandable-content {
@@ -56,6 +79,7 @@
 
       &__content {
         @include lg-breakpoint($columns-breakpoint) {
+          margin-bottom: 0;
           text-align: left;
         }
 
--- a/projects/canopy/src/lib/table/table-cell/table-cell.component.ts
+++ b/projects/canopy/src/lib/table/table-cell/table-cell.component.ts
@@ -101,6 +101,10 @@ export class LgTableCellComponent {
     return this._showLabel;
   }
 
+  @HostBinding('class.lg-table-cell--stacked')
+  @Input()
+  stack = false;
+
   private _align: AlignmentOptions = AlignmentOptions.Start;
 
   private _showLabel = true;
--- a/projects/canopy/src/lib/table/table.notes.ts
+++ b/projects/canopy/src/lib/table/table.notes.ts
@@ -1,10 +1,10 @@
-// Document the lower components
-
 export const notes = `
 # Table Component
 
 ## Purpose
-The table component provides a way to present tabular data in a responsive format.
+The table component provides a way to present tabular data in a responsive format. On small screens, the layout of the table changes to display the table headers alongside each value, providing an easier way for users to read the data as they scroll down the list of rows.
+
+There are several options that allow you to customise the table behaviour at for different uses cases and screen sizes.
 
 ## Usage
 ~~~js
@@ -14,7 +14,35 @@ The table component provides a way to present tabular data in a responsive forma
 })
 ~~~
 
-and in your HTML:
+### Simple example
+
+~~~html
+<table lg-table>
+  <thead lg-table-head>
+    <tr lg-table-row>
+      <th lg-table-head-cell>Author</th>
+      <th lg-table-head-cell>Book</th>
+      <th lg-table-head-cell>Published</th>
+    </tr>
+  </thead>
+  <tbody lg-table-body>
+    <tr lg-table-row>
+      <td lg-table-cell>Orhan Pamuk</td>
+      <td lg-table-cell>Strangeness in my Mind</td>
+      <td lg-table-cell>2016</td>
+    </tr>
+    <tr lg-table-row>
+      <td lg-table-cell>Albert Camus</td>
+      <td lg-table-cell>The Plague</td>
+      <td lg-table-cell>1947</td>
+    </tr>
+  </tbody>
+</table>
+~~~
+
+### Table with expandable detail rows
+
+This table adds the \`\`LgTableRowToggleComponent\`\`, and then an expandable detail row. This creates a table that allows the user to expand a row for more information. This can be seen on the <a href="/story/components-table--detail">Expandable Detail story</a>.
 
 ~~~html
 <table lg-table>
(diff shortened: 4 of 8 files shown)


projects/canopy/src/lib/table/table-cell/table-cell.component.scss
```diff
@@ -21,7 +41,11 @@
 
 .lg-table-cell__content {
   margin-left: auto;
-  white-space: nowrap;
+  // white-space: nowrap;
```
aaronBery, line 44: just a reminder to remove this :-)

projects/canopy/src/lib/table/table-cell/table-cell.component.ts
```diff
@@ -101,6 +101,12 @@ export class LgTableCellComponent {
     return this._showLabel;
   }
 
+  @Input() stackedOnMobile: boolean;
+
+  @HostBinding('class.lg-table-cell--stacked') get stackedlass() {
```
aaronBery, line 106: typo, should be `stackedClass`

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
owensgit: Thanks, done ✅

projects/canopy/src/lib/table/table-cell/table-cell.component.ts
owensgit: Nice spot - fixed ✅

elenagarrone:
I'm not sure I like the naming `stakedOnMobile`, what about something like 'stackSm' or `stackSmDevice`

owensgit:
> I'm not sure I like the naming `stakedOnMobile`, what about something like 'stackSm' or `stackSmDevice`

Ah yes, good point. I like `stackSm`. I'll change to that unless I get any more suggestions.

andy-polhill:
> > I'm not sure I like the naming `stakedOnMobile`, what about something like 'stackSm' or `stackSmDevice`
> 
> Ah yes, good point. I like `stackSm`. I'll change to that unless I get any more suggestions.

On the segment component we have `[stack]="sm"`, I wonder if we can copy that convention?
https://legal-and-general-canopy.netlify.app/?path=/story/components-form-segment--standard

owensgit:
> On the segment component we have `[stack]="sm"`, I wonder if we can copy that convention?
> https://legal-and-general-canopy.netlify.app/?path=/story/components-form-segment--standard

@andy-polhill  I'm not sure it would work in this scenario, because `stackSm ` relates to the stacking of the label and content inside the cell, which is tied to the `[columnBreakpoint]="sm | md | lg"` input. After that breakpoint, the table is in columns view, so the labels don't appear in the cells at all. So `[stack]="sm | md | lg"` could be out of step with `columnBreakpoint` and be a bit confusing. Unless we wanted to add fine control to the cell layout within the range before `columnBreakpoint`?

projects/canopy/src/lib/table/table-cell/table-cell.component.html
```diff
@@ -11,7 +11,8 @@
   #content
   [ngClass]="{
     'lg-table-cell__content--align-end': align === alignOptions.End,
-    'lg-table-cell__content': showLabel
+    'lg-table-cell__content': true,
```
aaronBery, line 14: does this need to use ngClass if it is always evaluates to true i.e. can just use a standard class

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -14,7 +14,35 @@ The table component provides a way to present tabular data in a responsive forma
...
+      <td lg-table-cell>The Plague</td>
+      <td lg-table-cell>1947</td>
+    </tr>
+  </tbody>
+</table>
+~~~
+
+### Table with expandable detail rows
+
+This table adds the \`\`LgTableRowToggleComponent\`\`, and then an expandable detail row. This creates a table that allows the user to expand a row for more information. This can be seen on the <a href="/story/components-table--detail">Expandable Detail story</a>.
```
aaronBery, line 45: i would use a relative link rather than absolute just in case we ever hosted the app where the storybook wasn't at root

owensgit: Interestingly Storybook seems to make this URL relative anyway. It takes the URL fragment and adds it to `?path=`, resulting in `?path=/story/example`. If I try to make it relative, e.g. `story/example` or just `example`, it results in this following path: `?path=story/example` or `?path=example`, which breaks. So I think it has to be done like this?

projects/canopy/src/lib/table/table-cell/table-cell.component.html
owensgit: Thanks, that's done ✅

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -46,29 +76,44 @@ and in your HTML:
...
-        <lg-table-expanded-detail>
-          Detail content goes here
-        </lg-table-expanded-detail>
+  </thead>
+
+  <tbody lg-table-body>
+    <tr lg-table-row>
+      <td lg-table-cell [stackSm]="true">
+        <h3>Benefit one: Dependant person benefit</h3>
+        <p>You have chosen a dependant person to receive your income when you die. This person is Joe Bloggs. If this person has passed away, is no longer your spouse or civil partner or no longer financially dependant on you at the date of your death, then we will not pay out any income. Please refer to your policy document for more details.</p>
```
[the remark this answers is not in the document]
owensgit: Cool, I'm happy to use Lorum Ipsum. I remember when we had the original Canopy repo, we were avoiding anything L&G specific before we went open source, but I don't think we've come to a conclusion about whether to use real content or not since. I'll change this one and then open a discussion topic

projects/canopy/src/lib/table/table-cell/table-cell.component.ts
```diff
@@ -101,6 +101,12 @@ export class LgTableCellComponent {
     return this._showLabel;
   }
 
+  @Input() stackSm: boolean;
+
+  @HostBinding('class.lg-table-cell--stacked') get stackedClass() {
+    return this.stackSm;
+  }
+
```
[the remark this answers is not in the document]
owensgit: Ah yes, that's better 🙂 
Changed 👍

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -4,7 +4,9 @@ export const notes = `
 # Table Component
 
 ## Purpose
-The table component provides a way to present tabular data in a responsive format.
+The table component provides a way to present tabular data in a responsive format. On small devices, the layout of the table changes to display the table headers alongside each value, providing an easier way for users to read the data as they scroll down the list of rows.
```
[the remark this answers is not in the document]
owensgit: Done ✅

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -4,7 +4,9 @@ export const notes = `
...
+
+There are sevreal responsive options that allow you to cusomise the behavior of the table at various breakpoints, along with other options that allow you to cusomise the appearance.
```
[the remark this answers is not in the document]
owensgit: Yes I think this is a little more concise 👍  
Changed ✅

projects/canopy/src/lib/table/table.notes.ts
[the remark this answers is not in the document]
owensgit: Done ✅

projects/canopy/src/lib/table/table.notes.ts
```diff
@@ -88,6 +133,7 @@ A component that acts as a standard table cell.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
+| \`\`stackSm\`\` | On small devices, stack the label and the content on top of each other, instead of side by side | boolean | false | No |
```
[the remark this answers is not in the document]
owensgit: Done ✅

projects/canopy/src/lib/table/table-cell/table-cell.component.ts
```diff
@@ -101,6 +101,10 @@ export class LgTableCellComponent {
     return this._showLabel;
   }
 
+  @HostBinding('class.lg-table-cell--stacked')
+  @Input()
+  stackSm: true;
```
aaronBery, line 106: should this read `stackSm = true;`

owensgit: 🤦   Actually it needs to be this:

```
@Input()
stackSm = false;
```

Fixed now!

aaronBery (dismissed):

projects/canopy/src/lib/table/table-cell/table-cell.component.ts
aaronBery: ah yes that matches the notes :-)

owensgit: Nice spot, thanks Aaron 👍

owensgit:
> > On the segment component we have `[stack]="sm"`, I wonder if we can copy that convention?
> > https://legal-and-general-canopy.netlify.app/?path=/story/components-form-segment--standard
> 
> @andy-polhill I'm not sure it would work in this scenario, because `stackSm ` relates to the stacking of the label and content inside the cell, which is tied to the `[columnBreakpoint]="sm | md | lg"` input. After that breakpoint, the table is in columns view, so the labels don't appear in the cells at all. So `[stack]="sm | md | lg"` could be out of step with `columnBreakpoint` and be a bit confusing. Unless we wanted to add fine control to the cell layout within the range before `columnBreakpoint`? In that case we might want to consider making the default behaviour "stacked", and then switching to "side-by-side" before the `columnBreakpoint` - that would make it similar to the Segment component, and probably make more sense from a responsive point of view (small downside is it'd be a breaking change).

I'm thinking that adding `[stack]="sm | md | lg"` sounds nice because of the consistency of API to the Segment component, but the actual usefulness of it seems like YAGNI. I can't ever really see us needing to switching between stacked and side-by-side on small screens. I would say let's go with the simple `true|false` for now, and then think about a breaking change later if we need to add that fine grain control.

CC @aaronBery

andy-polhill:
> 
> However the actual usefulness of it seems like YAGNI. I can't ever really see us needing to switching between stacked and side-by-side on small screens - whereas for the Segment, there's a genuine use-case of it.
> 
> I would say let's go with the simple `true|false` for now, and then think about a breaking change later if we need to add that fine grain responsive control.
> 
> CC @aaronBery

That makes sense, so would it be `[stack]="true | false"`?

owensgit:
> That makes sense, so would it be `[stack]="true | false"`?

Thanks Andy, yep I've changed it to `[stack]="true | false"` now  👍

andy-polhill (approved):

aaronBery (approved):

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
```diff
@@ -5,6 +5,26 @@
   display: flex;
   align-items: center;
 
+  &--stacked {
```
elenagarrone, line 8: Because the input is now `stack` should the class also be `--stack`?

elenagarrone (approved):
Looking good, just one small comment but it's not a blocker.

projects/canopy/src/lib/table/table-cell/table-cell.component.scss
owensgit: Really sorry Elena, I didn't see this message before I hit merge. Given that we no longer officially support flat HTML, perhaps changing underlaying classes doesn't constitute as a breaking change anymore? If so - I could change this in a new PR. What do you think?

elenagarrone: I think that's totally fine. I don't think it's a breaking change :)

PR #440, opened 2021-06-29
feat(grid): add breakpoint-specific alignment and order classes

pauleustice:
# Description

Ports over various grid classes from the original http://flexboxgrid.com/ styles. Allows ordering and alignment at different viewport sizes.

## Requirements

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/grid.scss
+++ b/projects/canopy/src/styles/grid.scss
@@ -49,6 +49,53 @@ $columns: 12;
   max-width: 100%;
 }
 
+@mixin lg-grid-alignment($breakpoint) {
+  // Alignment and distribution
+
+  .lg-start-#{$breakpoint} {
+    justify-content: flex-start;
+    text-align: start;
+  }
+
+  .lg-center-#{$breakpoint} {
+    justify-content: center;
+    text-align: center;
+  }
+
+  .lg-end-#{$breakpoint} {
+    justify-content: flex-end;
+    text-align: end;
+  }
+
+  .lg-top-#{$breakpoint} {
+    align-items: flex-start;
+  }
+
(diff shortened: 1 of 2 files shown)


projects/canopy/src/styles/grid.scss
```diff
@@ -49,6 +49,53 @@ $columns: 12;
   max-width: 100%;
 }
 
+@mixin lg-grid-alignment($breakpoint) {
+  // Alignment and distribution
+
+  .start-#{$breakpoint} {
```
elenagarrone, line 55: Looks good but could you add the `lg-` prefix to the classes please?

projects/canopy/src/styles/grid.scss
```diff
@@ -49,6 +49,53 @@ $columns: 12;
   max-width: 100%;
 }
 
+@mixin lg-grid-alignment($breakpoint) {
```
elenagarrone, line 52: I would probably create directives for these too

projects/canopy/src/styles/grid.scss
pauleustice: Done

projects/canopy/src/styles/grid.scss
pauleustice: See https://github.com/Legal-and-General/canopy/issues/441

elenagarrone (approved):
Thank you!

andy-polhill (approved):
Thanks @pauleustice. I had a slight hesitancy around the extra size it might add, but I can see there is a need for it. It could be in the future we try and break up css, or even just have a build size allowance that runs in the pipeline so we have visibility over the size

PR #442, opened 2021-07-01
fix(links): update line height for the links

elenagarrone:
# Description

Fix an issue where the inline block links had different line height than the inline ones.
Also the inline block links on hover were causing the below content to shift.

Behaviour before:

https://user-images.githubusercontent.com/8397116/124100493-6e72ec00-da56-11eb-9f1a-cbdb6278f703.mov

Behaviour after:

https://user-images.githubusercontent.com/8397116/124100090-0fad7280-da56-11eb-9638-3598df791685.mov

Storybook link: https://deploy-preview-442--legal-and-general-canopy.netlify.app/?path=/story/links--utils

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -24,10 +24,11 @@
 ) {
   $box-shadow-inset-width: -0.063rem;
 
-  color: $default-color;
-  text-decoration: none;
   border-bottom: 0.125rem solid $default-color;
+  color: $default-color;
+  line-height: initial;
   padding: 0 0.125rem;
+  text-decoration: none;
 
   &:hover,
   &:hover:visited {
--- a/projects/canopy/src/styles/links.notes.ts
+++ b/projects/canopy/src/styles/links.notes.ts
@@ -1,0 +1,7 @@
+export const notes = `
+# Links
+
+The links styles are applied automatically when adding the global Canopy css.
+
+For custom components it's possible to override the style of a link by using the \`lg-unstyled-link\`.
+`;
--- a/projects/canopy/src/styles/links.stories.ts
+++ b/projects/canopy/src/styles/links.stories.ts
@@ -1,0 +1,33 @@
+import { withKnobs } from '@storybook/addon-knobs';
+import { moduleMetadata } from '@storybook/angular';
+
+import { notes } from './links.notes';
+import { LgAlertModule } from '../lib/alert';
+import { LgMarginModule } from '../lib/spacing';
+
+export default {
+  title: 'Links',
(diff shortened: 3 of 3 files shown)


paul-request (dismissed):

projects/canopy/src/styles/mixins.scss
```diff
@@ -24,10 +24,11 @@
 ) {
   $box-shadow-inset-width: -0.063rem;
 
-  color: $default-color;
-  text-decoration: none;
   border-bottom: 0.125rem solid $default-color;
+  color: $default-color;
+  line-height: var(--link-line-height);
```
paul-request, line 29: Or should it just inherit from the parent?

elenagarrone:
@andy-polhill, @owensgit and @paul-request this might be a breaking change as some people might have changed their code to fix this? Do you know if this was the case?

projects/canopy/src/styles/mixins.scss
elenagarrone: I tried that but it doesn't seem to work 🤔

elenagarrone: Here is using `inherit`: 
Screenshot 2021-07-01 at 10 27 50

paul-request: I was just wondering about a case where someone modifies the font-size of the parent with say, `lg-font-size-2` which contains a link, would that break the layout?

elenagarrone: BUT `initial` works so I'll change it to that 🤔

paul-request: Oh, I see. It is about where the "underline" sits... not sure initial is what we actually want. maybe the actual variable is the best solution

elenagarrone: Updated to `initial` now :)

paul-request (approved):

aaronBery:
Screenshot 2021-07-01 at 16 29 34

@elenagarrone  just to check...is link meant to sit under utils as the top level navigation item or was it intended this way?

elenagarrone:
> Screenshot 2021-07-01 at 16 29 34
> 
> @elenagarrone just to check...is link meant to sit under utils as the top level navigation item or was it intended this way?

Definitely not, how weird - I'll check what's going on 👍

elenagarrone:
@aaronBery thanks I have now fixed that 👍

aaronBery (approved):

andy-polhill (approved):

owensgit:
Just saw this, love the video clips in the description 🙂

Issue #315, opened 2021-04-13
Deprecations & upcoming breaking changes


owensgit:
ℹ️  &nbsp; Small breaking changes are more easily consumed when shipped together in a major version bump.

💡  &nbsp; When identifying such a change, such as removing a CSS var, add a deprecation notice and where possible provide deprecation warnings in the console.

💬   &nbsp; Add a comment to this issue to log current deprecations and upcoming minor breaking changes.

🚀  &nbsp; The pull request should be merged into the `next` branch, where it can be part of the next **pre-release**. At an appropriate time, `next` can be merged into `master` creating a major release of Canopy (with some nice release notes).

♻️  &nbsp; When released, this issue can be closed down and another one spun up for the next batch.

owensgit:
**Deprecated SCSS import in the following files**

- projects/canopy/src/lib/alert/alert.component.scss
- projects/canopy/src/lib/details/details.component.scss
- projects/canopy/src/lib/forms/validation/validation.component.scss

These files currently import `variants.scss` to make the global variant classes available in these components.

These imports need to be removed and a breaking change needs to be flagged so projects know that they need to import the new variants.scss file to get the variant classes, such as `.lg-variant--info`.

andy-polhill:
(temporarily removing the pin in favour of #332, max of 3 pins, will swap it back once the other `next` branch is merged)

andy-polhill:
I believe the next schedules breaking change will be the Angular 12 upgrade.

owensgit:
**Breaking change:** Removed solid-secondary button variant: https://github.com/Legal-and-General/canopy/pull/451

elenagarrone:
Deprecated: `LgSortCodeComponent`. I'll create a PR soon to remove the component in the next breaking change.

owensgit:
**Breaking change:** removal of the pipes module: https://github.com/Legal-and-General/canopy/pull/610/commits

PR #451, opened 2021-07-05
Removed solid-secondary button

### LINKED ISSUE #450: Remove `solid-secondary` button as it's deprecated
The `solid-secondary` button needs removing as the designers said we should only be using the blue one.

We could avoid a breaking change by setting it to be blue initially as the `solid-primary`.

Patil2099:
Fixes #450

### DIFF
--- a/projects/canopy/src/lib/card/card.stories.ts
+++ b/projects/canopy/src/lib/card/card.stories.ts
@@ -158,7 +158,7 @@ export const standard = () => ({
         </lg-card-title>
       </lg-card-header>
       <lg-card-content>
-        {{cardContent}}
+        {{cardContent}} <a href="#">Test link</a>.
       </lg-card-content>
     </lg-card>
   `,
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -29,7 +29,8 @@
   border-bottom: 0.125rem solid $default-color;
   padding: 0 0.125rem;
 
-  &:hover {
+  &:hover,
+  &:hover:visited {
     color: $hover-color;
     border-bottom: 0;
     box-shadow: inset 0 $box-shadow-inset-width 0 0 $hover-color,


elenagarrone:
Thank you very much @Patil2099 for picking this up 👏  it all looks good, I've only updated the base branch to be `next` since this is going to be a breaking change and we usually group them in `next` before releasing them all together.

Could you please update the commit message to follow [conventional commits](https://github.com/Legal-and-General/canopy/blob/master/docs/CONTRIBUTING.md#conventional-commits) standards?

For example in this case I would write something like:
```
feat(buttons): remove solid-secondary button

fix: #450
```

Thanks for contributing!

andy-polhill:
> Thank you very much @Patil2099 for picking this up 👏 it all looks good, I've only updated the base branch to be `next` since this is going to be a breaking change and we usually group them in `next` before releasing them all together.
> 
> Could you please update the commit message to follow [conventional commits](https://github.com/Legal-and-General/canopy/blob/master/docs/CONTRIBUTING.md#conventional-commits) standards?
> 
> For example in this case I would write something like:
> 
> ```
> feat(buttons): remove solid-secondary button
> 
> fix: #450
> ```
> 
> Thanks for contributing!

This is ace thanks @Patil2099. Just to add, you can use `npm run commit`, this will help to build the commit message for you by asking prompts in the command line, from there you can actually mention that it's a breaking change when it asks.

andy-polhill:
linking to #315 as we are putting this into the `next` branch

elenagarrone:
> For example in this case I would write something like:
> 
> ```
> feat(buttons): remove solid-secondary button
> 
> fix: #450
> ```

Apologies this should actually be:
```
feat(buttons): remove solid-secondary button

BREAKING CHANGE: remove the `solid-secondary` button. Use the `solid-primary` or any other button instead.

fix: #450
```

owensgit:
Hi @Patil2099 just wondering if you've had a chance to look at the comments above? Please feel free to reach out to one of us if you'd like to chat in more detail, we're happy to help.

owensgit:
@Patil2099  - Apologies, I tried to edit the commit message for you, but for some reason this closed the PR and doesn't show the edited commit. I've opened a new PR now because I can't seem to reopen this one: https://github.com/Legal-and-General/canopy/pull/456

Many thanks again for your contribution, it's greatly appreciated! Don't hesitate to contribute again if you'd like or get in touch if you need any support with using Canopy.

PR #454, opened 2021-07-08
refactor: remove duplicate code from unit tests

Zimann:
**Test file changes**

- card-footer.component.spec.ts
Removed the deprecated async() and replaced it with waitForAsync()

- carousel.component.spec.ts
Extracted the duplicated logic and placed it inside a function expression which gets called in multiple places

- feature-toggle.guard.spec.ts 
Created a separate enum to check for guard type when testing and the duplicated logic was placed inside a function expression which gets tweaked with different parameters that cater to the different test cases

**Style file changes**

-accordion-panel-heading.component.scss 
A simple class nesting change 

-button.component.scss 
Used the short-hand version of : top,left,right,bottom: 0 -> inset: 0 

**Directive file changes**

-feature-toggle.directive.ts 
Used the optional chaining operator and removed the check for the subscription

**Html file changes**

- modal-body-timer.component.html
'this' is not needed when referencing public component properties

**TS file changes**
- brand-icon.component.ts 
Removed the variable  _colour: string; as this was not used anywhere

## Requirements

Testing if all the unit tests pass like before with the command: npm run test 
Visually testing the components affected by the style changes (button and brand-icon components)

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/button/button.component.scss
+++ b/projects/canopy/src/lib/button/button.component.scss
@@ -40,10 +40,7 @@
 
   &__spinner {
     position: absolute;
-    left: 0;
-    right: 0;
-    top: 0;
-    bottom: 0;
+    inset: 0;
   }
 
   &--sm {
--- a/projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
+++ b/projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
@@ -60,7 +60,6 @@ export class LgFeatureToggleDirective implements OnInit, OnDestroy {
   }
 
   ngOnDestroy() {
-    /* istanbul ignore else */
     this.subscription?.unsubscribe();
   }
 }
--- a/projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
+++ b/projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
@@ -61,8 +61,6 @@ export class LgFeatureToggleDirective implements OnInit, OnDestroy {
 
   ngOnDestroy() {
     /* istanbul ignore else */
-    if (this.subscription) {
-      this.subscription.unsubscribe();
-    }
+    this.subscription?.unsubscribe();
   }
 }
--- a/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
+++ b/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
@@ -1,1 +1,1 @@
-{{ this.timer }}
+{{ timer }}
--- a/projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts
+++ b/projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts
@@ -109,10 +109,7 @@ export class LgAccordionItemComponent implements AfterContentInit, OnChanges, On
   }
 
   ngOnDestroy() {
-    if (this._toggleSubscription) {
-      this._toggleSubscription.unsubscribe();
-    }
-
+    this._toggleSubscription?.unsubscribe();
     this._removeSingleItemSelectionListener();
   }
 
--- a/projects/canopy/src/lib/brand-icon/brand-icon.component.ts
+++ b/projects/canopy/src/lib/brand-icon/brand-icon.component.ts
@@ -31,7 +31,6 @@ export class LgBrandIconComponent {
   @HostBinding('class.lg-brand-icon') class = true;
   @HostBinding('attr.aria-hidden') hidden = true;
 
-  _colour: string;
   @Input()
   set colour(colour: string) {
     const el = this.hostElement.nativeElement.querySelector(
--- a/projects/canopy/src/lib/card/card-footer/card-footer.component.spec.ts
+++ b/projects/canopy/src/lib/card/card-footer/card-footer.component.spec.ts
@@ -1,4 +1,4 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 
 import { LgCardFooterComponent } from './card-footer.component';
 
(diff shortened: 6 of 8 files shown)


projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
```diff
@@ -41,7 +41,9 @@
     pointer-events: none;
   }
 
-  &--active .lg-accordion__heading-icon {
-    transform: rotateX(180deg);
+  &--active {
```
elenagarrone, line 44: I think it's ok to have this on a single line instead of nested since the `&--active` is not used by anything else 🤔

projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
```diff
@@ -41,7 +41,9 @@
...
+    .lg-accordion__heading-icon {
```
andy-polhill, line 45: @Zimann this was the only one I didn't quite get, although I think it may be a lack of understanding on my side, are these the same thing?

projects/canopy/src/lib/button/button.component.scss
```diff
@@ -40,10 +40,7 @@
 
   &__spinner {
     position: absolute;
-    left: 0;
-    right: 0;
-    top: 0;
-    bottom: 0;
+    inset: 0;
```
elenagarrone, line 43: Nice, I didn't know this property existed x) 
The only thing is that it's not supported by IE11 which I don't think should be an issue since we are not supporting it in our apps anymore, @andy-polhill have we decided if IE11 support is dropped for Canopy too?

projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
Zimann: Yep, it's the same thing. It's just making use of the nested selectors from scss instead of having it chained (.classA .classB -> can be written as : .classA { .classB }

projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
Zimann: Reverted it to how it was.

projects/canopy/src/lib/button/button.component.scss
owensgit: We don't support in our team either, I would suggest we drop it for Canopy if we haven't already

projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
```diff
@@ -61,8 +61,6 @@ export class LgFeatureToggleDirective implements OnInit, OnDestroy {
 
   ngOnDestroy() {
     /* istanbul ignore else */
```
owensgit, line 63: Just wondering, do we need this line anymore?

owensgit (commented):
Great work, and congrats on your first Canopy contribution! Just a couple of minor comments/questions from me.

owensgit:
Also just wondering about commit messages. We tend to always include a scope, although I know that's optional in the conventional commit format. Do we wan to continue that trend here, or shall we keep it optional? One suggestion on the two commits here could be:

> refactor(unit-tests): remove duplicate code
> revert(accordion): nest accordion heading selector

projects/canopy/src/lib/accordion/accordion-panel-heading/accordion-panel-heading.component.scss
owensgit: I think this change belongs in the previous commit: `revert: selection on accordion-panel-heading.scss`

projects/canopy/src/lib/button/button.component.scss
owensgit, line 43: I think this SCSS change could be in it's own commit, as it's not part of the unit tests?

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
```diff
@@ -1 +1 @@
-{{ this.timer }}
+{{ timer }}
```
owensgit, line 1: Also wondering if this should be a separate commit, as it doesn't look related to the unit test changes.

projects/canopy/src/lib/button/button.component.scss
Zimann: Sure, I will tidy up the commits and make them more granular. Thanks! :)

projects/canopy/src/lib/feature-toggle/feature-toggle.directive.ts
Zimann: Removed this comment.

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
Zimann: Placed this in a separate commit. See: 7e17a3ab16d28a24086f4d1e770a7ac65c7e097b

andy-polhill:
Hey @Zimann thanks for updating and apologies for being so fussy about commit messages, but scope wise we normally just use the higher level component name and not the file itself...
```
refactor(card): remove deprecated async
refactor(feature-toggle): use optional chaining
refactor(modal): remove unneeded 'this' reference
```

elenagarrone (approved):
Thanks @Zimann 🎊

projects/canopy/src/lib/button/button.component.scss
andy-polhill: good question, I think there may be a couple of outliers, but I'm happy to start moving away from IE11 too. We should document or raise an issue somewhere about IE11 support.

andy-polhill (approved):

PR #455, opened 2021-07-08
build(sass): migrate from node-sass to dart sass

### LINKED ISSUE #452: Node-Sass is being deprecated, update to Dart Sass
From https://www.npmjs.com/package/node-sass

LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

This will result is a compile warning - DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

E.g.
```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($i, $columns)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
72 │   $width: percentage($i / $columns);
   │                      ^^^^^^^^^^^^^
   ╵
    node_modules/@legal-and-general/canopy/styles/grid.scss 72:22  @import
    resources/styles/styles_canopy.scss 7:9                        @import
    resources/styles/styles.scss 119:9                             root stylesheet
```

As part of update to Dart also update Sass to use `math.div`

Hopkinsa:
# Description
LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.
This will result is a compile warning - DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
Therefore, as part of update to Dart also update Sass to use math.div

During the migration,  additional warnings were also observed i.e.
```WARNING in ./projects/canopy/src/lib/spacing/margin/margin.directive.ts 75:54-68
"export 'SpacingVariant' was not found in '../spacing.interface'
 @ ./projects/canopy/src/lib/spacing/margin/margin.module.ts
 @ ./projects/canopy/src/lib/card/card.stories.ts
 @ ./projects/canopy sync \.stories\.ts$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true&quiet=true
```

From stackoverflow (https://stackoverflow.com/questions/40841641/cannot-import-exported-interface-export-not-found)
The root cause is in Typescript and transpilation. Once TypeScript code is transpiled, interfaces/types are gone. They don't exist anymore in the emitted files.

While the types are erased, the imports/exports aren't necessarily. The reason is that there are ambiguities and that the compiler doesn't always know for sure whether something that is exported is a type or a value.

To correct these errors, 'type' was added to some interface imports

Fixes #452


owensgit (dismissed):
Nice work @Hopkinsa! 👍

andy-polhill (commented):
This is great @Hopkinsa cheers!

My only minor comment is that I wondered if the`type` changes should be in a separate commit? I think that they are unrelated to the dart sass upgrade, but they are definitely valuable.

Having them in a separate commit would just give us the opportunity to document the reason in the commit message, and maybe link out to that SO pose. Apologies for the fussiness!

Hopkinsa:
Agreed - got caught up in the moment removing warnings

owensgit:
> Having them in a separate commit would just give us the opportunity to document the reason in the commit message, and maybe link out to that SO pose.

Great idea, I think having the SO link in the long description part of the message will be great 👍

andy-polhill (approved):
Nice one @Hopkinsa, thanks for breaking those commits up.

owensgit (approved):
Nice work @Hopkinsa!

PR #475, opened 2021-07-15
fix(button): corrects buttons not being styled in iOS Safari

### LINKED ISSUE #473: Form buttons do not render correctly in iOS 14
Buttons are not styled in Safari iOS 14.
before

Hopkinsa:
iOS Safari does not correctly style buttons, this may also affect other browsers on mobile devices.
This resets the appearence of the button, allowing it to be styled.

Fixes #473

The inclusion of both -moz-appearence and appearence is to ensure cross-browser compatability

Below is a screenshot taken on a physical iPhone, rather than a simulator
before

This is with the fix applied
after


projects/canopy/src/styles/mixins.scss
```diff
@@ -234,3 +234,11 @@ $breakpoints: (
   margin-bottom: var(--grid-gutter);
   padding: var(--component-padding);
 }
+
+@mixin lg-element-reset {
```
elenagarrone, line 238: This should be automatically done by `autoprefixer`. Does the button have the `appearance` property at all?

elenagarrone: Doesn't look like the button has that property. You can just add `appearance: none` in the style for `.lg-btn`. `autoprefixer` will handle the rest. If you are using scss only files in your project make sure you have `autoprefixer` setup.

elenagarrone (changes requested):
Please see comment :)

projects/canopy/src/styles/mixins.scss
Hopkinsa: Thanks, removed the vendor specific tags to let autoprefixer do it's job.

All form elements _should_ have an appearence property - from caniuse.com
> The appearance property defines how elements (particularly form controls) appear by default. By setting the value to none the default appearance can be entirely redefined using other CSS properties.

This also appears to be a known and ongoing issue with Safari on iOS.  The following was first asked 10 years ago, and still active in 2019.
[https://stackoverflow.com/questions/5449412/styling-input-buttons-for-ipad-and-iphone](url)

Hopkinsa:
> Please see comment :)

The style was updated and moved from the mixin file.
I hadn't seen the autoprefixer, so originally placed it there for reuse if needed elsewhere.

projects/canopy/src/styles/mixins.scss
elenagarrone: Yes we should probably have a look at those form elements and make sure they look alright in iOS 🤔

elenagarrone (approved):
Nice one thanks!

andy-polhill (approved):

PR #476, opened 2021-07-15
feat(modal): format timer input

Devaney:
# Description

Takes the input of the timer in seconds and converts it to a minutes/seconds format e.g. from 90 to "1:30"
image

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
+++ b/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
@@ -1,1 +1,1 @@
-{{ timer }}
+{{ formattedTime }}
--- a/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
+++ b/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
@@ -3,6 +3,7 @@ import {
   Component,
   HostBinding,
   Input,
+  OnInit,
   ViewEncapsulation,
 } from '@angular/core';
 
@@ -13,8 +14,16 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgModalBodyTimerComponent {
-  @Input() timer: string;
+export class LgModalBodyTimerComponent implements OnInit {
+  @Input() timer: number;
+  formattedTime: string | number;
 
   @HostBinding('class.lg-modal-body__timer') class = true;
+
+  ngOnInit() {
+    this.formattedTime = `${Math.floor(this.timer / 60)}:${(
+      '0' +
+      (this.timer % 60)
+    ).slice(-2)}`;
+  }
 }
--- a/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
+++ b/projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
@@ -3,7 +3,6 @@ import {
   Component,
   HostBinding,
   Input,
-  OnInit,
   ViewEncapsulation,
 } from '@angular/core';
 
@@ -14,16 +13,21 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgModalBodyTimerComponent implements OnInit {
-  @Input() timer: number;
-  formattedTime: string | number;
+export class LgModalBodyTimerComponent {
+  private _timer: number;
+  formattedTime: string;
 
   @HostBinding('class.lg-modal-body__timer') class = true;
 
-  ngOnInit() {
-    this.formattedTime = `${Math.floor(this.timer / 60)}:${(
+  @Input()
+  set timer(seconds: number | string) {
+    if (typeof seconds === 'string') {
+      seconds = parseInt(seconds, 10);
+    }
+    this._timer = seconds;
+    this.formattedTime = `${Math.floor(this._timer / 60)}:${(
       '0' +
-      (this.timer % 60)
+      (this._timer % 60)
     ).slice(-2)}`;
   }
 }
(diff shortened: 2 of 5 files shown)


projects/canopy/src/lib/modal/modal.stories.ts
```diff
@@ -28,7 +28,7 @@ export const standard = () => ({
       <lg-modal-header [headingLevel]="headingLevel">Lorem ipsum</lg-modal-header>
       <lg-modal-body>
         Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
-        <lg-modal-body-timer timer="0:25"></lg-modal-body-timer>
+        <lg-modal-body-timer timer="90"></lg-modal-body-timer>
```
elenagarrone, line 31: Could you maybe update the notes to explain the type of input required please?

Devaney: updated

elenagarrone (approved):

paul-request (approved):

paul-request:
I wonder if anyone else is using this? We aren't so it won't break for us, but could be a breaking change for others 🤔

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
```diff
@@ -13,8 +14,16 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgModalBodyTimerComponent {
-  @Input() timer: string;
+export class LgModalBodyTimerComponent implements OnInit {
+  @Input() timer: number;
+  formattedTime: string | number;
```
aaronBery, line 19: can the return type here be just a string? I'm trying to work out why you've got a union type here

aaronBery:
interestingly on the modal component the close button doesn't work. if it's an easy fix would be great to fix that (feel free to say no though and we can create a separate ticket)

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
```diff
@@ -1 +1 @@
-{{ timer }}
+{{ formattedTime }}
```
aaronBery, line 1: I'm wondering whether a pipe would have been a better solution here

elenagarrone: @Devaney initially made a pipe in our project but I was the one that suggested to maybe just use the code directly into the component. I couldn't think of any other scenario where we would want to use this format so I thought this was the best place. We could always make it into a pipe if we need it in the future?

elenagarrone:
> I wonder if anyone else is using this? We aren't so it won't break for us, but could be a breaking change for others 🤔

The modal is very recent and I'm pretty sure we are the only team that has used it to create the timeout modal.

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
```diff
@@ -13,8 +14,16 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgModalBodyTimerComponent {
-  @Input() timer: string;
+export class LgModalBodyTimerComponent implements OnInit {
+  @Input() timer: number;
```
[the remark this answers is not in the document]
Devaney: Option 1 seems better to me. This works.
```typescript
  ngOnInit() {
    if(typeof this.timer === 'string') {
      this.timer = parseInt(this.timer);
    }

    this.formattedTime = `${Math.floor(this.timer / 60)}:${(
      '0' +
      (this.timer % 60)
    ).slice(-2)}`;
  }
```
not sure how much back. and forth you usually have in canopy? but will push this unless I see some objections.

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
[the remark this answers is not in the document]
Devaney: No it cannot 👍🏻 updating

Devaney: It can and should be 👍🏻 updating

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
```diff
@@ -13,8 +14,16 @@ import {
...
+  formattedTime: string | number;
 
   @HostBinding('class.lg-modal-body__timer') class = true;
+
+  ngOnInit() {
```
aaronBery, line 23: can the timer value change before the component is destroyed? if it could then we could put the assignment in a setter on the timer input

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.html
Devaney: It was a pipe to start with (hence the branch name) and moved it into the timer component later. The reasoning being, its rather specific formatting and couldn't imagine it being used elsewhere. If we find somewhere else to use it we could move it to a pipe at that point. What do you think?

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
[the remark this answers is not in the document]
andy-polhill: that works, I wonder even if you need the `if` statement, you could just `parseInt` regardless, no big deal though.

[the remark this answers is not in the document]
Devaney: I tried and `parseInt` expects a string only. Without the `typeof` check it is smart enough to see it could get a number so complains.

projects/canopy/src/lib/modal/modal-body-timer/modal-body-timer.component.ts
Devaney: Have updated as suggested, seems simpler to me. What do you think?

aaronBery (dismissed):

elenagarrone (approved):

andy-polhill (approved):

aaronBery (approved):

PR #479, opened 2021-07-16
GitHub pages deploy

andy-polhill:
# Description

This PR will enable deployment of the master branch to GitHub pages. It works by executing the necessary commands to run the storybook static build and commit them to the gh-pages branch.

I decided to roll this manually as we have been wary of using actions from unknown sources, at least for the time being. We could fork an action and host it within our org but this would require quite a lot of extra work

There are some hopefully workable compromises.
1. The commit message in the gh-pages branch does not have the latest release version. This can be determined using Git but I was hoping to get this part working first.  
2. The default token doesn't have the ability to setup Github Pages, it can commit to the branch, but can't create the static site. This will have to be done manually.
3. There are no unit tests for the action. Given it only executes commands and there is no logic, all we could do is assert that the commands are run.
4. The node modules have to be manually installed before the action is run, before deciding this I tried the following.
  - Committing the node modules, just felt too wrong.
  - Generating a built version via a Husky commit hook and https://github.com/vercel/ncc, too complex

Once deployed we will have to start to redirect traffic to the new URL.

# Checklist:

### DIFF
--- a/.github/actions/gh-pages-storybook/action.yml
+++ b/.github/actions/gh-pages-storybook/action.yml
@@ -1,0 +1,6 @@
+name: '@legal-and-general/gh-pages-storybook'
+description: 'Deploy Storybook to GitHub Pages'
+runs:
+  using: 'node12'
+  main: 'index.js'
+
--- a/.github/actions/gh-pages-storybook/index.js
+++ b/.github/actions/gh-pages-storybook/index.js
@@ -1,0 +1,44 @@
+const core = require('@actions/core');
+const exec = require('@actions/exec');
+
+async function run () {
+  try {
+    if (core.isDebug()) {
+      core.startGroup('Debug: context');
+      core.debug(core.context);
+      core.endGroup();
+    }
+
+    const buildPath = 'storybook-static';
+    
+    core.info('[INFO]: commencing storybook gh-pages deploy')
+
+    core.info('[INFO]: running storybook build')
+    await exec.exec('npm', ['run', 'build-storybook']);
+
+    core.info('[INFO]: configuring git')
+    await exec.exec('git', ['config', '--global', 'user.email', 'github-actions[bot]@users.noreply.github.com']);
+    await exec.exec('git', ['config', '--global', 'user.name', 'github-actions[bot]']);
+
+    core.info('[INFO]: checking out gh-pages branch')
+    // -B: don't error if the branch already exists
+    await exec.exec('git', ['checkout', '-B', 'gh-pages']);
+
+    core.info('[INFO]: adding storybook static files')
+    // --force: override .gitignore
+    await exec.exec('git', ['add', '--force', buildPath]);
+
+    core.info('[INFO]: committing changes');
+    // TODO: retrieve latest release and put it in the commit msg, token does not have privileges
+    await exec.exec('git', ['commit', '-m', 'docs(gh-pages): latest storybook build']);
+
+    core.info('[INFO]: pushing to gh-pages');
+    await exec.exec('git', ['push', '-f', '--set-upstream', 'origin', 'gh-pages']);
+
+
+  } catch (error) {
+    core.setFailed(error.message);
+  }
+}
+
+(async () => { await run() })();
--- a/.github/actions/gh-pages-storybook/README.md
+++ b/.github/actions/gh-pages-storybook/README.md
@@ -1,0 +1,15 @@
+# GitHub Pages Storybook action
+
+This action will:
+- run the storybook static build
+- commit and push the resultant directory to the `gh-pages` branch
+
+The node_modules are not checked in and there is currently no separate build process for the action. For this reason we need to install the node_modules ahead of time.
+
+## Example usage
+```
+- name: install action dependencies
+  run: ci .github/actions/gh-pages-storybook && npm ci
+- name: deploy to gh-pages
+  uses: ./.github/actions/gh-pages-storybook
+```
--- a/.github/actions/gh-pages-storybook/package-lock.json
+++ b/.github/actions/gh-pages-storybook/package-lock.json
@@ -1,0 +1,58 @@
+{
+  "name": "@legal-and-general/gh-pages-storybook",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@legal-and-general/gh-pages-storybook",
+      "devDependencies": {
+        "@actions/core": "^1.4.0",
+        "@actions/exec": "^1.1.0"
+      }
+    },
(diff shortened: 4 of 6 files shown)


.github/actions/gh-pages-storybook/action.yml
```diff
@@ -0,0 +1,6 @@
+name: 'gh-pages-storybook'
```
elenagarrone, line 1: Do we want this action to be in the Canopy repo? I would have probably created a separate repo for it within `Legal-and-General`? you never know that someone else might want to use it?

.github/actions/gh-pages-storybook/index.js
```diff
@@ -0,0 +1,44 @@
...
+    core.info('[INFO]: checking out gh-pages branch')
+    // -B: don't error if the branch already exists
+    await exec.exec('git', ['checkout', '-B', 'gh-pages']);
+
+    core.info('[INFO]: adding storybook static files')
+    // --force: override .gitignore
+    await exec.exec('git', ['add', '--force', buildPath]);
+
+    core.info('[INFO]: committing changes');
+    // TODO: retrieve latest release and put it in the commit msg, token does not have privileges
```
elenagarrone, line 32: You mean the access token doesn't have privileges?

andy-polhill: as opposed to `does not`? not sure why I used long hand version.

.github/actions/gh-pages-storybook/action.yml
andy-polhill: I wanted to trial it here first for speed, it's relatively bespoke to our needs. While it's a local action we don't have to worry about versioning, package creation, linters, project setup etc. There are already some open source actions out there that do this, but I had some hesitancy while there isn't much security tooling around those actions.

I have modified the name though `name: '@legal-and-general/gh-pages-storybook'`

.github/actions/gh-pages-storybook/index.js
elenagarrone: no, sorry :D I meant to ask if you maybe only checked with the `GIHUB_TOKEN` instead of the `ACCESS_TOKEN` ;)

owensgit:
Looking good @andy-polhill!

> * The commit message in the gh-pages branch does not have the latest release version. This can be determined using Git but I was hoping to get this part working first.

Am I right by saying that these commits would only ever live on the `gh-pages` branch, so that we're able to push the static files up where they can be served by GitHub pages?

> * The default token doesn't have the ability to setup Github Pages, it can commit to the branch, but can't create the static site. This will have to be done manually.

This is just the initial setup and config in the Settings tab right? Or am I missing something?

> * There are no unit tests for the action. Given it only executes commands and there is no logic, all we could do is assert that the commands are run.

I suppose it coud be useful to test that the correct steps are ran int he correct order, and things like the build path ('storybook-static/') hasn't changed?

> * The node modules have to be manually installed before the action is run.

When you say manually, do you mean as part of the `install action dependencies` part in the workflow? Sorry I'm still fairly new to GH Actions.

Finally, two more questions :D 

1. I guess this doesn't cover branch deploys yet? Did you have anything in mind for that? I know you mentioned something about separate directory for branches at some point.
2. I think we might need to do something about the `next` branch, as currently we run `name: Push to master` on every push to `next`. So will we need to create a separate `name: Push to next` so we don't deploy the next version over the current master version?

andy-polhill:
> Looking good @andy-polhill!
> 
> > * The commit message in the gh-pages branch does not have the latest release version. This can be determined using Git but I was hoping to get this part working first.
> 
> Am I right by saying that these commits would only ever live on the `gh-pages` branch, so that we're able to push the static files up where they can be served by GitHub pages?

That is correct 👍
> 
> > * The default token doesn't have the ability to setup Github Pages, it can commit to the branch, but can't create the static site. This will have to be done manually.
> 
> This is just the initial setup and config in the Settings tab right? Or am I missing something?
> 
Yep thats all, I did try and do it programatically, but I think it would require giving the token more access privileges.

> > * There are no unit tests for the action. Given it only executes commands and there is no logic, all we could do is assert that the commands are run.
> 
> I suppose it could be useful to test that the correct steps are ran in the correct order, and things like the build path ('storybook-static/') hasn't changed? 🤔
> 

I could do it, but the tests would almost be a copy of the file itself, I'm not sure it would ever catch anything. I'm happy to do it if we think it would help?

> > * The node modules have to be manually installed before the action is run.
> 
> When you say manually, do you mean as part of the `install action dependencies` part in the workflow? Sorry I'm still fairly new to GH Actions.

Yep, the other option might be to create this as an external standalone action, you would then inject the dependencies as part of the build, but it would be quite a bit more work.

> 
> Finally, two more questions :D
> 
> 1. I guess this doesn't cover branch deploys yet? Did you have anything in mind for that? I know you mentioned something about separate directory for branches at some point.

Correct, not entirely sure how to tackle that yet. One way might be to deploy to sub directories of the same branch, but I'm not sure how feasible that is.

> 2. I think we might need to do something about the `next` branch, as currently we run `name: Push to master` on every push to `next`. So will we need to create a separate `name: Push to next` so we don't deploy the next version over the current master version?

Good point 👍  I'll put in a check

.github/actions/gh-pages-storybook/index.js
andy-polhill: yep, I tried with [@actions/github](https://www.npmjs.com/package/) which is already hydrated with the GITHUB_TOKEN. I'm not sure we have another token? we could create one, but I'd rather the pipeline not having too many privileges. What do you think? there may be a more fine grained option, I was running a bit low on time.

owensgit:
> > > * There are no unit tests for the action. Given it only executes commands and there is no logic, all we could do is assert that the commands are run.
> > 
> > 
> > I suppose it could be useful to test that the correct steps are ran in the correct order, and things like the build path ('storybook-static/') hasn't changed? 🤔
> 
> I could do it, but the tests would almost be a copy of the file itself, I'm not sure it would ever catch anything. I'm happy to do it if we think it would help?

You're right, I think it's a bit overkill to have tests for this. Happy not to include any 👍

andy-polhill:
@owensgit / @elenagarrone I decided to put the pages deploy in a separate flow in the end..
- there is no branch granularity for individual jobs in Actions, so it would get a bit hacky
- a separate workflow isolates the functionality, so we can treat breakages differently here from the main package creation.

elenagarrone (approved):
Looking good 👍

owensgit (approved):
Thanks for doing this Andy 👍

andy-polhill:
🤞 Here goes, first run should just do the git work but not actually turn gh-pages on.

Issue #483, opened 2021-07-19
Sort code fields are too wide


aaronBery:
Some work was recently done to fix spacing between sort code fields on https://github.com/Legal-and-General/canopy/pull/464 however one side effect is that it has widened the field.

Before (55px wide)

![Screenshot 2021-07-19 at 10 33 42](https://user-images.githubusercontent.com/7768665/126138573-42653c5d-bd21-4518-8c62-541c5d7edf8a.png)

After (58px wide)
![Screenshot 2021-07-19 at 10 29 29](https://user-images.githubusercontent.com/7768665/126138647-e677da97-65e7-44d6-b8e4-4ca864e8b358.png)

Tests performed on MS Edge on a Macbook Pro

aaronBery:
It may be worth investigating whether a similar issue now occurs on the date field which was also worked on as part of this work

elenagarrone:
Would we consider this an issue? one for @alexcanning I guess?

Hopkinsa:
Hi,  Just opened a PR for this.  Looking at the scss files the size of the sort code input fields were originally meant to be about 60px. However, through padding and other css it ended up being about 55px. The update from #464 forced the input to resize correctly, which has led to this discrepancy. 
The width for the individual sort code input boxes has now been set to 55px (3.4375rem), rather than the original 60px (3.75rem).

aaronBery:
> Would we consider this an issue? one for @alexcanning I guess?

I consulted Michelle who considered it an issue and one that we needed to resolve before upgrading our projects canopy version

elenagarrone:
Thanks for checking @aaronBery 👍  PR merged now 👍

PR #464, opened 2021-07-12
style(date-field): corrects date display on mobile devices

Hopkinsa:
# Description

The date fields do not have any spacing between them on mobile devices.  This change corrects the
issue

#462

## Requirements

Storybook link: 
https://deploy-preview-464--legal-and-general-canopy.netlify.app/?path=/story/components-form-date-input--standard

https://deploy-preview-464--legal-and-general-canopy.netlify.app/?path=/story/components-form-sort-code--standard

Screenshot:
Samsung S21 Before:
Screenshot 2021-07-12 at 09 42 18

After:
Samsung S21
S21

iPhone IOS 14
IOS14

Edge - 91
Edge-91

Firefox - 89
Firefox-89

Chrome - 91
Chrome-91

Opera - 77
Opera-77

Safari - 14
Safari-14

Sort Code
Safari - 14 Before
Screenshot 2021-07-12 at 10 43 13

Safari - 14 After
Screenshot 2021-07-12 at 10 43 19

PR #484, opened 2021-07-19
fix(sort-code): changes the size of the sort code fields

Hopkinsa:
The sort code fields are slightly larger than they should be.  This change addresses that.

#483

Please note that the original size of the sort code input fields was originally intended to be about 60px.  However, through padding and other css it ended up being about 55px.  The update from #464 forced the input to resize correctly, which has led to this discrepancy.  The width for the individual sort code input boxes has been set to 55px (3.4375rem), rather than the original 60px (3.75rem).  

Netlify link: https://deploy-preview-484--legal-and-general-canopy.netlify.app/?path=/story/components-form-sort-code--standard


elenagarrone (approved):
Thanks for adjusting this 👍

andy-polhill (approved):

PR #485, opened 2021-07-19
feat(primary-message): add new component

elenagarrone:
# Description

This PR adds the message with brand icon component. 

I kept the name that the designers suggested but i don't really like it 🤔  it's a bit long. I would have called it `state-message` or similar?

Design link: https://legalandgeneral.invisionapp.com/console/Canopy---Specs-cknlj6aj81j0u01yz3e5ssrmv/cknlj8vfz1kj801ymrx5afz5f/play

Desktop:
Screenshot 2021-07-19 at 18 27 04

Small device:
Screenshot 2021-07-19 at 18 27 17

# Checklist:


projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.scss
```diff
@@ -0,0 +1,16 @@
+@import '../../styles/mixins';
+
+.lg-message-with-brand-icon {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+
+  &__icon {
+    width: var(--brand-icon-xxs) !important;
```
elenagarrone, line 10: The `!important` here is needed to override the brand icon size.

dannyoz: 🤮

dannyoz (dismissed):

projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.scss
andy-polhill: at first I was going to ask... "could you not set it implicitly in the html rather than via the css?"

...then I saw the breakpoint...

which is interesting as I guess this is a scenario that might occur quite a lot with brand-icons, and is maybe something that the component itself could be extended towards, similar to the #362 proposal.

elenagarrone: I guess the brand icons could simply always be `xxs` on `sm` breakpoint but that would be a breaking change.
I'll check with James.

@andy-polhill do we think we could merge this bit of code in the meantime as it's blocking one of my project stories? In that case I'll make sure to create a new PR after discussing the options with the designers.

andy-polhill:
I have slightly mixed feelings about this component, initially I thought it could be built into the existing `card` or `alert` components as it seems to veer into similar territory, but after looking through I can see that it is quite different.

Maybe it's the name that is the jarring part, which is obviously very specific. Given we use content projection you could in theory create one of these without a brand icon, which makes me wonder if we consider a more generic name?

'lg-message'
'lg-hero-alert' 😬

From what I've seen it looks like a different type of `alert`. the `lg-alert` is used in addition to other content, where as this component is the content.

elenagarrone:
> I have slightly mixed feelings about this component, initially I thought it could be built into the existing `card` or `alert` components as it seems to veer into similar territory, but after looking through I can see that it is quite different.
> 
> Maybe it's the name that is the jarring part, which is obviously very specific. Given we use content projection you could in theory create one of these without a brand icon, which makes me wonder if we consider a more generic name?
> 
> 'lg-message'
> 'lg-hero-alert' 😬
> 
> From what I've seen it looks like a different type of `alert`. the `lg-alert` is used in addition to other content, where as this component is the content.

Maybe I misunderstood your sentence but you can't create one without the brand icon because it's in the template of the component 🤔 
It's not always an alert because it can be used as an "All done" or "confirmation" page and in that case you wouldn't want it to be an alert.

I'd say it's used to show a certain state of the application or journey so what about `lg-state-message`?

owensgit:
> I'd say it's used to show a certain state of the application or journey so what about `lg-state-message`?

We have a similar component which we use to describe certain states, such as "you have no products", or "you have no death benefits", which we'd replace with this one. So I'd be happy with `lg-state-message`.

projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { BrandIconName } from '../brand-icon';
+
+@Component({
+  selector: 'lg-message-with-brand-icon',
+  templateUrl: './message-with-brand-icon.component.html',
+  styleUrls: ['./message-with-brand-icon.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class LgMessageWithBrandIconComponent {
+  @Input() iconName: BrandIconName;
```
owensgit, line 12: In cases where we use icons, such as the buttons or quick action, the icon component itself is inserted via a named content projection slot. I wonder if we should do that here as well for consistency, rather than via an input. It also feels a bit more like the Canopy style, as we use content projection a lot. Happy to discuss though.

elenagarrone: Yes, that's a good point - the only issue is that by allowing the user to add the brand icon by themselves they have access to the `size` input which we want control over for this specific component.

andy-polhill:
> Maybe I misunderstood your sentence but you can't create one without the brand icon because it's in the template of the component 🤔

Nope you are right, that was my bad reading the code, it will currently always have a Brand Icon.

projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.ts
andy-polhill: Thats a good point, it could make it more consistent, it might also provide you with the extra css specificity you need to override the `!important`.

We have set a precedent where components can modify internal icon styles, this breaks the rule about components not being aware of other components, but feels like an ok exception.
e.g.
```css
.lg-accordion .lg-icon { /* this is ok */
  margin-right: var(--space-md);
}

.lg-accordion .lg-alert { /* this is not ok */
  margin-right: var(--space-md);
}
```

projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.scss
andy-polhill: I think thats fine as it could balloon otherwise, I might convert this thread to a new issue though.

andy-polhill:
regarding `lg-state-message`, I'm not too sure about `state` as it could mean a few different things. For me it feels more like 'main', 'primary', 'hero', or is the usage more specific 'result' e.g. 'lg-result-message'.

Is it always going to be the main message on any page? `lg-main-message`

Just wondering if it linked to the main element at least partly in it's usage.

```
The main content area consists of content that is directly related to or expands upon the central topic of a document, or the central functionality of an application.
```
This is tricky, if the consensus is for `state-message` I'm happy to go with that so we can move on.

projects/canopy/src/lib/message-with-brand-icon/message-with-brand-icon.component.ts
elenagarrone: Yes, agreed and if we tackle #488 in the near future the css code would go away anyway. I'll make that change 👍

owensgit:
> lg-main-message

I think this is better. In contexts where I've seen it, it's always the main message on a card, tab or page. And I like that it's less tied to the idea of it being about "state".

elenagarrone:
As discussed on Slack I have updated the name of the component to be `lg-primary-message`.

projects/canopy/src/lib/primary-message/primary-message-title/primary-message-title.component.html
```diff
@@ -0,0 +1 @@
+<ng-content></ng-content>
```
aaronBery, line 1: Given this is only used to display a title I wonder whether the content should be wrapped in an lg-heading to force the correct semantics

projects/canopy/src/lib/primary-message/primary-message.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, HostBinding, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+  selector: 'lg-primary-message',
+  templateUrl: './primary-message.component.html',
+  styleUrls: ['./primary-message.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
owensgit, line 7: Should this also have `changeDetection: ChangeDetectionStrategy.OnPush`? Not _exactly_ sure where we do and don't use that.

owensgit (approved):
Thanks for doing this @elenagarrone, I've approved but just with one final question.

projects/canopy/src/lib/primary-message/primary-message.component.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+          ),
+        ],
+        imports: [MockModule(LgBrandIconModule)],
+      }).compileComponents();
+    }),
+  );
+
+  beforeEach(() => {
+    fixture = MockRender(`
+      <lg-primary-message [iconName]="iconName" [hasRole]="hasRole">
```
aaronBery, line 34: are these inputs initially passed in as undefined?

aaronBery: is icon name an input?

projects/canopy/src/lib/primary-message/primary-message.component.ts
elenagarrone: yes, great point, i forgot it 👍

projects/canopy/src/lib/primary-message/primary-message.stories.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+};
+
+export const standard = () => ({
+  template: `
+    <lg-primary-message-story [hasRole]="hasRole"></lg-primary-message-story>
+  `,
+  props: {
+    hasRole: boolean('has role alert', true),
```
aaronBery, line 64: This is a hard one to describe... but something closer to 'Set role attribute to alert' is arguably clearer

projects/canopy/src/lib/primary-message/primary-message-title/primary-message-title.component.html
elenagarrone: I thought about this while working on it and I don't believe the content of the title is actually a heading - just styled like one.
To me the title is just a simple message (like a validation or error message); "Unfortunately something went wrong" or "There are no transactions to filter" and "All done" shouldn't be headings but just simple text that is read out when the message is triggered (that's why this component had a default role of alert).

In case of the component being used in a confirmation page, the title might be a heading and it's left to the user to decide that.

projects/canopy/src/lib/primary-message/primary-message.component.spec.ts
elenagarrone: Oh i need to remove `iconName`. 
`hasRole` is passed as `true` by default.

projects/canopy/src/lib/primary-message/primary-message.stories.ts
elenagarrone: 👍  yes, I'll update that thank you!

projects/canopy/src/lib/primary-message/primary-message.component.ts
elenagarrone: Updated :)

projects/canopy/src/lib/primary-message/primary-message.stories.ts
elenagarrone: Updated :)

projects/canopy/src/lib/primary-message/primary-message.component.spec.ts
elenagarrone: Removed both of the inputs from the `MockRender` :)

aaronBery (approved):

projects/canopy/src/lib/primary-message/primary-message-title/primary-message-title.component.html
aaronBery: ok that makes sense :-)

owensgit (approved):

Issue #184, opened 2021-01-29
Issue with date-input component validation


AlexandruHoisan:
The following is applicable for both sort-code component and date-input component:

When the sort-code component is empty or after the first interaction with one of it's inputs, clicking on submit we can see that the sort-code form-control status is valid even if that is not the case. The form-control is being set to its correct validation status right after. 
![Screenshot 2021-01-29 at 15 18 44](https://user-images.githubusercontent.com/68600084/106281746-4fbebe00-6248-11eb-9082-f06f634ddebf.png)
![Screenshot 2021-01-29 at 15 34 49](https://user-images.githubusercontent.com/68600084/106281748-50efeb00-6248-11eb-8637-83039ad49f6d.png)
![Screenshot 2021-01-29 at 15 13 19](https://user-images.githubusercontent.com/68600084/106281843-71b84080-6248-11eb-9579-d53f933d70a8.png)

This leads to successfully submitting a form even if the sort-code component is not valid.

This issue could potentially be caused by a time race between the validation calls of the individual input component and the sort-code component that wraps them up, or it might also be caused by the way Angular Reactive Forms handle the validation on complex nested forms( I am using complex here as in this scenario we have a formGroupDirective witch is also a AbstractControl).

To replicate, console.log() the form-control validity status when the `onSubmit` method is called
![Screenshot 2021-01-29 at 15 11 24](https://user-images.githubusercontent.com/68600084/106285895-b397b580-624d-11eb-8054-aa54afc9cc9f.png).

elenagarrone:
The sort code has been fixed by removing the component and creating a directive that gets applied to the input. The sort code is now a single field.

I guess this is still an issue with the date field.

PR #494, opened 2021-07-27
fix(sort-code): sync validation with the parent form

Zimann:
This PR has the following changes: 

- sort.code.component.ts 
  Subscribed to the `statusChanges` observable of the parent control. Inside the subscribe, we mark the form as touched 
  and individually mark the sort code fields as dirty, in order to trigger the `lg-input--error` and we also use `.setErrors` to trigger the validation message.

This fix addresses this bug, for the sort-code: https://github.com/Legal-and-General/canopy/issues/184

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
@@ -20,7 +20,7 @@ import {
 } from '@angular/forms';
 
 import { Subscription } from 'rxjs';
-import { filter } from 'rxjs/operators';
+import { filter, take } from 'rxjs/operators';
 
 import { LgErrorStateMatcher } from '../validation/error-state-matcher';
 import { LgDomService } from '../../utils/dom.service';
@@ -45,7 +45,6 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
   second: FormControl;
   third: FormControl;
   valueChanges: Subscription;
-  hasError: boolean;
 
   @Input() value: string;
   @Input() disabled: false;
@@ -89,18 +88,12 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
     if (this.ngControl != null) {
       this.ngControl.valueAccessor = this;
     }
-    this.first = new FormControl(null, [
+    this.first = new FormControl('', [Validators.required, Validators.pattern(/^\d\d$/)]);
+    this.second = new FormControl('', [
       Validators.required,
       Validators.pattern(/^\d\d$/),
     ]);
-    this.second = new FormControl(null, [
-      Validators.required,
-      Validators.pattern(/^\d\d$/),
-    ]);
-    this.third = new FormControl(null, [
-      Validators.required,
-      Validators.pattern(/^\d\d$/),
-    ]);
+    this.third = new FormControl('', [Validators.required, Validators.pattern(/^\d\d$/)]);
     this.sortCodeFormGroup = new FormGroup(
       {
         first: this.first,
@@ -119,6 +112,19 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
       this.ngControl.control.setValidators([this.validate.bind(this)]);
       this.ngControl.control.updateValueAndValidity();
     }
+
+    const sortCodeParentControl = this.parentFormGroupDirective.form.controls['sortCode'];
+
+    // show the error message when the input is invalid
+    sortCodeParentControl.statusChanges.pipe(take(1)).subscribe(() => {
+      this.sortCodeFormGroup.markAllAsTouched();
+      sortCodeParentControl.setErrors({
+        invalidField: true,
+      });
+      Object.keys(this.sortCodeFormGroup.controls).forEach((fieldName) => {
+        this.sortCodeFormGroup.controls[fieldName].markAsDirty();
+      });
+    });
 
     this.valueChanges = this.sortCodeFormGroup.valueChanges.subscribe(
       (sortCode: SortCodeField) => {
--- a/projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
+++ b/projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
@@ -113,7 +113,8 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
       this.ngControl.control.updateValueAndValidity();
     }
 
-    const sortCodeParentControl = this.parentFormGroupDirective.form.controls['sortCode'];
+    const sortCodeParentControl =
+      this.parentFormGroupDirective.form.controls[this.ngControl.name];
 
     // show the error message when the input is invalid
     sortCodeParentControl.statusChanges.pipe(take(1)).subscribe(() => {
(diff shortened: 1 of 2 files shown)


projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -114,6 +109,14 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
...
 
   ngOnInit(): void {
+    /* this is needed to sync validation when the parent form is first submitted
+       without having any values in the sort code inputs */
+    this.invalidSortCode$?.pipe(take(1))?.subscribe(() => {
+      this.sortCodeFormGroup.markAllAsTouched();
+      Object.keys(this.sortCodeFormGroup.controls).forEach((fieldName) => {
+        this.sortCodeFormGroup.controls[fieldName].markAsDirty();
+      });
+    });
```
owensgit, line 119: Just wondering if there's a way to cover this in the unit tests? 🤔

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -114,6 +109,14 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
   }
 
   ngOnInit(): void {
+    /* this is needed to sync validation when the parent form is first submitted
+       without having any values in the sort code inputs */
+    this.invalidSortCode$?.pipe(take(1))?.subscribe(() => {
```
elenagarrone, line 114: Is there maybe a way of setting the formGroup as required whenever the three fields are empty?

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -47,6 +47,7 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
   valueChanges: Subscription;
   hasError: boolean;
 
+  @Input() invalidSortCode$: Subject<void>;
```
owensgit, line 50: Just wondering, does this input get used in the template? Could you add an example to the Story notes? We also have a table in the story notes which describes each input, so it would be great to have another entry there for this one.

owensgit (commented):
Thanks for doing this @Zimann . Does this address https://github.com/Legal-and-General/canopy/issues/184? If so, it also applies to the date input, do it would be great to add the same fix there too.

owensgit:
It would be great to include more information in the long description about this fix too, perhaps even a slight expansion on what you wrote in the PR description 🙂

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
Zimann: Yes, this input gets used in the template as well :). I will take a look at adding an example in story notes as well as documenting this input. Thanks Owen 👍

owensgit: Great thanks @Zimann! Documentation is something we're really keen to improve on, so I think it's worth taking extra care with 🙂

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
Zimann: I will think of some tests to add 😎 👍

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
Zimann: It is set to required and it is marked as invalid, it's just that the sort code input fields are still seen as valid when the check from `IsControlInvalid` runs. I suspect it has something to do with the fact that the form has the option to `updateOn: 'blur'` so I suspect if no value is present there, it doesn't update properly as it's not being funnelled through the `blur` logic.  I also checked to see if we have the option to maybe put two values for the `updateOn` key but it only accepts one of these three 😢 `'blur', 'submit', 'onChange'` and in our case we need to have `blur` and `submit`.

projects/canopy/src/lib/forms/sort-code/sort-code.notes.ts
```diff
@@ -31,6 +31,7 @@ and in your HTML:
 ### LgSortCodeComponent
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
+| \`\`invalidSortCode$\`\` Subject that is passed from the parent form if the sort code fields are empty in order to mark the fields and set an error so that the validity state gets propagated | Subject<void> | No |
```
owensgit, line 34: Thanks for adding this. I'm still unsure about how to use this in an application, so perhaps we should have a section in the notes explaining how to use it within a form. 

E.g.
Screenshot 2021-07-28 at 09 49 26

elenagarrone: We've discussed the solution with @Zimann and we are trying to investigate a way not to have this input.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
```diff
@@ -120,6 +113,19 @@ export class LgSortCodeComponent implements OnInit, ControlValueAccessor {
       this.ngControl.control.updateValueAndValidity();
     }
 
+    const sortCodeParentControl = this.parentFormGroupDirective.form.controls['sortCode'];
```
elenagarrone, line 116: I guess we need to import the name of "sortCode" to the users. Maybe something to add to the notes?

Zimann: Very good point. Added a new line in the sort-code.notes.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
elenagarrone: @Zimann can you please write some unit tests?

Zimann: We could have a conversation about where to place these tests but for now, I'd say we let this one roll forward without them. Is that alright? 😄

owensgit: Unit tests tend to be super important for us, especially considering that Canopy is a package that's used in several applications. I would suggest that we give it a go now, and add some tests which mock the `statusChanges` observable to check that certain things have happened - e.g. that the controls are marked as dirty.

owensgit (commented):
Thanks for the updates @Zimann! Just a couple more things from me:

- Can you please squash these two commits into one and double check that the commit message is still relevant? It's really helpful for us to keep a clean commit history.
- In the squashed commit, could you also add some detailed information into the long description? This will help future devs, or even just us in the future, to understand why the change was made and what it's for. I think it's really worth the time to craft a great commit message and description for tricky bugs like this.

projects/canopy/src/lib/forms/sort-code/sort-code.component.ts
elenagarrone: I agree with @owensgit. We don't ship any code without unit tests ;) @Zimann I'm happy to help with the tests if needed.

elenagarrone:
Closing this and will open a new PR once I have all the changes ready.

PR #495, opened 2021-07-27
fix(date-field): corrects date display on IE11

bhawna123:
This explicitly set width to 100% which fix this issue

# Description

The date fields don't display properly on IE11. This change corrects the
issue

Fixes 
set width to 100% on actual input element. Parent element width doesn't reflect on IE11

Storybook link: 
https://deploy-preview-495--legal-and-general-canopy.netlify.app/?path=/story/components-form-date-input--standard

Screenshot: 
Before fix
Screenshot 2021-07-27 at 15 23 50

After Fix
Screenshot 2021-07-27 at 15 32 26

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/date/date-field.component.scss
+++ b/projects/canopy/src/lib/forms/date/date-field.component.scss
@@ -16,6 +16,10 @@
 
   & .lg-input-field .lg-input-field__inputs {
     width: 100%;
+
+    & .lg-input {
+      width: 100%;
+    }
   }
 }
 


projects/canopy/src/lib/forms/date/date-field.component.scss
```diff
@@ -16,6 +16,10 @@
 
   & .lg-input-field .lg-input-field__inputs {
     width: 100%;
+
+    & .lg-input {
```
elenagarrone, line 20: The only thing I'd say is that I don't think we are supporting IE11 anymore in Canopy so should this fix be here?
@owensgit @andy-polhill ?

bhawna123: @elenagarrone I spoke to the Product Team, got confirmation that we don't need to support IE11. I will go ahead and close PR.

bhawna123:
I spoke to the Product Team, got confirmation that we don't need to support IE11. I will go ahead and close PR.

PR #510, opened 2021-08-06
fix(page): accessibility audit fixes to the skip link

### LINKED ISSUE #508: Skip link missing id reference
"Skip link is not keyboard accessible due to the missing id reference to the main content. Without a skip link, it may be difficult for a screen reader or a keyboard user to locate certain points on the page. (AP071, AP072) [My Account Details, Personal Details, Change your Name/Title, Contact Details, Change address – Postcode Lookup, Change address – Manually, Change address – Confirmation & success page, Change address – Specific Home Insurance message, Log in and Security]"

paul-request:
# Description

The skip link does not link to an id via an href which has caused issues to be raised in the Axe tool and by our external accessibility testing partner. This PR adds in the missing href, but prevents the default event and continues to handle the skipping by using an event handler the way it previously did, avoiding a breaking change.

Fixes #508

# Checklist:

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx --no-install commitlint --edit $1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx lint-staged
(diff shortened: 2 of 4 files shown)


.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+if [ -z "$husky_skip_init" ]; then
```
owensgit, line 2: Did you commit this by mistake? I think if you remove this and then do an `npm i` it will go away

PR #517, opened 2021-08-09
build(deps): remove pretty-quick npm package

owensgit:
# Description

I'm proposing we remove this dependency, as we don't use it. Prettier is being used directly by our **lint-staged** setup, which follows the conventions in the lint-staged docs, so I'd be keen not to change that.

# Checklist:


elenagarrone (approved):

andy-polhill (approved):
Good call @owensgit 🎉

PR #537, opened 2021-08-31
feat(footer): include ID in footer link model

owensgit:
# Description

Updated the footer inputs for primary and secondary links, so that they can accept and render an ID
value

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/footer/footer.component.html
+++ b/projects/canopy/src/lib/footer/footer.component.html
@@ -9,6 +9,7 @@
       <li *ngFor="let link of primaryLinks" class="lg-footer__primary-nav-item">
         <a
           (click)="handlePrimaryLinkClick($event)"
+          [attr.id]="link.id"
           [attr.href]="link.href"
           [attr.target]="link.target || '_blank'"
           class="lg-footer__nav-link"
@@ -29,6 +30,7 @@
       <li *ngFor="let link of secondaryLinks" class="lg-footer__secondary-nav-item">
         <a
           (click)="handleSecondaryLinkClick($event)"
+          [attr.id]="link.id"
           [attr.href]="link.href"
           [attr.target]="link.target || '_blank'"
           class="lg-footer__nav-link"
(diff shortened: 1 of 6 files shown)


paul-request (dismissed):

projects/canopy/src/lib/footer/footer.component.ts
```diff
@@ -10,6 +10,7 @@ import {
 interface Link {
   text: string;
   href: string;
+  id?: string;
```
aaronBery, line 13: this might be a good opportunity to move this out into it's own interface file :-)

aaronBery: as in the whole `Link`

projects/canopy/src/lib/footer/footer.component.html
```diff
@@ -9,6 +9,7 @@ <h2 class="lg-visually-hidden" id="primary-links">Primary links</h2>
       <li *ngFor="let link of primaryLinks" class="lg-footer__primary-nav-item">
         <a
           (click)="handlePrimaryLinkClick($event)"
+          [attr.id]="link.id"
```
aaronBery, line 12: should the id be optional in the `Link` ? it seems like it's needed otherwise it'll set the id to undefined

owensgit: Good point, it should be optional in this case, so I'll remove the attribute in the template if it's not provided

projects/canopy/src/lib/footer/footer.component.ts
owensgit: Good idea, done ✅

projects/canopy/src/lib/footer/footer.component.html
owensgit: Just checked and if no ID is provided it won't add the attribute, so it won't be undefined 👍 

We decided to make it optional as the links don't need a ID value, it's really up to the consiming app. Also making it required would be a breaking change.

paul-request: yeah, if you set an attribute to undefined, it won't be added, whereas `null` will

paul-request (approved):

Nupur04 (approved):

PR #540, opened 2021-09-01
fix(page): sets tabindex on skip link to 0

paul-request:
The skip link needs a tab index of 0 so that it can be focused before all other focusable elements

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/page/page.component.html
+++ b/projects/canopy/src/lib/page/page.component.html
@@ -1,4 +1,9 @@
-<a (click)="skipToMain($event, target)" class="lg-page__skip-link" href="#main">
+<a
+  (click)="skipToMain($event, target)"
+  class="lg-page__skip-link"
+  href="#main"
+  tabindex="0"
+>
   Skip to main content
 </a>
 <ng-content select="[lg-header]"></ng-content>


HarshaDewani:
@paul-request I had a look on canopy and My accounts even if we add tabindex="0" it still doesn't work. Sorry, I think it could be my cache which is why I said that it simply works with tabindex="0"

I think we might need something like this.
```
.lg-page__skip-link {
  background: var(--color-white);
  @include lg-visually-hidden;

  &:focus,
  &:active {
    left: 0;
    clip: auto !important;
    width: auto !important;
    height: auto !important;
  }
} 
```

If you want I am happy to take a look at this as I can access the code now.

owensgit:
Thanks for your help with this Paul! It looks like Harsha's fix in #541 has done the trick, I just tested it locally on the page component story and it looks good. Do we still need this change as well?

paul-request:
> Thanks for your help with this Paul! It looks like Harsha's fix in #541 has done the trick, I just tested it locally on the page component story and it looks good. Do we still need this change as well?

Possibly not

paul-request:
Haha, didn't mean for there abrupt comment, I will close as Harsha has fixed her issue

PR #541, opened 2021-09-01
fix(page component): fix the display of skip link on focus

HarshaDewani:
# Description

Fix the display of the "Skip to main content" link on focus.
It seems like the width and height were not been overridden as the lg-visually-hidden mixin is using !important to apply width and height. This seems to be hiding the link to display while pressing tab.

Storybook link: https://deploy-preview-541--legal-and-general-canopy.netlify.app/

## Screenshot: 
Skip to main content on focus:

image

# Checklist:


owensgit (approved):
Nice one @HarshaDewani 🙌

paul-request (commented):
Is there a reason for the `important`?

HarshaDewani:
> Is there a reason for the `important`?

Yes, because we are using !important to set width and height in `lg-visually-hidden` which is been applied to this link.
So when the focus is set to the link those styles are not overridden hence we might need !important unless we have any other solution

paul-request (approved):

PR #543, opened 2021-09-02
Export link model and set z order for skip link

HarshaDewani:
# Description

This PR adds the z-order for `skip to main content` link also export the footer model.

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/variables/_main.scss
+++ b/projects/canopy/src/styles/variables/_main.scss
@@ -13,6 +13,7 @@
   --z-index-modal-overlay: 1000;
   --z-index-modal: 1010;
   --z-index-carousel-autoplay: 1;
+  --z-index-skip-link: 1020;
 
   /* Animation */
   --animation-duration: 0.25s;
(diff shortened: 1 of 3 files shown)


projects/canopy/src/styles/variables/_main.scss
```diff
@@ -13,6 +13,7 @@
   --z-index-modal-overlay: 1000;
   --z-index-modal: 1010;
   --z-index-carousel-autoplay: 1;
+  --z-index-skip-link: 9999;
```
paul-request, line 16: Do you think this should be a more considered value like `1020` and then if we need it to be greater in our app because we have higher z-indexes in our custom nav, then we could override this to suit our needs

paul-request (changes requested):

projects/canopy/src/styles/variables/_main.scss
HarshaDewani: Updated

paul-request (approved):

aaronBery (approved):

bhawna123 (approved):

PR #551, opened 2021-09-08
fix(footer-component): removes extra space from underline on link hover

HarshaDewani:
# Description
Removes extra space from underline on link hover

Screenshot: 
Before the fix, we were able to see some spaces when we hover on the link
image

After this change it looks like this:

image

# Checklist:


andy-polhill (approved):
Thats interesting, so it was adding some leading whitespace?

HarshaDewani:
> Thats interesting, so it was adding some leading whitespace?

Yes, that right.

AlexandruHoisan (approved):

aaronBery (approved):

Issue #439, opened 2021-06-29
Remove static HTML examples from Story notes


owensgit:
We have taken the decision not to support static HTML for non-Angular projects.

Our story notes still contain a lot of examples of static HTML. 

These should be removed.

Issue #332, opened 2021-04-26
Deprecation of SCSS only documentation


andy-polhill:
We are tentatively deciding to deprecate documentation for those applications only using the SCSS/CSS. This refers to the manual documentation that can be found in the storybook notes.

If any projects are using SCSS/CSS only, things will still work as expected, but in the future you will have to inspect the code to determine what classes to provide.

- The manual documentation was providing quite a time overhead in terms of contributing new components.
- The manual documentation was difficult to keep up to date
- The manual documentation could draw a team into using only the scss/css where it would have been preferable to use the Angular components.

Initially the intention of this documentation was to help with teams not using Angular, it is however quite cumbersome and received little interest for the amount of effort put into it. If you are a team who use this approach, or would be interested in using Canopy not with Angular we do have some advanced experiments with Angular Elements. Please respond on this issue.

In time we will start to remove the `scss only` documentation, and will keep this issue up to date.

andy-polhill:
This will most likely be a part of #35

PR #566, opened 2021-09-24
docs(storybook notes): removes static html notes

HarshaDewani:
This PR removes the instructions from storybook notes that are needed to add static HTML when only
SCSS files are used

#439

# Description

This was been discussed a few months back that as none of the teams are planning to use just the SCSS files and also it's getting difficult to maintain the SCSS instructions for complex components like tabs etc, it makes sense to remove the relevant instructions from storybook.

Further information can be found here - https://github.com/Legal-and-General/canopy/issues/332

Fixes # 439

# Checklist:


elenagarrone (approved):
Nice one! thanks for picking this up @HarshaDewani 🎊

paul-request (approved):

PR #573, opened 2021-10-01
feat(pipes): adds separate module for camel-case & kebab-case

### LINKED ISSUE #393: Create separate modules for the pipes
The pipes currently live in a pipe module. We should follow the same pattern we have for the other modules and create a module for each pipe separately.

HarshaDewani:
This PR deprecates pipes module also adds separate module for camel-case & kebab-case

# Description
This PR allows importing the camel-case and kebab-case individually rather than importing the whole pipe module.

Fixes #393 

# Checklist:


projects/canopy/src/lib/pipes/camel-case/camel-case.module.ts
```diff
@@ -0,0 +1,9 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+@NgModule({
+  imports: [CommonModule],
+  declarations: [LgCamelCasePipe],
+  exports: [LgCamelCasePipe],
+})
+export class LgCamelCasePipe {}
```
owensgit, line 9: I think this export class needs to be `LgCamelCasePipeModule`

projects/canopy/src/lib/pipes/kebab-case/kebab-case.module.ts
```diff
@@ -0,0 +1,9 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+@NgModule({
+  imports: [CommonModule],
+  declarations: [LgKebabCasePipe],
+  exports: [LgKebabCasePipe],
+})
+export class LgKebabCasePipe {}
```
owensgit, line 9: I think this export class needs to be `LgKebabCasePipeModule`

projects/canopy/src/lib/pipes/pipes.module.ts
```diff
@@ -11,4 +11,14 @@ const pipes = [LgCamelCasePipe, LgKebabCasePipe];
...
 })
-export class LgPipesModule {}
+/**
+ * @deprecated Pipe module is deprecated.
+ * Please import camel-case or kebab-case module instead.
+ */
+export class LgPipesModule {
+  constructor() {
+    console.warn(
+      'Pipe module is deprecated. Please import camel-case or kebab-case modules instead',
```
owensgit, line 21: I think it would be great to include the names of the module classes, like this:

LgPipesModule module is deprecated. Please import LgCamelCasePipeModule or LgKebabCasePipeModule instead.

projects/canopy/src/lib/pipes/pipes.module.ts
```diff
@@ -11,4 +11,14 @@ const pipes = [LgCamelCasePipe, LgKebabCasePipe];
   declarations: [...pipes],
   exports: [...pipes],
 })
-export class LgPipesModule {}
+/**
+ * @deprecated Pipe module is deprecated.
+ * Please import camel-case or kebab-case module instead.
```
owensgit, line 16: I think it would also be good to include the real module names here too.

owensgit (changes requested):
Thanks for picking this up @HarshaDewani! 🙌  Just a few requests from me.

projects/canopy/src/lib/pipes/pipes.module.ts
HarshaDewani: Absolutely

aaronBery:
If we are deprecating the pipes module it seems this needs to be marked as a breaking change

owensgit:
> If we are deprecating the pipes module it seems this needs to be marked as a breaking change

I don't think we need to mark this as a breaking change right now. My understanding is that we mark it deprecated now, but when we remove it, then it becomes a breaking change and a major version bump. Happy to discuss though.

PR #607, opened 2021-11-03
feat(brand-icon): upgrade to CSF format

owensgit:
# Description

Upgrades the Brand Icon to CSF format. Makes the brand icon story work in Storybook v6.

### DIFF
--- a/projects/canopy/src/lib/brand-icon/brand-icon.notes.ts
+++ b/projects/canopy/src/lib/brand-icon/brand-icon.notes.ts
@@ -1,7 +1,4 @@
 export const notes = `
-# Brand Icon Component
-
-## Purpose
 Provides a way of adding common brand icons.
 
 ## Usage
@@ -14,11 +11,15 @@ Import the component in your application:
 })
 ~~~
 
-Import the \`LgBrandIconRegistry\` and register your brand icons:
+Import the \`LgBrandIconRegistry\` service and register your brand icons inside your module:
 
 ~~~js
+// import the desired icon
+import { lgBrandIconSun } from '@legal-and-general/canopy';
+
 export class AppModule {
   constructor(private brandIconRegistry: LgBrandIconRegistry) {
+    // register the icon using the \`brandIconRegistry\` service
     this.brandIconRegistry.registerBrandIcon([
       lgBrandIconSun
     ]);
@@ -26,24 +27,21 @@ export class AppModule {
 }
 ~~~
 
-and in your HTML:
+Your HTML:
 
 ~~~html
+<!-- with default size and colour -->
 <lg-brand-icon name="sun"></lg-brand-icon>
+
+<!-- with a size set -->
+<lg-brand-icon name="sun" size="sm"></lg-brand-icon>
+
+<!-- with a colour set -->
+<lg-brand-icon name="sun" colour="--color-lily-green"></lg-brand-icon>
 ~~~
 
-## Inputs
+## Branding / Colours
+The yellow fill colour of the brand icons can be changed globally  by overriding the \`--brand-icon-fill-primary\` css variable. Note that changing this variable will update the fill colour of all the icons.
 
-| Name | Description | Type | Default | Required |
-|------|-------------|:----:|:-----:|:-----:|
-| \`\`name\`\` | the name of the icon | string | undefined | yes |
-| \`\`size\`\` | the size of the icon | BrandIconSize | 'sm' | no |
-| \`\`colour\`\` | the specific colour of the icon (for global colours see the "Branding" section below) | css variable as a string | undefined | no |
-
-## Branding
-The yellow fill colour of the brand icons can be changed by overriding the \`--brand-icon-fill-primary\` css variable.
-
-Note that changing that variable will update the fill colour of all the icons.
-
-To change the colour of a specific icon use the \`colour\` input (see details in the table above).
+To change the colour of a specific icon use the \`colour\` input on the component.
 `;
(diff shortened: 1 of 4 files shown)


projects/canopy/src/lib/brand-icon/brand-icon.notes.ts
```diff
@@ -14,36 +11,37 @@ Import the component in your application:
 })
 ~~~
 
-Import the \`LgBrandIconRegistry\` and register your brand icons:
+Import the \`LgBrandIconRegistry\` service and register your brand icons inside your module:
 
 ~~~js
+// import the descired icon
```
elenagarrone, line 17: Typo here

paul-request (approved):

projects/canopy/src/lib/brand-icon/brand-icon.notes.ts
owensgit: Whoops, thanks for spotting, fixed now

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
```diff
@@ -96,22 +95,6 @@ class SwatchBrandIconComponent implements OnChanges {
 
 const sizes = ['xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl'];
```
elenagarrone, line 96: Should we have this as an `enum` somewhere?

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
```diff
@@ -120,21 +103,81 @@ const colours = [
...
+          summary: '--color-dandelion-yellow',
+        },
+      },
+      control: {
+        type: 'select',
+      },
+    },
+    specificColour: {
+      options: colours,
+      description: 'The colour of a specific icon, using the `colour` input',
```
elenagarrone, line 151: I think this is a bit confusing. This in my opinion should be called `colour` since that's the name of the input.

While the above should be call `globalColour`.

elenagarrone:
@owensgit something doesn't look right in the template code 🤔  it gets displayed like:
Screenshot 2021-11-03 at 11 42 38

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
```diff
@@ -120,21 +103,81 @@ const colours = [
...
+    },
   },
+};
+
+const template = `
+  <lg-swatch-brand-icon [size]="size" [colour]="colour" [specificColour]="specificColour"></lg-swatch-brand-icon>
+`;
+const brandIconsTemplate: Story<LgBrandIconComponent> = (args: LgBrandIconComponent) => ({
+  props: args,
+  template: template,
```
elenagarrone, line 172: I would personally move the template straight into here since it's only one line, no?

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
owensgit: I'm not sure, as it's just used to create the options for controls on the story, and not by the component itself. The `size` input on the component uses this type: 

```
export type BrandIconSize = 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl';`
```

What do you think? I'd say maybe leave it as is for now as it's how was set before the upgrade and I don't think it would be currently useful outside this story. Happy to discuss though 🙂

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
owensgit: Ah yes I wondered that too, the original story I think had it like this as well, but I've changed it now to how you suggested. I've also improved the description a little.

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
owensgit: > @owensgit something doesn't look right in the template code 🤔 it gets displayed like: Screenshot 2021-11-03 at 11 42 38

Ah yes, I'll take a look into that, thanks :)

owensgit:
> @owensgit something doesn't look right in the template code 🤔 it gets displayed like: Screenshot 2021-11-03 at 11 42 38

Ah yes, I'll take a look into that, thanks :)

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
elenagarrone: I think we said we were going to change those types to enums so that we can also use them better in the stories. @paul-request , what do you think?

elenagarrone:
> > @owensgit something doesn't look right in the template code 🤔 it gets displayed like: Screenshot 2021-11-03 at 11 42 38
> 
> Ah yes, I'll take a look into that, thanks :)

This still looks weird 🤔  but maybe better as the code would appear as the `swatch` component. 
I would refactor the bit of code on line 172 from `template: template` to just `template` ;)

elenagarrone:
> > > @owensgit something doesn't look right in the template code 🤔 it gets displayed like: Screenshot 2021-11-03 at 11 42 38
> > 
> > 
> > Ah yes, I'll take a look into that, thanks :)
> 
> This still looks weird 🤔 but maybe better as the code would appear as the `swatch` component. I would refactor the bit of code on line 172 from `template: template` to just `template` ;)

As discussed this was my fault! looks good 👍

projects/canopy/src/lib/brand-icon/brand-icons.stories.ts
paul-request: Yes, in order to reuse, rather than redefining in the story I suggested converting types to enums so that we can just export the enum from the component/inferface, and use it in CSF definition:
```
table: {
  type: {
    summary: Object.values(BrandIconSize),
  },
}

It seems like this type is already exported in `brand-icon.component.ts`, so we should convert to an enum and use as above IMO.

elenagarrone (approved):

PR #623, opened 2021-11-19
fix(carousel): pause on action and fix carousel item break on tab switch

### LINKED ISSUE #622: Fix the advertising Carousel scroll behaviour when CTA is interacted with or has focus
If a CTA on one of the advertising tiles has focus, and the animation / scrolling is on, it breaks.

This will be fixed in Canopy.

### Tabbing between the controls on the ads
Legal   General - Online Self Service

### Clicking to an overlay and then coming back

I assume the issue here is that the ad has scrolled to the next ad when focus is returned (overlays / modals must return focus to the button that opened them when the overlay is closed [for accessibility]).
from-overlay

For the first issue, perhaps the solution is to disable the animation if an ad has focus? I'm not sure what we can do about the second issue as focus will have gone to the overlay.

Related issue in OSS: [DCI-17572](https://landg.spectrumhosting.net/jira/browse/DCI-17572)

AlenNS:
### Stop carousel when any button inside carousel item has been clicked and fix broken view on tab switch

# Description

If there are any buttons in the carousel item, we assume there is an action such as routing to another page or opening a modal interface. In this case we want to prevent carousel running and we should trigger an event that will pause the carousel on the current item. Once we switch back to the page with carousel we can click on play button and run the carousel again.

If we try to switch between the carousel items with tab key, we want our carousel item to be fully visible. Currently that view is broken, since there are previous and next carousel items broken in half. With this change it should switch though the items pressing tab key on keyboard without any issues.

Fixes #622

## Requirements

Storybook link: https://legal-and-general.github.io/canopy/?path=/story/components-carousel--auto-play-enabled
Screenshot:

1. Pause on CTA
https://user-images.githubusercontent.com/19712379/142638021-fe7e1e7d-f369-42ac-b020-d7d2cc169d40.mov

2. Switch items on tab
https://user-images.githubusercontent.com/19712379/142638100-7d9fae7e-e340-44aa-89fd-67c95a3c8e33.mov

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
+++ b/projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
@@ -2,7 +2,10 @@ import {
   ChangeDetectionStrategy,
   Component,
   ElementRef,
+  EventEmitter,
   HostBinding,
+  HostListener,
+  Output,
   ViewEncapsulation,
 } from '@angular/core';
 
@@ -17,6 +20,15 @@ export class LgCarouselItemComponent {
   @HostBinding('class.lg-carousel-item') class = true;
   @HostBinding('attr.role') role = 'tabpanel';
   @HostBinding('attr.aria-selected') public selected = false;
+
+  @Output() pauseEmit = new EventEmitter<string>();
+
+  @HostListener('click', ['$event.target.type'])
+  onClick(btn: string): void {
+    if (btn === 'button') {
+      this.pauseEmit.emit(btn);
+    }
+  }
 
   constructor(public element: ElementRef) {}
 
--- a/projects/canopy/src/lib/carousel/carousel.component.ts
+++ b/projects/canopy/src/lib/carousel/carousel.component.ts
@@ -4,6 +4,7 @@ import {
   ChangeDetectorRef,
   Component,
   ContentChildren,
+  HostListener,
   Input,
   OnDestroy,
   QueryList,
@@ -24,6 +25,8 @@ import { LgCarouselItemComponent } from './carousel-item/carousel-item.component
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class LgCarouselComponent implements AfterContentInit, OnDestroy {
+
+  constructor(private cd: ChangeDetectorRef) {}
   @Input() description: string;
   @Input() headingLevel: HeadingLevel;
   @Input() loopMode = false;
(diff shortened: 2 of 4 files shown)


projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -41,13 +44,32 @@ export class LgCarouselComponent implements AfterContentInit, OnDestroy {
   pause = new BehaviorSubject<boolean>(false);
   pausableTimer$: Observable<void>;
 
+  @HostListener('document:keydown.tab', ['$event'])
+  onKeydownHandler(event: KeyboardEvent) {
+    const target = event.target as HTMLElement;
+    const parent = target.parentElement as HTMLElement;
+    const el = parent.className;
+
+    if (el === 'lg-carousel-item') {
```
markblandford, line 53: I'm not sure about this. Doesn't this programmatically change the tab order of the document? Furthermore, if there are two elements that can receive focus on the `lg-carousel-item`, for example two buttons, doesn't this prevent the second element from ever being able to be focused on? Effectively making it only accessible to customers with a mouse?

markblandford: (thinking out loud) could an alternative be an extension of what you've done with the pausing when a button is clicked? I.e. whilst an element within the `lg-carousel-item` has focus, the carousel is paused (and restarts when focus is lost [if it was auto-paused])?

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -18,6 +21,15 @@ export class LgCarouselItemComponent {
   @HostBinding('attr.role') role = 'tabpanel';
   @HostBinding('attr.aria-selected') public selected = false;
 
+  @Output() pauseEmit = new EventEmitter<string>();
+
+  @HostListener('click', ['$event.target.type'])
```
markblandford, line 26: Will this trigger when a button is 'clicked' with a keyboard?

elenagarrone: Good point, `click` usually covers the `enter` key but I don't think it covers the space 🤔

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -41,13 +44,32 @@ export class LgCarouselComponent implements AfterContentInit, OnDestroy {
...
+
   selectCarouselItem(index: number): void {
     this.selectedItemIndex = index;
     this.selectedItem = this.carouselItems.get(index);
     this.selectedItemContent = this.selectedItem.itemContent;
 
     this.carouselItems.forEach((carouselItem: LgCarouselItemComponent) => {
       carouselItem.selected = false;
+      carouselItem.pauseEmit.subscribe((el: string) => {
+        if (el === 'button') {
```
elenagarrone, line 68: Wouldn't you want the same behaviour for a link?

projects/canopy/src/lib/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -18,6 +21,15 @@ export class LgCarouselItemComponent {
...
+  onClick(btn: string): void {
```
owensgit, line 27: I'd consider renaming this arg to `type`, as it's more descriptive

elenagarrone:
@AlenNS any update on this one?

AlenNS:
> @AlenNS any update on this one?

@elenagarrone not yet, priority has been changed so I had to work on other tasks.
Anyway, @markblandford suggested trying to control it from the OSS instead of making the change in the Canopy.

elenagarrone:
@AlenNS @markblandford should we close this PR? We can always open it back if needed?

markblandford:
This PR should be closed as it has been superseded by other changes:
* #685
* #689

PR #626, opened 2021-11-25
build(node): update to Node v15.14.0 and NPM v7.7.6

owensgit:
This bumps the Node version to 15.14.0 and also uses the corresponding NPM v7.7.6 to install packages. As a result of the NPM bump, the package-lock.json file has been updated to `lockfileVersion: 2`.

`lockfileVersion: 2` is a different lock file format, but should be backwards compatible with older versions of NPM. However, it's still recommended to use NPM v7.7.6 with this version of Node, so I added an NPM version to the package.json file.

# Checklist:


elenagarrone (approved):

aaronBery (changes requested):
not sure node v15 is lts so probably better to go for one that has the support
https://nodejs.org/en/about/releases/

owensgit:
> not sure node v15 is lts so probably better to go for one that has the support https://nodejs.org/en/about/releases/

Oh I didn't know that, so really we should be looking to go to version 16 next. 

It's tricky because we'd need to go to Angular 12 first, as the current version of Angular doesn't support v8 of NPM 🤔 

Maybe I should close this down until we're able to go to Node v16?

aaronBery:
> > not sure node v15 is lts so probably better to go for one that has the support https://nodejs.org/en/about/releases/
> 
> Oh I didn't know that, so really we should be looking to go to version 16 next.
> 
> It's tricky because we'd need to go to Angular 12 first, as the current version of Angular doesn't support v8 of NPM 🤔
> 
> Maybe I should close this down until we're able to go to Node v16?

Oh that's a shame. I agree, let's (plan to) update Angular and then upgrade node versions after :-)

owensgit:
Closing this as Node v15 doesn't have long term support and we're unable to go to v16 until we're on Angular 12.

PR #645, opened 2021-12-15
Rollback version of angular devkit causing the storybook to not load

elenagarrone:
# Description

Version `0.1102.15` of `@angular-devkit/build-angular` was causing an issue where storybook wouldn't start with the following error:
```
Module build failed (from ./node_modules/postcss-loader/src/index.js):
Error: PostCSS plugin autoprefixer requires PostCSS 8.
```
even though in the `package.json` we specifically install `postcss` at version 8.

After some googling I found this stackoverflow: https://stackoverflow.com/questions/64925926/error-postcss-plugin-tailwindcss-requires-postcss-8

and realised that `tailwindcss` is actually used by the angular devkit. Rolling back the version seem to fix the issue.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "zone.js": "^0.10.3"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^0.1102.15",
+    "@angular-devkit/build-angular": "^0.1002.0",
     "@angular/cli": "^11.2.15",
     "@angular/compiler-cli": "^11.1.0",
     "@angular/language-service": "^11.2.14",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,19 +5,19 @@
   "requires": true,
   "dependencies": {
     "@angular-devkit/architect": {
-      "version": "0.1102.15",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "zone.js": "^0.10.3"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^0.1102.15",
+    "@angular-devkit/build-angular": "^0.1002.0",
```
owensgit, line 54: I don't know a lot about this package, but looking at the [releases page](https://www.npmjs.com/package/@angular-devkit/build-angular), I get the impression that this version is related Angular v10. Do you know if that's the case? If so, I wonder if we should be using it when we're currently on Angular v11.

Screenshot 2021-12-16 at 09 28 04

owensgit (commented):
Thanks for spotting this! Just a question from me

paul-request (approved):

package.json
elenagarrone: That's what's currently working :/ 11 Hopefully when we upgrade storybook and angular 12 it will all work fine.

owensgit (approved):

PR #652, opened 2021-12-23
fix(header): update header to allow for branding

elenagarrone:
# Description

The changes allow for co branding of the header.

https://user-images.githubusercontent.com/8397116/147258037-262fedd2-5e4b-411c-801e-cddf7c6e3669.mov

## Requirements

The header should allow for two logos.

# Checklist:


aaronBery:
i noticed you haven't checked the selection to indicate the pr has sufficient unit tests... from first inspection it looks like there could be some testing around the secondary logo logic :-)

elenagarrone:
> i noticed you haven't checked the selection to indicate the pr has sufficient unit tests... from first inspection it looks like there could be some testing around the secondary logo logic :-)

Thanks! completely forgot about the tests 😂  I have added them now 🙂

elenagarrone:
Do we think this should be flagged as breaking change? It's adding the grid into the header.

owensgit (approved):

owensgit:
> Do we think this should be flagged as breaking change? It's adding the grid into the header. I don't think anyone was using this before since it didn't have the grid and the slot to put the navigation in.

The Partial Registration app is using `lg-header`, but I wouldn't have thought it's a breaking change if it's not changing any of the API, sounds like the use of grid is just an internal layout thing and the API to the consuming app remains the same?

elenagarrone:
> > Do we think this should be flagged as breaking change? It's adding the grid into the header. I don't think anyone was using this before since it didn't have the grid and the slot to put the navigation in.
> 
> The Partial Registration app is using `lg-header`, but I wouldn't have thought it's a breaking change if it's not changing any of the API, sounds like the use of grid is just an internal layout thing and the API to the consuming app remains the same?

Yes, that's correct. Registration doesn't use it the component after all - at least where I checked.

projects/canopy/src/lib/header/header.notes.ts
```diff
@@ -37,4 +37,17 @@ and in your HTML:
 | \`\`logo\`\` | A url link to the logo | string | null | Yes |
 | \`\`logoAlt\`\` | alt text to display alongside the logo | string | '2rem' | Yes |
 | \`\`logoHref\`\` | Url link if the logo is clickable | string | null | No |
+| \`\`secondLogo\`\` | A url link to the second logo | string | null | Yes |
```
aaronBery, line 40: it looks like these are technically `undefined` by default not null

projects/canopy/src/lib/header/header.component.html
```diff
@@ -1,7 +1,52 @@
...
-</a>
+<div lgContainer>
+  <div lgRow>
+    <div lgCol="12">
+      <div class="lg-header__wrapper">
+        <div class="lg-header__logos-wrapper">
+          <ng-container
+            *ngTemplateOutlet="
+              logoHref ? linkTpl : imgTpl;
+              context: { img: { logoAlt: logo, logoHref: logoHref, logo: logo } }
```
aaronBery, line 9: not 100% sure but is it possible to use the shorthand properties notation for the logo and logoHref within a ngTemplateOutlet?

Zimann (approved):

projects/canopy/src/lib/header/header.component.html
elenagarrone: No, I tried and I couldn't :/

projects/canopy/src/lib/header/header.notes.ts
elenagarrone: @aaronBery updated thanks 👍

aaronBery (approved):

PR #663, opened 2022-01-05
Add extra content projection inside the radio buttons

Zimann:
# Description

This PR: 
- adds an extra content projection element `<ng-content select="lg-hint">` so that we're able to add text below the radio button for the cases where design dictates so.

Check the PS below: 

internal_custom_text

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/radio/radio-button.component.html
+++ b/projects/canopy/src/lib/forms/radio/radio-button.component.html
@@ -6,6 +6,7 @@
   [attr.name]="name"
   [attr.value]="value"
   [attr.aria-invalid]="errorClass"
+  [attr.aria-describedby]="ariaDescribedBy || null"
   class="lg-radio-button__input"
   type="radio"
 />
@@ -13,3 +14,7 @@
 <label [attr.for]="id" class="lg-radio-button__label">
   <ng-content></ng-content>
 </label>
+
+<div *ngIf="hintPresent" class="lg-radio-button__hint">
+  <ng-content select="lg-hint"></ng-content>
+</div>
--- a/projects/canopy/src/lib/forms/radio/radio-button.component.scss
+++ b/projects/canopy/src/lib/forms/radio/radio-button.component.scss
@@ -1,5 +1,10 @@
 @import '../../../styles/mixins';
 
+.lg-radio-button__hint {
+  margin-top: var(--space-xs);
+  padding-left: calc(var(--space-lg) + var(--space-xxxs));
+}
+
 .lg-radio-button--radio {
   display: block;
   margin-bottom: var(--space-sm);
--- a/projects/canopy/src/lib/forms/radio/radio-button.component.spec.ts
+++ b/projects/canopy/src/lib/forms/radio/radio-button.component.spec.ts
@@ -1,17 +1,60 @@
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
+import { Component, DebugElement } from '@angular/core';
+import {
+  FormBuilder,
+  FormGroup,
+  FormsModule,
+  ReactiveFormsModule,
+  Validators,
+} from '@angular/forms';
 
 import { anything, instance, mock, verify, when } from 'ts-mockito';
+import { MockComponents } from 'ng-mocks';
 
 import { LgErrorStateMatcher } from '../validation/error-state-matcher';
 import { LgRadioButtonComponent } from './radio-button.component';
 import { LgRadioGroupComponent } from './radio-group.component';
+import { LgHintComponent } from '../hint';
+
+const hintTestId = 'test-hint-id';
+
+@Component({
+  template: `
+    <form [formGroup]="form" #testForm="ngForm">
+      <lg-radio-group formControlName="color">
+        Color
+        <lg-radio-button value="red">Red</lg-radio-button>
+        <lg-radio-button value="yellow"
+          >Yellow
+          <div class="lg-radio-button__content">
+            <lg-hint id="${hintTestId}">Custom text</lg-hint>
+          </div>
+        </lg-radio-button>
+        <lg-radio-button value="blue">Blue</lg-radio-button>
+      </lg-radio-group>
+    </form>
+  `,
+})
+class TestRadioButtonComponent {
+  form: FormGroup;
+
+  constructor(public fb: FormBuilder) {
+    this.form = this.fb.group({
+      color: [{ value: '', disabled: false }, [Validators.required]],
+    });
+  }
+}
 
 describe('LgRadioButtonComponent', () => {
   let component: LgRadioButtonComponent;
+  let testComponent: TestRadioButtonComponent;
   let fixture: ComponentFixture<LgRadioButtonComponent>;
+  let testFixture: ComponentFixture<TestRadioButtonComponent>;
   let errorStateMatcherMock: LgErrorStateMatcher;
   let radioGroupMock: LgRadioGroupComponent;
+  let hintDebugElement: DebugElement;
+  let inputDebugElement: DebugElement;
 
   beforeEach(
     waitForAsync(() => {
@@ -21,7 +64,13 @@ describe('LgRadioButtonComponent', () => {
       when(radioGroupMock.variant).thenReturn('segment');
 
       TestBed.configureTestingModule({
-        declarations: [LgRadioButtonComponent],
+        imports: [FormsModule, ReactiveFormsModule],
+        declarations: [
+          LgRadioButtonComponent,
+          LgRadioGroupComponent,
+          TestRadioButtonComponent,
+          MockComponents(LgHintComponent),
+        ],
         providers: [
           {
             provide: LgRadioGroupComponent,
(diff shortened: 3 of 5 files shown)


projects/canopy/src/lib/forms/radio/radio-button.component.scss
```diff
@@ -1,5 +1,9 @@
 @import '../../../styles/mixins';
 
+lg-hint {
+  width: 100%;
```
elenagarrone, line 4: I wouldn't have thought you needed this? also could you target the class instead of the tag please?

elenagarrone:
@Zimann can we get a screenshot of how it looks?

projects/canopy/src/lib/forms/radio/radio-button.component.scss
Zimann: yeah, it was needed, otherwise, it stayed on the same row 😄

Zimann:
Added a screenshot in the description ☝️

projects/canopy/src/lib/forms/radio/radio-button.component.html
```diff
@@ -13,3 +13,4 @@
 <label [attr.for]="id" class="lg-radio-button__label">
   <ng-content></ng-content>
 </label>
+<ng-content select="lg-hint"></ng-content>
```
elenagarrone, line 16: Can you modify the story so that we can have a preview of how it looks with the `lg-hint` please?

Zimann: Added an `<lg-hint>` tag to one of the radio buttons 😉

projects/canopy/src/lib/forms/radio/radio-button.component.scss
Zimann: Actually, upon better inspection, there was no need for me to add the styling, I thought it was needed because at first I had a different approach which didn't use the `lg-hint` so this styling part is out of the picture 😄

elenagarrone:
@Zimann looking good but:
1. you need to push the `rebase` as I think you only did it locally 😉 
2. the `input` should be connected to the `lg-hint` with an `aria-describedby` attribute - there is a similar implementation that you can look at in the `radio-group component` line `94`.

elenagarrone:
@Zimann can you also please update all the commits for this work to be a single commit e.g.:
`fix(radio-button): add slot for hint message`

projects/canopy/src/lib/forms/radio/radio-button.component.ts
```diff
@@ -82,6 +86,17 @@ export class LgRadioButtonComponent implements OnInit {
     return this.errorState.isControlInvalid(this.control, this.controlContainer);
   }
 
+  _hintElement: LgHintComponent;
```
elenagarrone, line 89: @Zimann could you add some unit tests please?

Zimann: Added test for this change 😉 .

elenagarrone:
@Zimann one last thing - can you squash your commits into one please?

projects/canopy/src/lib/forms/radio/radio-button.component.scss
```diff
@@ -1,5 +1,12 @@
 @import '../../../styles/mixins';
 
+.lg-radio-button {
```
elenagarrone, line 3: this can be all in a line without being split

projects/canopy/src/lib/forms/radio/radio-button.component.html
```diff
@@ -6,10 +6,14 @@
...
   [attr.aria-invalid]="errorClass"
+  [attr.aria-describedby]="ariaDescribedBy || null"
   class="lg-radio-button__input"
   type="radio"
 />
 
 <label [attr.for]="id" class="lg-radio-button__label">
   <ng-content></ng-content>
 </label>
+<div class="lg-radio-button__content">
```
elenagarrone, line 17: Could you look at doing something like in this answer please? https://stackoverflow.com/questions/35107211/how-to-check-whether-ng-content-is-empty-in-angular-2-till-now

There is no need for the content wrapper if the hint is not added. Also could it be called `lg-radio-button__hint`?

projects/canopy/src/lib/forms/radio/radio-button.component.scss
Zimann: It can, but my first thought was that we could leverage nesting so we could have a top-level `.lg-radio-button` with the suffixes (`&__content, &--radio, &__input`, etc) used for more targeted selection. If it seems more clear though, having it on one line, I will change it, let me know 😄.

elenagarrone: You could do that but in this case you have only added the `&__content` so either you add everything else or leave the selector as `lg-radio-button__hint`.

Zimann: Reverted it.

projects/canopy/src/lib/forms/radio/radio-button.component.html
Zimann: Renamed the class & and placed conditional render logic on the div wrapper.

projects/canopy/src/lib/forms/radio/radio-button.component.html
```diff
@@ -6,10 +6,15 @@
...
+  [attr.aria-describedby]="ariaDescribedBy || null"
   class="lg-radio-button__input"
   type="radio"
 />
 
 <label [attr.for]="id" class="lg-radio-button__label">
   <ng-content></ng-content>
 </label>
+
+<div *ngIf="ariaDescribedBy" class="lg-radio-button__hint">
```
owensgit, line 18: Just wondering why we need `*ngIf="ariaDescribedBy"`? From a quick look at the code, it seems that we set `aria-describedby` based on the id of the hint component. So maybe this condition doesn't need to be here? E.g. we just display the hint and then set the aria attribute accordingly?

projects/canopy/src/lib/forms/radio/radio-button.component.spec.ts
```diff
@@ -1,17 +1,60 @@
...
+@Component({
+  template: `
+    <form [formGroup]="form" #testForm="ngForm">
+      <lg-radio-group formControlName="color">
+        Color
+        <lg-radio-button value="red">Red</lg-radio-button>
+        <lg-radio-button value="yellow"
+          >Yellow
+          <div class="lg-radio-button__content">
+            <lg-hint id="${hintTestId}">Custom text</lg-hint>
```
owensgit, line 31: Do you need to add an ID here? I think the hint component adds one automatically.

owensgit (commented):
Nice work @Zimann, just a couple of comments from me :)

projects/canopy/src/lib/forms/radio/radio-button.component.html
owensgit: Oh I see in other comments that you added conditional logic so that you don't display the div wrapper if the hint isn't used. It does still seem odd that you'd tie this to the `ariaDescribedBy` attribute though, could we use the existence of the hint component instead?

projects/canopy/src/lib/forms/radio/radio-button.component.spec.ts
Zimann: Yes, it seems to be needed for the last assertion where we check if the id matches with the value of the `aria-describedBy` attribute, however; I've had some inspiration on the tests from this file (https://github.com/Legal-and-General/canopy/blob/master/projects/canopy/src/lib/forms/radio/radio-group.component.spec.ts#L35) where I've seen that the approach is similar 😉 . 

Indeed when the component is being generated in a project, that id will be automatically created 👍 .

projects/canopy/src/lib/forms/radio/radio-button.component.html
Zimann: Indeed 😄 . I've noticed that this is a way that doesn't require checking for template referencing where we would potentially have to view the child and then do extra checks on it. I figured we could use a value from an existing input (`ariaDescribedBy`) that we have. Still, I can understand if that seems odd and if a different approach is preferred, I'm down to change it like so 😉 .

projects/canopy/src/lib/forms/radio/radio-button.component.spec.ts
owensgit: Ah, gotcha, thanks 👍

projects/canopy/src/lib/forms/radio/radio-button.component.html
owensgit: My instinct is to try getting a reference to the hit component - maybe try that and see if it isn't too cumbersome? I'm thinking that we should avoid using `ariaDescribedBy`, as the developer might choose to set this independently and not use the hint component inside the radio button component.

elenagarrone: Yes, I agree with @owensgit 👍

elenagarrone (approved):
Great work @Zimann 👍

owensgit (approved):

PR #669, opened 2022-01-10
Add option to co-brand the footer

elenagarrone:
# Description

Update the footer so that it can be co-branded.
Also add the option to have buttons in the secondary links. This is for scenarios where a user has to add the cookie preference functionality. 
In addition to that the interface for the secondary link now allows a `class` field because in our specific case the cookie popup doesn't work unless there is both the `id` and the `class` specified on the element.

Also rename `footer.model.ts` to be `footer.interface.ts` for consistency.

https://user-images.githubusercontent.com/8397116/148810911-fb56bbe5-0dc9-4f51-971b-161126bc6f75.mov

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/footer/footer.component.ts
+++ b/projects/canopy/src/lib/footer/footer.component.ts
@@ -7,7 +7,7 @@ import {
   ViewEncapsulation,
 } from '@angular/core';
 
-import { Link } from './footer.model';
+import { Link, SecondaryLink } from './footer.interface';
 
 @Component({
   selector: '[lg-footer]',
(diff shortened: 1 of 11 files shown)


owensgit (approved):
Nice work 👍

projects/canopy/src/lib/footer/footer.component.ts
```diff
@@ -19,33 +19,44 @@ export class LgFooterComponent {
...
   }
-
   set logo(logo) {
     this._logo = logo;
     if (!this.logoAlt) {
       this.logoAlt = '';
     }
   }
 
+  private _secondLogo: string;
```
paul-request, line 37: I wonder if this should be `secondaryLogo` instead, to keep it consistent with the variable names for links? Bit pedantic I know.

paul-request (approved):
Just a minor comment from me, not a blocker though

projects/canopy/src/lib/footer/footer.component.ts
elenagarrone: As discussed I will create a separate PR and put the change into `next` since it will be a breaking change because as the header also needs changing.

PR #688, opened 2022-01-31
fix(tabs): remove heading from tab component to fix a11y issue

### LINKED ISSUE #676: Tabs Component a11y issue
We've had an audit carried out by Dig Inclusion and the following issue was reported for the [`lg-tabs`](https://legal-and-general.github.io/canopy/?path=/story/components-tabs-tabbed-content--standard) component:

> There is an incorrect structure that prevents the tablist from working as intended. As per the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role): 
Elements with the role `tab` must either be a child of an element with the `tablist` role or have their `id` as part of the `aria-owns` property of a `tablist`.

> In this case, the tab must be a direct child of the parent tablist. Currently, the tab is a "grandchild" element, which may cause issues.  The section needs to be restructured so that the element with the role of `tab` is a direct child of the element with role `tablist`.

image

sacalusalina:
# Description

Fix accessibility bug described in issue #676. 
The team has decided to introduce a breaking change and remove the heading from the tab system
as unnecessary.

Fix: #676

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/tabs/tabs.component.html
+++ b/projects/canopy/src/lib/tabs/tabs.component.html
@@ -1,30 +1,26 @@
 <div class="lg-tabs__list" role="tablist" [attr.aria-label]="label">
-  <lg-heading
-    class="lg-tabs__list-item-heading"
+  <button
     *ngFor="let tab of tabs; let index = index"
-    [level]="headingLevel"
+    class="lg-tabs__list-item-toggle"
+    [ngClass]="{
+      'lg-tabs__list-item-toggle--selected': selectedIndex === index
+    }"
+    role="tab"
+    [lgFocus]="isKeyboardEvent && selectedIndex === index"
+    [attr.tabIndex]="index !== selectedIndex ? '-1' : 0"
+    [attr.aria-selected]="index === selectedIndex"
+    [attr.aria-controls]="'tab-item-content-' + id + '-' + index"
+    [attr.keyboard-focus]="(isKeyboardEvent && selectedIndex === index) || null"
+    [id]="'tab-item-heading-' + id + '-' + index"
+    (click)="navigateToTab(index)"
+    (blur)="blur(index)"
+    (keyup)="keyboardNavigation($event)"
   >
-    <button
-      class="lg-tabs__list-item-toggle"
-      [ngClass]="{
-        'lg-tabs__list-item-toggle--selected': selectedIndex === index
-      }"
-      role="tab"
-      [lgFocus]="isKeyboardEvent && selectedIndex === index"
-      [attr.tabIndex]="index !== selectedIndex ? '-1' : null"
-      [attr.aria-selected]="index === selectedIndex"
-      [attr.aria-controls]="'tab-item-content-' + id + '-' + index"
-      [attr.keyboard-focus]="(isKeyboardEvent && selectedIndex === index) || null"
-      [id]="'tab-item-heading-' + id + '-' + index"
-      (click)="navigateToTab(index)"
-      (blur)="blur(index)"
-      (keyup)="keyboardNavigation($event)"
-    >
+    <span class="lg-tabs__list-item-heading">
       <ng-container *ngTemplateOutlet="tab.navItemTemplate"></ng-container>
-    </button>
-  </lg-heading>
+    </span>
+  </button>
 </div>
-
 <div
   *ngFor="let tab of tabs; let index = index"
   class="lg-tabs__content"
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -90,7 +90,7 @@
       <div lgCol="12" lgColLg="9" lgColMdOffset="0">
         <lg-card lgPaddingTop="none">
           <lg-card-content>
-            <lg-tabs [headingLevel]="2" label="Annuities">
+            <lg-tabs label="Annuities">
               <lg-tab-item>
                 <lg-tab-item-heading>Tab 1</lg-tab-item-heading>
                 <lg-tab-item-content>
--- a/projects/canopy/src/lib/tabs/tabs.component.spec.ts
+++ b/projects/canopy/src/lib/tabs/tabs.component.spec.ts
@@ -43,7 +43,7 @@ describe('LgTabsComponent', () => {
 
   beforeEach(() => {
     fixture = MockRender(`
-      <lg-tabs [headingLevel]="1">
+      <lg-tabs>
         <lg-tab-item>
           <lg-tab-item-heading>Heading 1</lg-tab-item-heading>
             <lg-tab-item-content>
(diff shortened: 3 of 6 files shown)


projects/canopy/src/lib/tabs/tabs.component.html
```diff
@@ -1,8 +1,10 @@
-<div class="lg-tabs__list" role="tablist" [attr.aria-label]="label">
+<div class="lg-tabs__list">
   <lg-heading
     class="lg-tabs__list-item-heading"
     *ngFor="let tab of tabs; let index = index"
     [level]="headingLevel"
+    role="tablist"
```
markblandford, line 6: This won't address the issue. What you've done here is make each tab an element with the `tablist` role, so we end up with multiple tablists, each with a single tab instead of one tablist with multiple tabs.

The issue in #676 , is that the button (from line 9) has the `role="tab"` attribute rather than the grandparent, `lg-heading`. For a tablist to be accessible, the elements with the `role="tab"` must be a direct child of the tablist element (or have their id as part of the `aria-owns` property of a tablist).

I think the tablist element needs to remain as it was, and either the direct children have the `role="tab"` attribute or possibly the order of the elements needs to be changed (so the `lg-heading` is a child of the `button` element).

markblandford (changes requested):

HarshaDewani:
Hey @sacalusalina 
A minor one, maybe you can specify the name of the file or component in the commit message that you are trying to work on. For e.g. something like - `fix(component/file changed): "your commit message here describing the changes"` as these commit messages are then used to create release notes, so might be super useful at that point to understand what changes have been done :) 

If you are trying to make changes for different components in the same PR then maybe you can add different commits one for each. I believe in this PR we are trying to fix the accessibility issues for two different components? If that's the case then maybe we can two different commits, one with changes in `carousel component`  - `fix(carousel)-"your commit message here"` and the other for changes in `tabs component` -`fix(tabs)-"your commit message here"` Shout if you need any help on this :)

markblandford (dismissed):
Good work Alina.

The changes to the tab I think now not only resolve the issue where the `role="tab"` element wasn't a child of `role="tablist"` but I think it also resolves the syntax issue where the heading element previously contained non-phrasing content ([the only permitted child of a `<h1>-<h6>` element is 'phrasing content'](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)).

projects/canopy/src/lib/tabs/tabs.component.html
sacalusalina: I committed the suggested changes. Thank you.

dannyoz (dismissed):

dSotropa (dismissed):

paul-request (dismissed):

projects/canopy/src/lib/tabs/tabs.component.html
```diff
@@ -1,30 +1,26 @@
 <div class="lg-tabs__list" role="tablist" [attr.aria-label]="label">
-  <lg-heading
```
elenagarrone, line 2: Moving the heading inside the button will cause another accessibility issue because buttons cannot contain heading. The heading should be wrapping the button as it was before 🤔 

Maybe the solution is to use `aria-owns` on the `tablist`. Something like:

```
<div class="lg-tabs__list" role="tablist" [attr.aria-label]="label" aria-owns="tab1-id tab2-id etc">
```

where `tab1-id tab2-id etc` is the button id for all the tabs.

elenagarrone:
@sacalusalina could you please squash your commits and write something like:

```
fix(tabs): handle issue with the roles tablist and tab

fix: #676
```

elenagarrone (changes requested):

projects/canopy/src/lib/tabs/tabs.component.html
markblandford: Ah yes, and headings cannot contain buttons either: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

markblandford: Thinking about it, does it even make sense for tabs to have heading markup? It feels like the tabs should just be buttons.

elenagarrone: Yes, I think we used to have the heading/button structure like that in our project and Grant from DIG once told me there was no real need for the tab to be a heading.

paul-request: Yeah I agree, would have to be a breaking change though. Which is fine, would jus tneed t get that work scheduled in to update it.

owensgit:
One of our best practices in Canopy is paying special attention to the commit messages, and maintaining a good commit history, especially considering this is an open source repo. It looks like your commit messages contain project-specific ticket numbers as the scope, e.g.

```
fix(18155 18510): etc
```

Ideally the scope should be the component (or area) the change is being made on, and they should be separated to ensure each commit makes changes to each component (where possible)  e.g.

```
fix(tabs): etc
fix(carousel): etc
```

Can you update the commit messages please?

projects/canopy/src/lib/tabs/tabs.component.html
owensgit: This sounds good, thanks for flagging @elenagarrone. In terms of the breaking change, we could merge it into the `next` branch and then release alongside the other [upcoming breaking changes](https://github.com/Legal-and-General/canopy/issues) now if needed as Canopy v7, and then leave the Angular upgrade etc for the next major version.

elenagarrone: @owensgit I thought about it and I think the problem is that it might slow down quite a lot the merging of this fix. We will have to rebase `next` and then merge it and the teams will have to apply all the breaking changes at once.

I guess it depends if this is urgent and if the team (@markblandford) has time to make all the other changes required by the `next` version.

markblandford (changes requested):

markblandford (approved):

projects/canopy/src/lib/carousel/auto-play/auto-play.component.ts
```diff
@@ -19,7 +19,5 @@ export class LgAutoplayComponent {
 
   @HostBinding('class.lg-carousel-autoplay') class = true;
 
-  @HostBinding('attr.aria-live') role = 'polite';
```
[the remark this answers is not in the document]
owensgit: I agree, it looks like another PR has been created for this, but this still needs to be taking out of this PR :)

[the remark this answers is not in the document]
sacalusalina: I added the attribute back

markblandford (approved):

elenagarrone:
@sacalusalina before we approve this change we need the below circled commit to be removed as it's in the other PR:
Screenshot 2022-02-07 at 10 01 02

elenagarrone (approved):

paul-request (approved):

PR #689, opened 2022-01-31
fix(carousel): carousel focus fix

Nupur04:
keyboard focus carousel fix

# Description
Stop carousel when any button inside carousel item has been clicked and fix broken view on tab
switch
If a CTA on one of the advertising tiles has focus, and the animation / scrolling is on, it breaks.
Tabbing between the controls on the ads DCI-17572

selectedItemIndexSet set to BehaviourSubject so that selectedItemIndexSet$ be exposed by targeting the Carousel with ViewChild decorators.
Advertising Carousel scroll behaviour when CTA is interacted with or has focus.

Fixes # (issue)
DCI-18540, DCI-17572

## Requirements

Screenshot 2022-01-31 at 16 23 09
Screenshot 2022-01-31 at 16 22 54

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
+++ b/projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
@@ -1,3 +1,5 @@
+@import '../../../styles/mixins.scss';
+
 .lg-carousel-autoplay {
   position: absolute;
   right: var(--space-sm);
@@ -12,5 +14,9 @@
     height: var(--space-xxl);
     font-size: var(--text-fs-3-size);
     cursor: pointer;
+
+    &:focus {
+      @include lg-outer-focus-outline(var(--default-focus-color));
+    }
   }
 }
--- a/projects/canopy/src/lib/carousel/carousel.component.spec.ts
+++ b/projects/canopy/src/lib/carousel/carousel.component.spec.ts
@@ -9,6 +9,7 @@ import {
 import { By } from '@angular/platform-browser';
 
 import { MockComponents } from 'ng-mocks';
+import { Subscription } from 'rxjs';
 
 import { HeadingLevel, LgHeadingComponent } from '../heading';
 import { LgIconComponent } from '../icon';
@@ -56,6 +57,7 @@ class TestWrapperComponent {}
 describe('LgCarouselComponent', () => {
   let component: LgCarouselComponent;
   let fixture: ComponentFixture<TestWrapperComponent>;
+  let subscription: Subscription;
 
   const timerSelectionCheck = () => {
     component.ngAfterContentInit();
@@ -86,6 +88,10 @@ describe('LgCarouselComponent', () => {
     fixture = TestBed.createComponent(TestWrapperComponent);
     component = fixture.debugElement.children[0].children[0].componentInstance;
     fixture.detectChanges();
+  });
+
+  afterEach(() => {
+    subscription?.unsubscribe();
   });
 
   it('should create', () => {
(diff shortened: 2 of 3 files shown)


sacalusalina (approved):

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -91,16 +100,18 @@ export class LgCarouselComponent implements AfterContentInit, OnDestroy {
...
     if (this.autoPlay) {
       this.setAutoPlayInterval();
     }
   }
 
-  ngOnDestroy(): void {
-    this.unsubscribe.next();
-    this.unsubscribe.complete();
+  ngOnDestroy() {
+    this.subscription.unsubscribe();
```
dannyoz, line 114: Is.unsubscribe is used in the setAutoplayInterval on line 92.` takeUntil(this.unsubscribe)` Removing this will cause a potential memory leak when the component is destroyed.  I suggest you bring it back and use `takeUntil(this.unsubscribe)` as part of the `this.selectedItemIndexSet$` subscription on line 103

dannyoz (changes requested):

projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
```diff
@@ -12,5 +14,9 @@
     height: var(--space-xxl);
     font-size: var(--text-fs-3-size);
     cursor: pointer;
+
+    &:focus {
+      @include lg-outer-focus-outline(var(--default-focus-color));
```
elenagarrone, line 19: This is ok but just FYI you don't have to pass the default colour since it's applied automtically.

projects/canopy/src/lib/carousel/carousel.component.spec.ts
```diff
@@ -144,9 +144,11 @@ describe('LgCarouselComponent', () => {
     });
 
     it('should select the chosen carousel item as expected', () => {
-      expect(component.selectedItemIndex).toBe(0);
-      bullet2.nativeElement.click();
-      expect(component.selectedItemIndex).toBe(1);
+      component.selectedItemIndexSet$.subscribe((selectedItem) => {
```
elenagarrone, line 147: You should unsubscribe these e.g.:

```
let subscription: Subscription;
...
afterEach(() => {
  subscription.unsubscribe();
});
...
it('should select the chosen carousel item as expected', () => {
  subscription = component.selectedItemIndexSet$.subscribe((selectedItem) => {
    ...
});

File: projects/canopy/src/lib/carousel/carousel.component.ts [hunk 4]
```diff
@@ -91,14 +92,19 @@ export class LgCarouselComponent implements AfterContentInit, OnDestroy {
...
+      });
     this.carouselItemCount = this.carouselItems.length;
     this.selectCarouselItem(0);
     if (this.autoPlay) {
       this.setAutoPlayInterval();
     }
   }
 
-  ngOnDestroy(): void {
+  ngOnDestroy() {
```
elenagarrone, line 107: Could you please add back the return type?

projects/canopy/src/lib/carousel/auto-play/auto-play.component.scss
Nupur04: Yes, I was expecting it to apply the default colour but but for some reason it was not and it came as an accessibility bug in testing so added the style.

dannyoz (approved):

owensgit:
@Nupur04 just taking a look but don't fully understand the changes, could you add a more in-depth description of the issue this change is fixing to the original PR description please?

owensgit:
@Nupur04 thanks for the description, the code looks good!  

However one of our best practices in Canopy is to squash individual commits, ensuring that the commit message is as descriptive as possible, and possibly adding a commit body if it needs further details.

Also, ideally, each commit should address one fix - I can see you have two different fixes in this PR.

If possible, it would be great to change the commits on this branch so that we have two commits, one for each fix with special attention to the commit message.

E.g. maybe something like this:

```
fix(carousel): stop when button inside is clicked

fix(carousel): broken layout when button inside gains focus
```

If that's not possible, or if I understood it wrong, maybe you could squash it into one commit and provide a commit body/long description, something like:

```
fix(carousel): stop on button click & broken layout
stop carousel when any button inside carousel item has been clicked and fix issue where a button gains focus during animation/scrolling and the layout breaks
```

Nupur04:
@owensgit I've squashed the commits into one and added the body message to explain it better in this PR.

elenagarrone (approved):

markblandford (approved):

paul-request (approved):

PR #708, opened 2022-02-16
Upgrade styles to CSF format

paul-request:
# Checklist:


.storybook/preview-head.html
```diff
@@ -31,4 +31,8 @@
   .sbdocs.sbdocs-a, .sbdocs.sbdocs-p {
     font-size: 14px;
   }
+
```
paul-request, line 34: There is a bug where an empty doc block is displayed when using the component parameters to hide the code preview, so this just prevents it from displaying for now

owensgit: Maybe we should have a comment here referencing the bug? Is there an open issue for it?

owensgit (commented):
Thanks Paul, just one small note from me

.storybook/preview-head.html
paul-request: Yeah I did think about that after raising the PR. I can't see an open issue for it, need t get round t raising one for it also

paul-request:
> Thanks Paul, just one small note from me

@owensgit I managed to find some info on adding documentation-only `mdx` files, so have added those instead for the ones where there is only some docs and nothing to render to the canvas (spacing and mixins)

owensgit (approved):

PR #713, opened 2022-02-23
fix(hero-card-notification): fix link not clickable in notification

AlexandruHoisan:
fix link not being clickable in the notification message, because of the overlay covering it

# Description

Because of the overlay, the link in the notification message could not be interacted with. The solution was to add ``pointer-events: none;``, transforming the overlay into a layer that should ignore / pass-through click-, tap-, scroll- and hover events.

## Requirements

Screenshot:
Screenshot 2022-02-21 at 14 50 37

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/variables/components/_hero.scss
+++ b/projects/canopy/src/styles/variables/components/_hero.scss
@@ -2,9 +2,9 @@
   --hero-padding: var(--space-sm);
   --hero-color: var(--color-white);
   --hero-bg-color: var(--color-super-blue-dark);
-  --hero-overlay-bg-color: var(--color-black);
   --hero-rule-color: rgba(255, 255, 255, 0.2);
   --hero-bg-alpha: 0.1;
+  --hero-overlay-bg-color: rgba(0, 0, 0, var(--hero-bg-alpha));
 
   --hero-link-color: var(--hero-color);
   --hero-link-hover-color: var(--hero-color);
--- a/projects/canopy/src/lib/hero/hero-card-notification/hero-card-notification.component.html
+++ b/projects/canopy/src/lib/hero/hero-card-notification/hero-card-notification.component.html
@@ -1,4 +1,3 @@
-<div class="lg-hero-card-notification__overlay" aria-hidden="true"></div>
 <div class="lg-hero-card-notification__icon">
   <ng-content select="lg-icon"></ng-content>
 </div>
--- a/projects/canopy/src/lib/hero/hero-card-notification/hero-card-notification.component.scss
+++ b/projects/canopy/src/lib/hero/hero-card-notification/hero-card-notification.component.scss
@@ -6,6 +6,8 @@
   position: relative;
   display: inline-flex;
   flex-direction: row;
+  background-color: var(--hero-overlay-bg-color);
+  border-radius: var(--border-radius-sm);
 
   @include lg-breakpoint(lg) {
     padding-right: var(--space-sm);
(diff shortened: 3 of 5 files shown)


HarshaDewani (approved):

paul-request:
This appears to be a really odd way of adding a background color to an existing element 🤔 @owensgit @elenagarrone does anyone know why it was done this way?

AlexandruHoisan:
> This appears to be a really odd way of adding a background color to an existing element 🤔 @owensgit @elenagarrone does anyone know why it was done this way? surely the overlay should just wrap the content and have the correct background applied?

Another fix could be removing the overlay completely and adding some background style to the actual element.
I chose the ``pointer-events: none;`` as it was the least intrusive way of fixing this. :) 
image

paul-request:
@AlexandruHoisan it seems even with pointer events, there would still be an accessibility issue anyway? Possibly, just moving the overlay after the content in the template would also work?

paul-request:
@AlexandruHoisan Ok, I can see this is probably to do with adding opacity to the overlay... but this can be achieved using `rgba(0,0,0,0.1)`, maybe needs a calc/mixin to use the `--hero-bg-alpha` alongside the `--hero-overlay-bg-color`.

AlexandruHoisan:
> @AlexandruHoisan Ok, I can see this is probably to do with adding opacity to the overlay... but this can be achieved using `rgba(0,0,0,0.1)`, maybe needs a calc/mixin to use the `--hero-bg-alpha` alongside the `--hero-overlay-bg-color`.
> 
> This should work I think: `background-color: rgba(var(--hero-overlay-bg-color), var(--hero-bg-alpha));`
> 
> Or, maybe it will need to be: `background-color: #{'rgba(var(--hero-overlay-bg-color), rgba(var(--hero-bg-alpha))'};`
> 
> Can't remember if SASS compilation will handle this or not?

@paul-request I agree with you that the overlay method is masking a questionable implementation so I removed the overlay completely and now we have this:
* the ``--hero-overlay-bg-color`` is now ``rgba(0, 0, 0, var(--hero-bg-alpha))`` from ``var(--color-black)``
* the ``overlay`` was removed completely from both ``lg-hero-card-notification`` and ``lg-hero-header`` and was added as bg color to the parent .
The overlay approach felt a bit hacky and I think these approach is more suited in this instance.

paul-request (approved):

projects/canopy/src/styles/variables/components/_hero.scss
```diff
@@ -2,9 +2,9 @@
   --hero-padding: var(--space-sm);
   --hero-color: var(--color-white);
   --hero-bg-color: var(--color-super-blue-dark);
-  --hero-overlay-bg-color: var(--color-black);
   --hero-rule-color: rgba(255, 255, 255, 0.2);
   --hero-bg-alpha: 0.1;
+  --hero-overlay-bg-color: rgba(0, 0, 0, var(--hero-bg-alpha));
```
HarshaDewani, line 7: I don't see any overlay now so maybe we can rename this to ```---hero-card-notification-bg-color``` ?

paul-request: Good point @HarshaDewani

AlexandruHoisan: Yea, I thought about that but it's still like an overlay bg for those components. In hero header it was also used as an overlay over the component, that is why I left it as overlay :)

owensgit (approved):

projects/canopy/src/styles/variables/components/_hero.scss
HarshaDewani: I am happy to keep it as it is given the fact changing css vars name can cause breaking changes.

PR #730, opened 2022-03-16
Add actions to combine dependabot PRs into one

elenagarrone:
# Description

Screenshot 2022-03-09 at 15 44 04

This PR:

- Adds a scheduled and ad-hoc workflow for batching Dependabot PRs
    - This will fetch all Dependabot PRs that are open, adding their commits into a combined PR branch (which is created if it doesn't exist)
    - This means that manually combining dependency updates is a thing of the past
    - ...and will save everyone a lot of time :)
- Changes the dependabot config to auto-rebase its PRs
- Removes the Dependabot section of the PR template

How this works in practice:

- Every Monday and Wednesday morning the scheduled workflow combines any open Dependabot PRs into one
- The combined PR runs the full checks as per any other PR
- Approve and merge

NB: If there are conflicts (which happens if two adjacent dependencies in `package.json` have updates), just merge the combined PR, then run the ad-hoc script and the remaining Dependabot PRs should be mergeable)
NB2: the dependabot PRs will not run the pull request and code analysis anymore. Only the combined branch will.

*The code has been tested in our private FEP repo and works fine but I might need to update the `GITHUB_TOKEN` once this PR gets merged.*

# Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -55,3 +55,4 @@ updates:
         versions: ["6.x"]
       - dependency-name: "@storybook/theming"
         versions: ["6.x"]
+      - dependency-name: "jasmine-marbles"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,16 +6,17 @@ updates:
     # default location of `.github/workflows`
     directory: '/'
     schedule:
+      day: 'sunday'
       interval: 'weekly'
       time: '00:00'
       timezone: 'Europe/London'
-    rebase-strategy: 'disabled'
 
   # Maintain dependencies for npm
   - package-ecosystem: 'npm'
     # Files stored in repository root
     directory: '/'
     schedule:
+      day: 'sunday'
       interval: 'weekly'
       time: '00:00'
       timezone: 'Europe/London'
--- a/.github/workflows/dependabot.adhoc.yml
+++ b/.github/workflows/dependabot.adhoc.yml
@@ -1,0 +1,26 @@
+name: Dependabot Batcher Ad-hoc
+on: workflow_dispatch
+
+jobs:
+  # Batches Dependabot PRs into one by merging them into a combined branch, then raising a new PR
+  dependabot-batcher:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/setup-node@v2.5.1
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: 'Combine Dependabot PRs'
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const combineDependabotPrs = require('./.github/workflow-scripts/combine-dependabot-prs.js');
+            return await combineDependabotPrs({
+              github,
+              context,
+            });
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   verify:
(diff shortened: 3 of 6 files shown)


.github/workflows/pull_request.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   verify:
+    if: ${{ !startsWith(github.head_ref, 'dependabot/')  }}
```
owensgit, line 9: Will this not run the checks on Dependabot PRs?

.github/workflows/dependabot.adhoc.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: 'Combine Dependabot PRs'
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const combineDependabotPrs = require('./.github/workflow-scripts/combine-dependabot-prs.js');
+            return await combineDependabotPrs({
+              github,
+              context,
+            });
```
owensgit, line 26: Am I right by saying this job is the same as in the other two workflow riles? I'm not super familiar with Github workflows, is there a way to store a job and then call on it inside multiple workflows instead of repeating it?

.github/dependabot.yml
```diff
@@ -54,3 +55,4 @@ updates:
         versions: ["6.x"]
       - dependency-name: "@storybook/theming"
         versions: ["6.x"]
+      - dependency-name: "jasmine-marbles"
```
owensgit, line 58: Wondering why we're ignoring this one?

owensgit (commented):
Nice work @elenagarrone! A few questions from me 👍

.github/workflows/pull_request.yml
owensgit: If not, I can understand why we wouldn't run linting, formatting etc, but I think it's important to run the build to make sure nothing breaks with the updates. May have misunderstood though :)

elenagarrone: That's right, there is no need to waste github actions minutes since the workflow will run on the new combined PR

.github/workflows/dependabot.adhoc.yml
elenagarrone: This does the same thing as `dependabot.scheduled.yml`. The only difference is that their run is triggered by different events: one manually bu a user (adhoc) and the other one on schedule.

There is a way of reusing workflows but we still need to look into it as it's a fairly new feature. 
We will be refactoring this code once we have time to investigate the above 👍

.github/dependabot.yml
elenagarrone: Sorry, forgot to explain this in the description. We have had the dependabot PR open for `jasmine-marbles` for a very long time. I have tried to update it multiple times but when running in the workflow it always fail (locally it's fine) and I can't figure out the reason.
If i don't ignore the dependency here it will try to add it to the combined PR each time and block it from being merged without manual intervention.

.github/workflows/pull_request.yml
owensgit: Ah I see, so the combined PR will run them, just not the individual Dependabot ones.

Can we prevent the Dependabot ones from being merged since they can't be verified the same as others? Or does that happen anyway due to branch protection rules?

.github/workflows/dependabot.adhoc.yml
owensgit: Thanks, sounds good 👍

.github/workflows/pull_request.yml
elenagarrone: Yes, that happens automatically since they will not be passing the `verify` stage which is required to be able to merge.

.github/dependabot.yml
elenagarrone: Done 👍

paul-request (approved):

owensgit (approved):

PR #731, opened 2022-03-16
style(hero-img): style updates

IdaHW:
Style updates to the Hero Image component

# Description

Style updates to the Hero Image component; background image behaviour on smaller viewports and browsers, Hero Image Card background opacity.

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/hero-img/hero-img-card/hero-img-card.component.scss
+++ b/projects/canopy/src/lib/hero-img/hero-img-card/hero-img-card.component.scss
@@ -4,12 +4,14 @@
   background-color: var(--hero-img-bg-color);
   display: block;
   margin-top: 20rem;
+  min-height: 15rem;
   padding: var(--space-md) var(--space-md) var(--space-xxs);
 
   @include lg-breakpoint('md') {
+    background-color: rgba(var(--hero-img-bg-color-rgb), 0.9);
     margin-top: var(--space-xxl);
     margin-bottom: var(--space-xxl);
-    opacity: 0.9;
+    min-height: auto;
     width: 50%;
   }
 
--- a/projects/canopy/src/lib/hero-img/hero-img.component.scss
+++ b/projects/canopy/src/lib/hero-img/hero-img.component.scss
@@ -2,11 +2,16 @@
 
 .lg-hero-img {
   background-repeat: no-repeat;
-  background-position: 100% 0;
-  background-size: 215%;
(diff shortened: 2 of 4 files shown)


projects/canopy/src/lib/hero-img/hero-img-card/hero-img-card.component.scss
```diff
@@ -4,12 +4,14 @@
   background-color: var(--hero-img-bg-color);
   display: block;
   margin-top: 20rem;
+  min-height: 15rem;
   padding: var(--space-md) var(--space-md) var(--space-xxs);
 
   @include lg-breakpoint('md') {
+    background-color: rgba(var(--hero-img-bg-color-rgb), 0.9);
```
elenagarrone, line 11: We tend to avoid using `rgb` colours. Using the `hex` value and having the `opacity` should work.

IdaHW: @elenagarrone in this case I had to add a `rgb` value as to only apply the opacity to the container and not its content. I doesn't work with a `hex` value

elenagarrone: @IdaHW thanks for clarifying 👍  Then I would probably remove the `--hero-img-bg-color` and use the rgb value on line 4 too. Otherwise users that want to update the "theme" will have to update two variables instead of one.

owensgit: Ah yes, we could just keep `--hero-img-bg-color-rgb` for both and drop `--hero-img-bg-color` if it's not used in any other places.

@IdaHW if you're not familiar, in our self serve apps we tend to do a lot of theming/syndication, where the app is given a different colour treatment for a particular brand, using the colour variables. E.g. an app might be styled with Sainsbury's brand colours for those with a Sainsbury's workplace pension.

owensgit (approved):

projects/canopy/src/lib/hero-img/hero-img-card/hero-img-card.component.scss
IdaHW: Thank you @elenagarrone - I've updated the variables. Merged master into my feature branch but am getting linting errors on the merge. Might close this pr and create a new one from master to resolve

IdaHW:
Closing PR as creating an updated branch

PR #772, opened 2022-03-30
fix(segment): stack input is now set correctly

elenagarrone:
# Description

The stack input for the segment component was not working when added to an application (worked on Storybook).
This has now been fixed (tested within an application):

https://user-images.githubusercontent.com/8397116/160885436-47298e6a-9900-4c23-adac-88fff17ebef7.mov

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/radio/radio-group.component.ts
+++ b/projects/canopy/src/lib/forms/radio/radio-group.component.ts
@@ -1,4 +1,5 @@
 import {
+  AfterContentInit,
   Component,
   ContentChild,
   ContentChildren,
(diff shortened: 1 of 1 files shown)


projects/canopy/src/lib/forms/radio/radio-group.component.ts
```diff
@@ -157,6 +152,14 @@ export class LgRadioGroupComponent implements ControlValueAccessor {
     }
   }
 
+  ngAfterContentInit(): void {
+    if (this.radios && this.stack) {
+      this.radios.toArray().forEach((radio) => {
```
Devaney, line 157: should radios be typed here `: LgRadioButtonComponent`? if not can remove the parenthesis?

elenagarrone: I think the type is unnecessary in this case as `this.radios` already implies it. 
Canopy has different lint rules so we currently keep the parenthesis across the codebase. I have a ticket to update the lint rules to match those of the monorepo 🙂

Devaney (approved):

AlexandruHoisan (approved):

paul-request (approved):

PR #848, opened 2022-05-13
docs(gh-pages): fix environment issue with deployment

elenagarrone:
# Description

Remove environment logic as it is unnecessary to create a different environment for each branch.
I have set the environment to the one GitHub pages already uses.

Workflow test run that replicates `master`: https://github.com/Legal-and-General/canopy/runs/6420417310

Workflow to deploy this environment: https://github.com/Legal-and-General/canopy/actions/runs/2318593788

Tested deployment status update using the `job.status` here: https://github.com/Legal-and-General/canopy/runs/6469799176?check_suite_focus=true

Screenshot 2022-05-17 at 12 39 19

### DIFF
--- a/.github/workflows/branch_pages_deploy.yml
+++ b/.github/workflows/branch_pages_deploy.yml
@@ -56,7 +56,6 @@ jobs:
             const createDeployment = require('./.github/workflow-scripts/gh-pages/create-gh-deployment.js');
             return await createDeployment({
               sha: '${{ fromJSON(steps.prData.outputs.result).sha }}',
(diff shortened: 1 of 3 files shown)


.github/workflows/branch_pages_deploy.yml
```diff
@@ -88,6 +87,10 @@ jobs:
               exec,
             });
 
+      - name: 'Checking out current branch'
```
elenagarrone, line 90: This is needed because in the previous step we checkout `gh-pages`. We want the rest of the workflow to run from the branch that triggered the workflow, not `gh-pages`.

pauleustice (approved):

PR #849, opened 2022-05-13
fix: carousel icon alignment

### LINKED ISSUE #432: Play Icon alignment / missing pause-spot icon
The canopy play icon path is not centred and needs to be adjusted.  This was discovered whilst working on the autoplay functionality of the carousel.

Screenshot 2021-06-25 at 09 41 55
Screenshot 2021-06-25 at 09 43 13

In the screenshots above the lgIcon component is centred with css inside a white rounded div element.

We also have a play-spot icon in the library which looks to be aligned correctly but no pause-spot equivalent so Ideally one should be created.

dannyoz:
# Description

The carousel play / pause icons were miss-aligned on iOS / macOS Safari and windows Firefox. This was due to inconsistencies with the centring of lg-icons within the circular button. This has been fixed by replacing to icons with SVGs that handle the style of both the icon element and the enclosing circle. This will be consistent cross-device and cross-OS.

Resolves #432

# Checklist:


sacalusalina (approved):

ciprianandrusca (dismissed):

projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
```diff
@@ -5,7 +5,13 @@
...
 >
   <span class="lg-visually-hidden">Stop Animation</span>
-  <lg-icon name="pause"></lg-icon>
+  <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+    viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
+  <g>
+    <path class="carousel-icon" d="M250.09,1C112.63,1,1.2,112.43,1.2,249.89s111.43,248.89,248.89,248.89s248.89-111.43,248.89-248.89
+      S387.55,1,250.09,1z M231.82,353.79h-77.64V146h77.64V353.79z M346,353.79h-77.64V146H346V353.79z"/>
+  </g>
+  </svg>
```
owensgit, line 14: I'm a bit weary of replacing `lg-icon` with an SVG directly in the file. Ideally we'd use the icon component so that there's always consistency across all icons in the app. Is there a problem with the original SVG file for these icons? If so, I think it would be better to edit/update those instead (while preserving the ability to change the colour). I've had issues before where icons weren't centered inside the SVG canvas properly, so may be something like that? Otherwise, perhaps you could align them by nudging a container? If there's no other way I'm happy with this change, but it's worth checking.

elenagarrone: Yes, I agree with what @owensgit said. Could you point out what styles were problematic in the `lg-icon`? we might want to update the `lg-icon` itself 🤔

dannyoz: The problem is the svg was just for the play / pause icons and does not include the circle shape (which was handled by the button) These svgs include all shapes needed for the buttons

dannyoz: I could add them to the list of lg-icons as **carousel-play** and **carousel-pause**

dannyoz: I think adding new icons rather than editing existing ones is a better way to go

elenagarrone: Yes, I was looking at the existing icons and I only found the play one with the circle but even that would be wrong since the actual play would take the colour of the background e.g. 

Screenshot 2022-05-17 at 10 58 23

I'm not sure I see these icons used elsewhere so I'm wondering if we should just keep them in here? @owensgit what do you think? 
Usually for videos we keep the play buttons of the players.

elenagarrone:
@dannyoz would you mind updating the commit message to something like:
```
fix(carousel): update icons alignment
```

projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
owensgit: I'd be happy to keep the SVG code in this component, as a one off for the carousel, considering those icons won't be used anywhere else 👍

owensgit (dismissed):

projects/canopy/src/lib/carousel/auto-play/auto-play.component.html
elenagarrone: @dannyoz I just deployed the branch and noticed that the play/pause icons take the background colour so I'd say you can use the `play-spot` icon and then ask one of the designers to create a new icon `pause-spot`. They will need to add it to the design library too.

projects/canopy/src/assets/icons/play-spot.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 32 32"><defs><path id="a" d="M15 28c6.627 0 12-5.373 12-12S21.627 4 15 4 3 9.373 3 16s5.373 12 12 12zm-3-17 8 5-8 5V11z"/></defs><use fill-rule="evenodd" xlink:href="#a"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 500 500"><defs><path id="a" d="M250.09 1.21C112.63 1.21 1.2 112.64 1.2 250.11S112.63 499 250.09 499s248.89-111.43 248.89-248.89S387.55 1.21 250.09 1.21zm-59.94 352.46-.84-207.14 177.72 104.29-176.88 102.85z"/></defs><use fill-rule="evenodd" xlink:href="#a"/></svg>
```
elenagarrone, line 1: The play spot should stay what it was 🤔

projects/canopy/src/assets/icons/pause-spot.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 32 32"><defs><path id="a" d="M250.09 1C112.63 1 1.2 112.43 1.2 249.89s111.43 248.89 248.89 248.89 248.89-111.43 248.89-248.89S387.55 1 250.09 1zm-18.27 352.79h-77.64V146h77.64v207.79zm114.18 0h-77.64V146H346v207.79z"/></defs><use fill-rule="evenodd" xlink:href="#a"/></svg>
```
elenagarrone, line 1: Did you get this from a designer?

projects/canopy/src/assets/icons/play-spot.svg
dannyoz: The play spot needed updating because the alignment was wrong

elenagarrone: The thing is that the `viewBox` is usually set to `0 0 32 32` 🤔  we need to makes sure the designer that is creating the icons is following the same rules as the other icons and that they update the deign library too.

projects/canopy/src/assets/icons/pause-spot.svg
dannyoz: Yes they were designed by Paul Farrignton

dannyoz:
@

> @dannyoz looks like the `pause-spot` is not showing at all - it's just an empty box 🤔 you can see it from the diffs tab too.

@elenagarrone the file is there now looks like an attribute was missing and that made it transparent somehow 🤷‍♂️  view boxes are now  0 0 32 32 too

elenagarrone:
> @
> 
> > @dannyoz looks like the `pause-spot` is not showing at all - it's just an empty box 🤔 you can see it from the diffs tab too.
> 
> @elenagarrone the file is there now looks like an attribute was missing and that made it transparent somehow 🤷‍♂️ view boxes are now 0 0 32 32 too

Thank you! I will deploy the branch once the `verify` step is done and double check it all looks ok 👍

elenagarrone (approved):

Issue #698, opened 2022-02-08
Deprecations & upcoming breaking changes - v8.0.0


elenagarrone:
ℹ️  &nbsp; Small breaking changes are more easily consumed when shipped together in a major version bump.

💡  &nbsp; When identifying such a change, such as removing a CSS var, add a deprecation notice and where possible provide deprecation warnings in the console.

💬   &nbsp; Add a comment to this issue to log current deprecations and upcoming minor breaking changes.

🚀  &nbsp; The pull request should be merged into the `next` branch, where it can be part of the next **pre-release**. At an appropriate time, `next` can be merged into `master` creating a major release of Canopy (with some nice release notes).

♻️  &nbsp; When released, this issue can be closed down and another one spun up for the next batch.

elenagarrone:
Within the Angular 12 upgrade PR we have dropped support for IE11.

elenagarrone:
Released: https://github.com/Legal-and-General/canopy/releases/tag/v8.0.0

PR #901, opened 2022-06-07
Fix bug with spacing not applied to lg-heading component

### LINKED ISSUE #171: Margin directive doesn't work on the Heading component
Applying the `lgMargin` directive to `<lg-heading>`  doesn't override the margin  as you might expect - this is because the margin actually comes from the inner HTML heading tag (e.g. `<h2>` ). So when using `lgMargin`, it's applying the margin overrides to the wrapper element that Angular creates, not the `<h2>` itself.

elenagarrone:
# Description

This change fixes a bug where the spacing applied via the LgPadding or LgMargin directives was not applied.

In addition to that allows for static attributes to be added to the component host metadata (https://netbasal.com/adding-static-attributes-to-the-host-element-in-angular-b16cd33b424e).

Fixes #171
Relates to #698

https://user-images.githubusercontent.com/8397116/172418914-5fb35ee4-fe83-4242-ae60-f496cd0ed127.mov

## Requirements

A user should be allowed to add spacing to the `lg-heading` component via the LgPadding and LgMargin directives.

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/heading/heading.component.spec.ts
+++ b/projects/canopy/src/lib/heading/heading.component.spec.ts
@@ -7,13 +7,11 @@ describe('LgHeadingComponent', () => {
   let component: LgHeadingComponent;
   let fixture: ComponentFixture<LgHeadingComponent>;
 
-  beforeEach(
-    waitForAsync(() => {
-      TestBed.configureTestingModule({
-        declarations: [ LgHeadingComponent ],
-      }).compileComponents();
-    }),
-  );
+  beforeEach(waitForAsync(() => {
+    TestBed.configureTestingModule({
+      declarations: [ LgHeadingComponent ],
+    }).compileComponents();
+  }));
 
   beforeEach(() => {
     fixture = TestBed.createComponent(LgHeadingComponent);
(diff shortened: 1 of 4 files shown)


projects/canopy/src/lib/heading/heading.component.spec.ts
```diff
@@ -7,13 +7,11 @@ describe('LgHeadingComponent', () => {
...
 
-  beforeEach(
-    waitForAsync(() => {
-      TestBed.configureTestingModule({
-        declarations: [ LgHeadingComponent ],
-      }).compileComponents();
-    }),
-  );
+  beforeEach(waitForAsync(() => {
+    TestBed.configureTestingModule({
```
owensgit, line 11: Just wondering why `waitForAsync` is removed? I guess it's not needed in this case?

projects/canopy/src/lib/heading/heading.component.ts
```diff
@@ -5,7 +5,11 @@ import type { HeadingLevel } from './heading.interface';
 @Component({
   selector: 'lg-heading',
   templateUrl: './heading.component.html',
+  styleUrls: [ 'heading.component.scss' ],
   encapsulation: ViewEncapsulation.None,
+  host: {
+    class: 'lg-heading',
```
owensgit, line 11: 🙌

owensgit (approved):

HarshaDewani (approved):

PR #914, opened 2022-06-16
style(colours): add new RGB colour variants

RazvanCoada:
# Description

Adding RGB color variant to some css variables so we can use them with rgba()


elenagarrone:
Looks good, could you please only update the commit message to:

```
style(colours): add new RGB colour variants
```

RazvanCoada:
Commit message changed

PR #1015, opened 2022-09-05
Upgrade to Angular 14

elenagarrone:
# Description

- run automated migration from `ng 12` to `14`
- remove `jasmine-marbles` from `dependabot` ignores list
- update library to build specifying the project name as Angular deprecated the `defaultProject` option
- update `@types/node` and `jasmine-marbles`
- remove `ts-mockito` and `replaces` it with the forked version `@typestrong/ts-mockito`
- update `node` version to `18.8.0`
- adds `legacy-peer-deps` setting to retain previous npm behaviour of not directly installing peer dependencies (and not erroring on mismatches)
- update `date-fns` imports

The commit is flagged as a `breaking change`. Will be merging this into `next` so that we can do a quick regression installing the version in one of our projects. I don't believe we have anything else that needs to go into this major version.

Relates to: #946

# Checklist:


projects/canopy/ng-package.json
```diff
@@ -2,15 +2,6 @@
   "$schema": "../../node_modules/ng-packagr/ng-package.schema.json",
   "dest": "../../dist/canopy",
   "lib": {
-    "entryFile": "src/public-api.ts",
-    "umdModuleIds": {
```
elenagarrone, line 6: `umdModuleIds` was automatically removed

.nvmrc
```diff
@@ -1 +1 @@
-14.17.0
+18.8.0
```
aaronBery, line 1: According to the node release guidance 'Production applications should only use Active LTS or Maintenance LTS releases' and v18 doesn't technically enter active LTS until 2022-10-25. I guess it depends on how that is interpreted given Canopy is a UI library but it is ultimately used by production applications so my interpretation is that we should use v16.x.x for the next month and a bit

aaronBery: link to node releases https://nodejs.org/en/about/releases/

elenagarrone: That's a good point, although there seem to be no issue running with this version. Also, the fact that the move to active status is only in a month makes me think most libraries have already upgraded to it 🤔 

```
Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them
```

@pauleustice do you have an opinion on this since we made the same change in the monorepo too?

package.json
```diff
@@ -123,10 +124,8 @@
     "stylelint-selector-bem-pattern": "^2.1.1",
     "svg-to-ts": "^8.8.1",
     "svgo": "^2.7.0",
-    "ts-mockito": "^2.6.1",
     "ts-node": "^10.9.1",
-    "tsickle": "^0.46.3",
-    "typescript": "4.3.5",
+    "typescript": "4.6.4",
```
pauleustice, line 128: How about `4.7.4`? Not critical though...

projects/canopy-test-app/src/test.ts
```diff
@@ -22,6 +22,9 @@ declare const require: {
 getTestBed().initTestEnvironment(
   BrowserDynamicTestingModule,
   platformBrowserDynamicTesting(),
+  {
+    teardown: { destroyAfterEach: false },
```
pauleustice, line 26: Does using the default `true` behaviour break a load of stuff?

projects/canopy/src/lib/forms/checkbox-group/checkbox-group.module.ts
```diff
@@ -10,6 +10,5 @@ import { LgCheckboxGroupComponent } from './checkbox-group.component';
   imports: [ LgLabelModule, LgMarginModule, LgFocusModule ],
   declarations: [ LgCheckboxGroupComponent ],
   exports: [ LgCheckboxGroupComponent ],
-  entryComponents: [ LgCheckboxGroupComponent ],
```
pauleustice, line 13: Is this no longer a thing?

projects/canopy/src/lib/forms/date/afterDate.validator.spec.ts
```diff
@@ -1,8 +1,6 @@
 import { AbstractControl, ValidatorFn } from '@angular/forms';
-import addDays from 'date-fns/addDays';
```
pauleustice, line 2: Nice 👍

projects/canopy/src/lib/forms/checkbox-group/checkbox-group.module.ts
elenagarrone: Doesn't look like it - it was automatically removed!

pauleustice (approved):

projects/canopy/src/lib/forms/checkbox-group/checkbox-group.module.ts
elenagarrone: > With Angular 9 coming in and Ivy as the new rendering engine, all the components would be considered as entering components and do not necessarily need to be specified inside the entryComponents array.

package.json
elenagarrone: Not sure why it was updated to that version - I've now updated to the latest `4.8.2` which seem to be supported by `14`

owensgit (approved):

.gitignore
```diff
@@ -36,6 +36,7 @@ speed-measure-plugin.json
 .history/*
 
 # misc
+/.angular/cache
```
paul-request, line 39: should this just be `./angular` like we have in our consumer apps?

paul-request (approved):

.gitignore
elenagarrone: It was automatically added by the upgrade so I assumed it was fine 🤔

elenagarrone: Updated just in case 👍

projects/canopy-test-app/src/test.ts
pauleustice: @elenagarrone If we're using true, then we can remove these lines altogether since it's the default in ng14

pauleustice (approved):

Xinchro (approved):

PR #1052, opened 2022-10-12
feat(lg-row-gap): adds new directive and classes

markblandford:
* enables consumers to use the `row-gap` CSS property using Canopy and it's spacing.
* accepts either a value or will default to `sm` row-gap.
* fixes a broken link in the storybook grid style docs.

Follows on from the discussion #1046.

**This draft PR is for master but I am open to it going into [@next](https://github.com/Legal-and-General/canopy/tree/next) if required.**

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/spacing/row-gap/row-gap.notes.ts
+++ b/projects/canopy/src/lib/spacing/row-gap/row-gap.notes.ts
@@ -1,0 +1,41 @@
+export const notes = `
+This directive allows you to add the \`row-gap\` CSS property using Canopy spacing.
+The spacing variables are also available as CSS custom properties (CSS variables)
+which can be viewed in _**canopy/projects/canopy/src/styles/spacing.scss**_.
+
+This directive and accompanying CSS classes, work great with our
+<a href="./?path=/docs/directives-grid--grid">Canopy Grid</a> but will
+also work with all other HTML elements that support
+<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap" target="_blank" rel='noopener noreferrer'>CSS row-gap</a>.
+
+~~~js
+@NgModule({
+  ...
+  imports: [ ..., LgRowGapModule ],
+})
+~~~
+
+### Examples
+
+Apply \`\`xxl\`\` row-gap.
+
+~~~html
+
+<div lgRowGap="xxl"></div>
+
+~~~
+
+Use the default (\`\`sm\`\`) row-gap.
+
+~~~html
+
+<div lgRowGap></div>
+
+~~~
+
+## Inputs
+
+The current available spacing variants are \`\`none\`\`, \`\`xxxs\`\`, \`\`xxs\`\`, \`\`xs\`\`, \`\`sm\`\`, \`\`md\`\`, \`\`lg\`\`, \`\`xl\`\`, \`\`xxl\`\`, \`\`xxxl\`\`, \`\`xxxxxl\`\`.
+
+If no value is provided, the default is \`\`sm\`\`.
+`;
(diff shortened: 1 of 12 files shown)


elenagarrone (commented):
I would have expected the directive to be part of the Grid module?

projects/canopy/src/lib/spacing/row-gap/row-gap.notes.ts
```diff
@@ -0,0 +1,41 @@
...
+
+This directive and accompanying CSS classes, work great with our
+<a href="./?path=/docs/directives-grid--grid">Canopy Grid</a> but will
+also work with all other HTML elements that support
+<a href="https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap" target="_blank" rel='noopener noreferrer'>CSS row-gap</a>.
+
+~~~js
+@NgModule({
+  ...
+  imports: [ ..., LgRowGapModule ],
```
elenagarrone, line 14: As per other comment I think I would move the directive within the grid module instead of having it as a separate module. 

I would move the above notes to the grid docs too.

projects/canopy/src/lib/spacing/row-gap/row-gap.stories.ts
```diff
@@ -0,0 +1,133 @@
+import { Component, Input } from '@angular/core';
```
elenagarrone, line 1: I would merge this into the grid as well

elenagarrone: Ah you've done that 👍 I don't think there is a need for this.

projects/canopy/src/lib/spacing/row-gap/row-gap.notes.ts
markblandford: I did initially have it in grid but according to the CSS specs `row-gap` isn't just a property for flexbox (i.e. Canopy grid); it is also applicable to multi-column layouts and CSS grid.

Plus it uses the spacing 'sm', etc. so the grid stylesheet would either have to duplicate this, or create a dependency on the spacing stylesheet (the `$spacing-list` variable).

elenagarrone: Ah got it, thanks for clarifying! I think what you've done is perfectly fine then 👍

projects/canopy/src/lib/spacing/spacing.module.ts
```diff
@@ -1,11 +1,12 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 
-import { LgMarginModule } from './margin/margin.module';
-import { LgPaddingModule } from './padding/padding.module';
+import { LgMarginModule } from './margin';
```
elenagarrone, line 4: How weird, I'm pretty sure I removed the spacing module in favour of separate modules for each directive 🤔 are you sure you are up to date?

elenagarrone: I think I'm wrong as I'm seeing the same in master. I was sure I did it but I must be confusing it with the `canopy.module`. Let's keep it as is for now 👍

elenagarrone:
@markblandford looking good 👍 I think we can move it to `non-draft` and then I'll deploy it for some further testing 👍

elenagarrone (approved):
Looking great, thanks for adding this 👍

pauleustice (approved):

paul-request (approved):

projects/canopy/src/styles/grid.stories.ts
```diff
@@ -207,6 +211,30 @@ const template = `
...
 
+    <div class="lg-container">
+      <div class="lg-row">
+        <div class="lg-col-xs-12">
+          <h2>Row Gap</h2>
+          <p>
+            Add a row-gap to the flex elements using the row-gap
+            <a href="./?path=/docs/style-spacing--page">Canopy classes</a>
+            or <a href="./?path=/story/directives-row-gap--row-gap--row-gap">Canopy directive</a>.
+            </p>
```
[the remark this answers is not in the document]
elenagarrone: I think we can fix this once we do the grid documentation

PR #1124, opened 2023-01-05
feat(header): add overlay to primary nav and fix tab order

paul-request:
# Description

1. Fixes the tab order across devices as the menus use flex box to place them in different orders on different devices and this was messing with the tab ordering, when going back and forth on sm devices.
2. Adds a transparent overlay over the page when the primary navigation is open.

### DIFF
--- a/projects/canopy/src/lib/header/header-logo/header-logo.component.html
+++ b/projects/canopy/src/lib/header/header-logo/header-logo.component.html
@@ -1,20 +1,14 @@
- <ng-container
-    *ngTemplateOutlet="
-      href ? linkTpl : imgTpl;
-      context: { img: { alt: alt, href: href, src: src } }
-    "
-  ></ng-container>
+<a *ngIf="href" [attr.href]="href" class="lg-header-logo__link" #logolink>
+  <img
+    [attr.alt]="alt"
+    [attr.src]="src"
+    [ngClass]="class"
+  />
+</a>
 
- <ng-template #linkTpl let-img="img">
-   <a [attr.href]="img?.href" class="lg-header-logo__link">
-     <ng-container *ngTemplateOutlet="imgTpl; context: { img: img }"></ng-container>
-   </a>
- </ng-template>
-
- <ng-template #imgTpl let-img="img">
-   <img
-     [attr.alt]="img?.alt"
-     [attr.src]="img?.src"
-     [ngClass]="class"
-   />
- </ng-template>
+<img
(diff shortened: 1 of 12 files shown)


projects/canopy/src/lib/header/header-logo/header-logo.component.html
```diff
@@ -1,20 +1,14 @@
...
-  ></ng-container>
+<a *ngIf="href" [attr.href]="href" class="lg-header-logo__link" #logolink>
+  <img
+    [attr.alt]="alt"
+    [attr.src]="src"
+    [ngClass]="class"
+  />
+</a>
 
- <ng-template #linkTpl let-img="img">
```
elenagarrone, line 8: Why the hell did I make it so complicated last time 😂

projects/canopy/src/lib/header/header-logo/header-logo.component.ts
```diff
@@ -27,12 +34,26 @@ export class LgHeaderLogoComponent implements AfterContentChecked {
...
+  }
+
   ngAfterContentChecked(): void {
     if (this.class && this.class !== this.currentClass) {
       this.currentClass = this.class;
       this.cdr.detectChanges();
     }
   }
+
+  focus() {
```
elenagarrone, line 56: Could you add the `void` return type please?

projects/canopy/src/lib/header/header.component.ts
```diff
@@ -77,18 +86,53 @@ export class LgHeaderComponent implements AfterContentInit, OnDestroy {
         : 'lg-header-logo__second-img';
     });
 
+    // When tabbing out of the logo on sm devices, focus the account menu not the primary nav if the primary nav is open
+    this.headerLogoTabOutSubscription = this.headerLogos.last.tabbedOut
```
elenagarrone, line 90: This is fine, but I would have done it by setting up a subscription array instead of having 5 different subscriptions variables. Then you can just do a `forEach` in the `ngOnDestroy`.

projects/canopy/src/lib/header/header-logo/header-logo.component.html
paul-request: haha, I was looking at it thinking, there must be some reason for it but appears to just be a very Angular-y template-y solution to if/else

elenagarrone (approved):
Nice one 👍

Xinchro (approved):

PR #1128, opened 2023-01-10
fix(header): handle tab out event for newly added account menu items

paul-request:
# Description

Fixes an issue where account menu items were added after `ngOnContentInit`, so the item was incorrect and the focus was not set on the logo when shift + tabbing out of it. This fixes tat issue by subscribing to the changes on the querylist and and subscribing to the first item's `tabbedOut` output.

### DIFF
--- a/projects/canopy/src/lib/header/header.component.spec.ts
+++ b/projects/canopy/src/lib/header/header.component.spec.ts
@@ -1,4 +1,4 @@
-import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
+import { TestBed, waitForAsync } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { DebugElement } from '@angular/core';
 import {
@@ -16,6 +16,10 @@ import { LgPrimaryNavListItemComponent } from './primary-navigation/primary-navi
 import { LgPrimaryNavItemDirective } from './primary-navigation/primary-navigation-item.directive';
 import { LgAccountMenuItemDirective } from './account-menu/account-menu-item.directive';
 import { LgIconComponent } from './../icon/icon.component';
+import { LgAccountMenuListItemComponent } from './account-menu/account-menu-list-item/account-menu-list-item.component';
+import { LgAccountMenuComponent } from './account-menu/account-menu.component';
+import { LgNotificationBadgeComponent } from './notification-badge/notification-badge.component';
+import { LgAccountMenuItemLabelComponent } from './account-menu/account-menu-item-label/account-menu-item-label.component';
 
 describe('HeaderComponent', () => {
   let component: DefaultRenderComponent<LgHeaderComponent>;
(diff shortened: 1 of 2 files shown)


projects/canopy/src/lib/header/header.component.ts
```diff
@@ -102,9 +102,15 @@ export class LgHeaderComponent implements AfterContentInit, OnDestroy {
     );
 
     if (this.accountMenuItems.length) {
+      // When the user shift + tabs of the first account menu item, which could be added asyncronously
       this.subscriptions.push(
-        // When a user shift + tabs on first account menu item on sm devices, focus the last logo not the primary nav
-        this.accountMenuItems.first.tabbedOut
+        this.accountMenuItems.changes
+          .pipe(startWith(this.accountMenuItems))
```
pauleustice, line 108: Could we combine the three pipes?
```suggestion
          .pipe(
            startWith(this.accountMenuItems),
            switchMap((item: QueryList<LgAccountMenuListItemComponent>) => item.first.tabbedOut),
            filter((event: KeyboardEvent) => event.shiftKey && this.showResponsiveMenu),
          )
```

paul-request: Yeah, good point, will change that

projects/canopy/src/lib/header/header.component.spec.ts
```diff
@@ -400,4 +404,80 @@ describe('HeaderComponent', () => {
...
+        .nativeElement;
+
+      focusSpy = spyOn(logoEl, 'focus');
+      tabKeyDownEvent = new KeyboardEvent('keydown', { key: 'Tab', shiftKey: true });
+      preventDefaultSpy = spyOn(tabKeyDownEvent, 'preventDefault');
+
+      component.showResponsiveMenu = true;
+    });
+
+    it('prevents the default event', fakeAsync(() => {
```
pauleustice, line 448: Do these methods need to be `fakeAsync`? I can't see why they would...but I'm a bit rusty on Angular these days so take that with a pinch of salt :D

paul-request: You are right, that was copied from another test, missed that. Will remove all the `fakeAsync` as it isn't needed in any of those tests

projects/canopy/src/lib/header/header.component.spec.ts
```diff
@@ -400,4 +398,76 @@ describe('HeaderComponent', () => {
...
     });
   });
+
+  describe('with account menu', () => {
+    let tabKeyDownEvent: KeyboardEvent;
+    let focusSpy: jasmine.Spy;
+    let preventDefaultSpy: jasmine.Spy;
+
+    beforeEach(() => {
+      fixture = MockRender(`
```
elenagarrone, line 408: Could you please update all the `MockRender` to have `reset: true`? 
```
MockRender(`<hmtl code>`, {}, { reset: true })
```

Otherwise we'll get all the warnings in the logs of the unit tests.

paul-request: Had to use the flush testbed method

paul-request:
@elenagarrone @pauleustice thanks for the comments, I have addressed them all, please could you check again 🙏

elenagarrone (approved):

pauleustice (approved):

PR #1145, opened 2023-01-30
fix(header): remove overflow style on body when clicked output emits

paul-request:
# Description

Fixes a bug where clicking on primary nav items in the responsive menu prevents the page from being scrolled after the menu closes.


Xinchro (approved):

projects/canopy/src/lib/header/header.component.spec.ts
```diff
@@ -214,6 +214,14 @@ describe('HeaderComponent', () => {
       expect(component.primaryNav.showResponsiveMenu).toBe(false);
     });
 
+    it('removes overflow style from the body nav item is clicked', () => {
```
elenagarrone, line 217: I think you are missing a word here?

elenagarrone (approved):

Xinchro (approved):

PR #1185, opened 2023-04-25
fix(date-field): a11y issue with labels

paul-request:
# Description

This PR fixes an issue with the `LgDateFieldComponent` where an empty duplicate label is generated. To fix this, I have removed the custom label and just used the default label generated by the `LgInputComponent` for each of the date inputs. This removes the need to specify custom ids as that will now be handled by the default behaviour in the children`LgInputComponent`s

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/date/date-field.component.spec.ts
+++ b/projects/canopy/src/lib/forms/date/date-field.component.spec.ts
@@ -226,14 +226,6 @@ describe('LgDateFieldComponent', () => {
     expect(component.form.controls.dateOfBirth.value).toEqual('1944-03-07');
   });
 
-  it('sets unique identifiers for each input field', () => {
-    fixture.detectChanges();
-
-    expect(/lg-input-year-\d{1,3}/.test(yearInput.nativeElement.id)).toBe(true);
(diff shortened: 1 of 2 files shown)


projects/canopy/src/lib/forms/date/date-field.component.spec.ts
```diff
@@ -229,9 +229,9 @@ describe('LgDateFieldComponent', () => {
   it('sets unique identifiers for each input field', () => {
     fixture.detectChanges();
 
-    expect(/lg-input-year-\d{1,3}/.test(yearInput.nativeElement.id)).toBe(true);
```
elenagarrone, line 232: I wonder if the ids were specific for a reason 🤔

paul-request: I think it was just because the `LgInputComponent` would set the `id` by default using the internal `nextUniqueId` and when using a custom label outside of that component, there is no way to know what it is to set the `for` attribute correctly

paul-request: I can probably just remove that test now actually

paul-request: I did the same too 😄

elenagarrone (approved):

PR #1189, opened 2023-05-03
Update Canopy components to be standalone

AndreiRosu:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=ug0iRam)
# Description

Update the Accordion, Alert and Breadcrumb components to be standalone.


projects/canopy-test-app/src/app/app.module.ts
```diff
@@ -2,12 +2,16 @@ import { NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { ReactiveFormsModule } from '@angular/forms';
 import {
-  LgAccordionModule,
-  LgAlertModule,
+  LgAccordionComponent,
```
elenagarrone, line 5: I think this is slightly wrong, I would expect you to import only the `LgAccordionComponent` and that component will have the imports for the other ones. Same for all the other main components.

projects/canopy/src/lib/accordion/accordion.component.ts
```diff
@@ -26,6 +26,7 @@ let nextUniqueId = 0;
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
   providers: [ { provide: LG_ACCORDION, useExisting: LgAccordionComponent } ],
+  standalone: true,
```
elenagarrone, line 29: After this I would add:
```
  imports: [ 
    CommonModule, 
    LgHeadingModule, 
    LgIconModule,   
    LgAccordionComponent,
    LgAccordionItemComponent,
    LgAccordionPanelHeadingComponent,
    LgAccordionItemContentDirective, 
],
exports: [
    LgAccordionComponent,
    LgAccordionItemComponent,
    LgAccordionPanelHeadingComponent,
    LgAccordionItemContentDirective, 
],
```

So that you only have to import the main component.

projects/canopy/src/lib/accordion/docs/accordion.stories.ts
```diff
@@ -2,16 +2,23 @@ import { CommonModule } from '@angular/common';
...
 // This default export determines where your story goes in the story list
 export default {
   title: 'Components/Accordion/Examples',
   component: LgAccordionComponent,
   decorators: [
     moduleMetadata({
-      imports: [ CommonModule, LgAccordionModule, LgButtonModule ],
+      imports: [
+        CommonModule,
+        LgAccordionComponent,
```
elenagarrone, line 17: Again, here you should only import `LgAccordionComponent`. Same for all the others.

elenagarrone (changes requested):
Adding request changes simply because I think we can put all the imports and exports into the main component.

projects/canopy-test-app/src/app/app.module.ts
AndreiRosu: That is a very good idea!

AndreiRosu: Even if the idea by itself is very good, after doing some research on this, I came to the conclusion that it is impossible to export other components from a standalone component. So the only way to fix this would be to import every component individually. Please let me know in case you have any other ideas.

PR #1219, opened 2023-08-04
FEP-1772: Fix top padding for lg-card-footer with lg-link-menu

### LINKED ISSUE #1192: `lg-card-footer` with `lg-link-menu` padding gap
## Prerequisites

* [x] I have checked the [open Issues](https://github.com/Legal-and-General/canopy/issues) to see if this has already been raised
* [x] I have searched the [closed Issues](https://github.com/Legal-and-General/canopy/issues?q=is%3Aissue+is%3Aclosed) to see if a similar bug has been fixed in the past

## Description

This bug affects:

* [ ] GitHub workflows
* [x] Component
* [ ] Config
* [ ] Performance
* [ ] Tooling / Frameworks
* [ ] Documentation
* [ ] Something else

### Steps to Reproduce

Put a `lg-link-menu`, with a `a`/`lg-link-menu` list, in a `lg-card-footer` as part of a `lg-card`.

#### Expected behaviour:

The `lg-link-menu` sitting snugly in a `lg-card-footer` without any unintended margins and/or padding.

#### Actual behaviour:

Has some padding in the `lg-card-footer`, causing a gap above `lg-link-menu`.

#### Recurrence frequency:

* [x] Occurs every time
* [ ] Occurs intermittently
* [ ] Unsure

## Package version

* [ ] This problem started happening recently
* [ ] This problem started happening a while ago
* [x] Unsure

## Configuration and environment

* [x] This pertains to **local development**
* [x] This pertains to **the CI**
* [x] This pertains to **Storybook**

## Additional Information
This is a result of `lg-card-footer` having 32px(`--space-lg`) as padding at the top, which some other designs may rely on.

AndreiRosu:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=z3l3fp7)
# Description

Removes top padding for lg-card-footer items that have lg-link-menu as first child

Fixes #1192 

# Checklist:

- [X] The commit messages follow the [convention for this project](./blob/master/docs/CONTRIBUTING.md#conventional-commits)
- [X] I have added the functionality to the [test app](./blob/master/docs/CONTRIBUTING.md#build-test-application)
- [X] I have provided a story in storybook to document the changes

### DIFF
--- a/projects/canopy/src/lib/card/card-footer/card-footer.component.ts
+++ b/projects/canopy/src/lib/card/card-footer/card-footer.component.ts
@@ -1,7 +1,10 @@
 import {
+  AfterViewInit,
   ChangeDetectionStrategy,
   Component,
+  ElementRef,
   HostBinding,
+  Renderer2,
   ViewEncapsulation,
 } from '@angular/core';
 
@@ -12,6 +15,17 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgCardFooterComponent {
+export class LgCardFooterComponent implements AfterViewInit {
   @HostBinding('class.lg-card-footer') class = true;
+
+  constructor(private renderer: Renderer2, private hostElement: ElementRef) {}
+
+  ngAfterViewInit() {
+    const parentElem = this.hostElement.nativeElement as HTMLElement;
+    const firstChildElem = parentElem.firstChild as HTMLElement;
+
+    if (firstChildElem.localName === 'lg-link-menu') {
+      this.renderer.addClass(parentElem, 'lg-padding__top--none');
+    }
+  }
 }


projects/canopy/src/lib/card/card-footer/card-footer.component.ts
```diff
@@ -12,6 +15,17 @@ import {
   encapsulation: ViewEncapsulation.None,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class LgCardFooterComponent {
+export class LgCardFooterComponent implements AfterViewInit {
   @HostBinding('class.lg-card-footer') class = true;
+
+  constructor(private renderer: Renderer2, private hostElement: ElementRef) {}
+
+  ngAfterViewInit() {
```
elenagarrone, line 23: Missing return type

projects/canopy/src/lib/card/card-footer/card-footer.component.ts
```diff
@@ -12,6 +15,17 @@ import {
...
+    const parentElem = this.hostElement.nativeElement as HTMLElement;
+    const firstChildElem = parentElem.firstChild as HTMLElement;
+
+    if (firstChildElem.localName === 'lg-link-menu') {
```
elenagarrone, line 27: This is ok, but I'm worried we'll end up making this component full of if statements like this for different components.

I would have personally used the `lgPaddingTop=none` directive on the story element instead as per Jira ticket 🤔

AndreiRosu: We were not sure initially if the ticket should cover all the `card-footer` components that contain a `link-menu-item`, or if it was just for the story, but then came to the conclusion that the fix should be generally available. 

I totally agree with the fact that it may get polluted with if statements, but it was the only decent approach I could find.

So, reading your comment, you suggest that we should manually add `lgPaddingTop=none` to the story and that will also need to be done for any Canopy usage where we have a `link-menu-item` inside a `card-footer.`.

Xinchro: i don't like the idea of muddying the code either, but at the same time the padding being there by default is a bug
we could maybe do the inverse of `lgPaddingTop=none` and have no padding at the top unless we stick in a class to _add_ it, rather than _remove_ it?

elenagarrone: I like your idea @Xinchro but it would have to be done as a breaking change 🤔 Also, I think we would have to identify what is used more default padding or non default? Maybe one for a separate ticket?

@AndreiRosu I do think using the directive is a better option. We have done that for other components. This is just part of a pattern. If we modify the story, the code in the `guide` will reflect it and people will simply copy paste it in their codebase.

@paul-request and @owensgit any opinion on this?

owensgit: I agree with @elenagarrone on `paddingTop="none"` as it helps us build the components in a composable and flexible way as part of a pattern. We've done this quite a lot with other patterns such as the [Form Journey](https://legal-and-general.github.io/canopy/?path=/story/components-card-examples--form-journey-card). Using this directive means it's achievable for this situation but there may be another instance the designs require a different padding when a link menu is used in a card footer.  It also helps to decouple components a bit which is always nice 🙂 

As for the question about what the default should be - that's a tough one, as the current padding was created as part of some early designs/requirements for the card and probably was intended to contain buttons, which would need that padding and also to stop elements from sitting up against the grey line. So I'd say that maybe the default is okay for the most common intended (buttons at the bottom of a card) - but then can be customised with `lgPadding` in other scenarios/patterns (like with the link menu), and we can leave as-is. Having said that, I'm not against looking into changing it if we feel that defaulting to no padding and then letting the developer choose their own with `lgPadding` is better - but I would definitely do that as separate spike/ticket and as @elenagarrone it would be a breaking change as some uses of it will already depend on that default.

PR #1243, opened 2023-08-31
fix(date-field): fix -- date value and partial values

eduardh-landg:
# Description

Currently, if the date field is **required** and a user fills the date fields (day, month, year) and then clears them out the FormControl will have a status of valid.

This PR fixes the issue. Additionally, it will disallow partial invalid dates such as:
- `--`
- `1999--`
- `1999-04-`
- `-04-`
- `--25`
- `-04-25`

Fixes https://github.com/Legal-and-General/canopy/issues/184

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/date/date-field.component.ts
+++ b/projects/canopy/src/lib/forms/date/date-field.component.ts
@@ -163,7 +163,13 @@ export class LgDateFieldComponent implements OnInit, ControlValueAccessor, OnDes
           ? date.year
           : '';
 
-        this.onChange(`${year}-${month}-${day}`);
+        let formattedDate = '';
+
+        if (day && month && year) {
+          formattedDate = `${year}-${month}-${day}`;
+        }
+
+        this.onChange(formattedDate);
       }),
 
       // submit the group when the parent form is submitted
--- a/projects/canopy/src/lib/forms/date/date-field.component.spec.ts
+++ b/projects/canopy/src/lib/forms/date/date-field.component.spec.ts
@@ -169,13 +169,12 @@ describe('LgDateFieldComponent', () => {
     expect(dateInput.nativeElement.value).toEqual(date);
   });
 
-  it('joins the individual input fields and sets the value even if some are null', () => {
+  it('disallows partial dates', () => {
     fixture.detectChanges();
     dateFieldInstance.year.setValue('1970');
     dateFieldInstance.month.setValue('5');
-    dateFieldInstance.date.setValue('');
 
-    expect(component.form.controls.dateOfBirth.value).toEqual('1970-05-');
+    expect(component.form.controls.dateOfBirth.value).toEqual('');
   });
 
   it('sets the individual input fields to the empty string when no date value is provided', () => {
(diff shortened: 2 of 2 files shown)


projects/canopy/src/lib/forms/date/date-field.component.ts
```diff
@@ -163,7 +163,13 @@ export class LgDateFieldComponent implements OnInit, ControlValueAccessor, OnDes
           ? date.year
           : '';
 
-        this.onChange(`${year}-${month}-${day}`);
+        let formattedDate = '';
+
+        if (day && month && year) {
+          formattedDate = `${year}-${month}-${day}`;
+        }
```
markblandford, line 170: Personally, I prefer the ternary operator but not a big deal.

```
const formatedDate = (day && month && year) ? `${year}-${month}-${day}` : '';
```

eduardh-landg:
@elenagarrone @owensgit please review

elenagarrone:
@eduardh-landg I've deployed it to the branch environment. Could you please make sure it's been tested?

You can find it here:
https://legal-and-general.github.io/canopy/lg-sb-date-field-double-dash/?path=/story/components-forms-form-validation-examples--form-validation

and here:
https://legal-and-general.github.io/canopy/lg-sb-date-field-double-dash/?path=/story/patterns-date-input-examples--date-input

Looks like the validation on the first example is now wrong 🤔 
Screenshot 2023-09-01 at 11 10 27

eduardh-landg:
@elenagarrone I'm confused.
In the screenshot it looks like the year has been filled as "23" which according to the validation is invalid.
image
What would be the expected outcome here?

elenagarrone:
> @elenagarrone I'm confused. In the screenshot it looks like the year has been filled as "23" which according to the validation is invalid. image What would be the expected outcome here?


@eduardh-landg the problem is that there is a double message here 'Enter a valid year' and `Enter a date of birth`. It should only be 'Enter a valid year'.
This is most probably an issue in the `form.stories.ts`.

eduardh-landg:
The issue is caused by the fact that no invalid value will now be written to the form. As such, the `Validators.required` set in `form.stories.ts` kicks in additionally to the errors which are treated inside the date-field component. Additionally, any other externally added validator that fails will have the same behavior. 

I personally don't know how I should proceed about this because the behavior in my interpretation is correct.

elenagarrone:
> The issue is caused by the fact that no invalid value will now be written to the form. As such, the `Validators.required` set in `form.stories.ts` kicks in additionally to the errors which are treated inside the date-field component. Additionally, any other externally added validator that fails will have the same behavior.
> 
> I personally don't know how I should proceed about this because the behavior in my interpretation is correct.
> 
> Previously this issue was not visible because even if the date was invalid (let's say `--1994`) it was written to the form, thus required was satisfied.

The required validator shouldn't be triggered if a value is written in the field. It should only be there if there is nothing in the field.

eduardh-landg:
But there are 3 fields, and the PR I proposed ensures no invalid / partial date is written to the custom form control until all of them are filled.

eduardh-landg:
As the current PR has uncovered another issue which needs discussion, I have opened another PR which fixes just the `--` issue https://github.com/Legal-and-General/canopy/pull/1250

PR #1250, opened 2023-09-06
fix(date-field): fix -- date value

eduardh-landg:
# Description

Currently, if the date field is required and a user fills the date fields (day, month, year) and then clears them out the FormControl will have a status of valid and the value is set to `--`.

This PR fixes it and returns `''` instead.

Fixes https://github.com/Legal-and-General/canopy/issues/184

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/date/date-field.component.ts
+++ b/projects/canopy/src/lib/forms/date/date-field.component.ts
@@ -163,7 +163,11 @@ export class LgDateFieldComponent implements OnInit, ControlValueAccessor, OnDes
           ? date.year
           : '';
 
-        this.onChange(`${year}-${month}-${day}`);
+        const formatedDate = day || month || year
+          ? `${year}-${month}-${day}`
+          : '';
+
+        this.onChange(formatedDate);
       }),
 
       // submit the group when the parent form is submitted


markblandford (approved):

elenagarrone:
@eduardh-landg I'm still seeing an issue with the validation and looks like it will be a breaking change.

image

To reproduce: type something in the first field and then delete it. Both errors come up, while the expected behaviour would be `Enter a date of birth`.

eduardh-landg:
I was able to reproduce the scenario.
Unfortunately I see no way around this other then to write a value to trick required not to show which essentially negates the bug fix. I guess this can only be truly fixed after we have a discussion around single vs multiple validation messages topic.

eduardh-landg:
image
@elenagarrone should be fixed now

elenagarrone:
> image @elenagarrone should be fixed now

Looks good but I think you'll need to update the unit tests as they are failing 😅

eduardh-landg:
@elenagarrone sorry about that. I pushed at the end of the day and I was on a rush on getting feedback. I have removed 2 tests that were failing because their use case scenario doesn't apply anymore with the current change.

eduardh-landg:
@elenagarrone I've added the unit tests but unfortunately the required scenario can't be tested as it's only a Storybook story

elenagarrone (approved):
Nice one! thanks for fixing it 🙏

PR #1260, opened 2023-09-11
Add a new set of brand icons

RazvanCoada:
Add devices, quote, evidence, claim, 0 to 9 numbers brand icons


elenagarrone:
@RazvanCoada looks like the new brand icon haven't been created using the correct layers that allow us to brand them.
Could you please ask James to do that and then re-commit?

Screenshot 2023-09-11 at 15 02 17

PR #1282, opened 2023-10-02
Fix Date Field bug when used inside nested forms

st3fun1:
# Description

This PR fixes a bug where the Date Field validation didn't work properly when used as part of a form group used in a child component. 
This was caused by the `@Host` decorator, which was taking into account only the direct `FormGroupDirective` of the parent component, it just worked if the date field was a direct child of the form where `[formGroup]` was defined.

image

# Checklist:

- [x] I have provided a story in storybook to document the changes - **extended existing story**

### DIFF
--- a/projects/canopy/src/lib/forms/date/date-field.component.ts
+++ b/projects/canopy/src/lib/forms/date/date-field.component.ts
@@ -1,7 +1,6 @@
 import {
   Component,
   ContentChild,
-  Host,
   HostBinding,
   Input,
   OnDestroy,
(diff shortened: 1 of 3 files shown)


elenagarrone:
@st3fun1 looks like the validation in the form validation story is not appearing correctly for the inner date field:
Screenshot 2023-10-02 at 17 29 59

It's probably only because it needs the `pastDateValidator` too 🤔

st3fun1:
Yes, it needs that validator.

elenagarrone (approved):

paul-request (approved):
Nice one @st3fun1

PR #1297, opened 2023-10-31
Additional brand icons

anca-david:
# Description

Add new brand icons: options, fixed-term retirement plan, pensions and retirement 

# Checklist:


elenagarrone:
@anca-david just deployed the environment and it looks like the icons have been build incorrectly (in design). They need to have layers so that we can override the colours. James Hayden-Ellis should know how to do it 👍

PR #1344, opened 2024-02-13
Add role property binding to alert

eduardh-landg:
# Description

This PR adds functionality to set / override the role property on an alert.  

# Checklist:


markblandford (approved):
I think this is great. It provides us with the ability to set the `role` attribute ourselves, without which it is simply not possible (unless info or generic type of alert).

projects/canopy/src/lib/alert/alert.component.ts
```diff
@@ -30,20 +30,30 @@ export class LgAlertComponent {
...
+    this.variant = 'generic';
   }
 
-  constructor(private renderer: Renderer2, private hostElement: ElementRef) {
-    this.variant = 'generic';
+  @Input() set role(role: string) {
+    this._role = role;
+  }
+  private setRole() {
+    if (!this._role) {
```
pauleustice, line 53: Could we combine the nested `if` conditions? I can't see any value to separating them in this case.

pauleustice (changes requested):
Looks good, thanks for the contribution! Could you please add the new input to `libs/alert/docs/alert.stories.ts` too, though?

That will allow others to test the behaviour in Storybook

projects/canopy/src/lib/alert/alert.component.ts
eduardh-landg: Sure thing!

markblandford: I'm not sure I completely agree. I see this `if` as being a sort of 'guard'. I think it's possibly easier to read too, I'm not keen on `if` statements containing multiple logical operators (especially when dealing with Not).

Xinchro: if we're going to keep it separate to escape early, would it be better to make it explicit and have something along the lines of...
```
private setRole() { 
  if (!this._role) return

  if...```

eduardh-landg: https://github.com/Legal-and-General/canopy/blob/c2d78bfb37bbc4ccfa0a495b341a253f2deee2e5/projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts#L117

All roads lead to Rome. I merged the if clauses as per the initial request.

markblandford (approved):

projects/canopy/src/lib/alert/alert.component.ts
eduardh-landg: @pauleustice I also added role input in storybook.

pauleustice (approved):
Good stuff - tested in Storybook deployment and all looks good!

markblandford (approved):

owensgit (approved):

Xinchro (approved):

PR #1439, opened 2024-08-20
Improve accordion accessibility

st3fun1:
# Description

After discussing within our team, we thought it would be a good idea to let screen reader users know more details for each accordion item. To provide more context, I used the `aria-describedby` attribute set on the accordion-item heading. By setting `role="list"` on the accordion and `role="listitem"` on each accordion item, the items can be easily selected with screen readers.

### aria-describedby on an accordion-panel-heading > button
image

### role="list" set on the accordion component
image

### role="listitem" set on an accordion item
image

### the accessibility tree for the list
image

### Live demo of the solution
https://github.com/user-attachments/assets/5df5e5a6-d134-4f83-8071-e3498b3eaa54

### Selecting the list using a screen reader
For example, on Voice Over, the list can be selected with the following key combination  **VO + Command + X**.

https://github.com/user-attachments/assets/e2f5705a-292e-4238-a4ab-1c0b8c7a67b7

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts
+++ b/projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts
@@ -16,6 +16,7 @@ import {
   TemplateRef,
   ViewChild,
   ViewEncapsulation,
+  HostBinding,
 } from '@angular/core';
 import { Subscription } from 'rxjs';
 
@@ -45,6 +46,9 @@ export class LgAccordionItemComponent implements AfterContentInit, OnChanges, On
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   _contentTemplate: TemplateRef<any>;
 
+  @HostBinding('role') role = 'listitem';
+  // used for adding more context to the heading
+  @Input() ariaDescribedBy: string = null;
   @Input() isActive = false;
   @Output() opened = new EventEmitter<void>();
   @Output() closed = new EventEmitter<void>();
--- a/projects/canopy-test-app/src/app/app.component.html
+++ b/projects/canopy-test-app/src/app/app.component.html
@@ -149,7 +149,7 @@
 
             <lg-accordion [headingLevel]="3">
               <lg-accordion-item>
-                <lg-accordion-panel-heading>Item 1</lg-accordion-panel-heading>
+                <lg-accordion-panel-heading ariaDescription="Test accordion">Item 1</lg-accordion-panel-heading>
                 <p>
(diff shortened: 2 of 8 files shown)


st3fun1:
this unit test fails also on the master branch:

image

st3fun1:
Note: aria-description is still in the W3C Editor’s Draft for ARIA 1.3, so this solution isn’t entirely reliable. I suggest that the developer using the accordion add an `aria-describedby` attribute for each panel heading in their code, making this change unnecessary.

st3fun1:
Note: aria-description is still in the W3C Editor’s Draft for ARIA 1.3, so this solution isn’t entirely reliable. I suggest that the developer using the accordion add an aria-describedby attribute for each panel heading in their code, making this change unnecessary.

st3fun1:
I've revived this PR as I found a better solution for improving accessibility.

projects/canopy/src/lib/accordion/accordion-item/accordion-item.component.ts
```diff
@@ -45,6 +46,9 @@ export class LgAccordionItemComponent implements AfterContentInit, OnChanges, On
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   _contentTemplate: TemplateRef<any>;
 
+  @HostBinding('role') role = 'listitem';
+  // used for adding more context to the heading
+  @Input() ariaDescribedBy: string = null;
```
eduardhojbota, line 51: Tip: unless you have dedicated implementation for using `null` and you just need a falsy value, there's no need to assign a value to the property because by default it will be undefined.

eduardhojbota (approved):

eduardhojbota:
The code is fine but I'm wondering if this is a real accessibility improvement.

Let's look at the following scenario based on a real implementation:
h1 Upload documents
  p Please provide as much relevant documentation as you can, for us to be able to process
    your claim as quickly as possible.
  p These are examples, not a mandatory list of what we need
  lg-accordion
    lg-accordion-item 
      lg-accordion-panel-heading Employed
    lg-accordion-item
      lg-accordion-panel-heading Employed Company Director

Now the screen reader will read when navigating trough it:
- Employed - Upload documents
- Employed Company Director - Upload documents

Given the previous paragraphs, I'm unsure if this proves value or confusion.
Has this implementation been validated by someone specialized in accessibility? If not, I would ask for a second opinion. Adding this might hurt accessibility rather than improving it.

st3fun1:
This is not required at the moment, so I'm closing this PR.

PR #1441, opened 2024-08-23
Fix carousel storybook documentation and unit tests

st3fun1:
# Description

At the moment, one of the Carousel unit tests is failing due to how a transition is defined as an inline style.

image

Fixes # (issue)

To fix this, I had to remove the `ease` `transition-timing-property` from the inline style at line 16 in `projects/canopy/src/lib/carousel/carousel.component.html`, as it seems it's ignored by Angular due to being the default value for that property(this is an assumption). After doing this, I also had to update the related unit test case.

image

Using any other `transition-timing-property` value other than the default results in that value being visible in the DOM.

image

image

To test the transition fix of the carousel, I also had to fix a bug where the Storybook story of the accordion wasn't visible due to a missing configuration import in `.storybook/main.js`.

Also fixes some standalone components imports and uses the `RegistryIconService` to inject missing icons in the carousel.

# Checklist:

- [x] I have provided a story in Storybook to document the changes
- [ ] I have added any new public feature modules to `public-api.ts`

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -22,6 +22,7 @@ module.exports = {
     // Components
     '../projects/canopy/src/lib/accordion/docs/guide.stories.mdx',
     '../projects/canopy/src/lib/accordion/docs/accordion.stories.ts',
+    '../projects/canopy/src/lib/carousel/docs/carousel.stories.ts',
     '../projects/canopy/src/lib/alert/docs/guide.stories.mdx',
     '../projects/canopy/src/lib/alert/docs/alert.stories.ts',
     '../projects/canopy/src/lib/banner/docs/guide.stories.mdx',
--- a/projects/canopy/src/lib/carousel/carousel.component.html
+++ b/projects/canopy/src/lib/carousel/carousel.component.html
@@ -13,7 +13,7 @@
     class="lg-carousel__wrapper"
     [style.width.%]="100 * carouselItemCount"
     [style.left.%]="-100 * selectedItemIndex"
-    [style.transition]="'left ease ' + slideDuration / 1000 + 's'"
+    [style.transition]="'left ' + slideDuration / 1000 + 's'"
   >
     <ng-content></ng-content>
   </div>
--- a/projects/canopy/src/lib/carousel/carousel.component.spec.ts
+++ b/projects/canopy/src/lib/carousel/carousel.component.spec.ts
@@ -131,7 +131,7 @@ describe('LgCarouselComponent', () => {
     expect(component.carouselItemCount).toBe(3);
 
     expect(wrapperElement.attributes['style']).toBe(
-      'width: 300%; left: 0%; transition: left 0.5s ease 0s;',
+      'width: 300%; left: 0%; transition: left 0.5s;',
     );
   });
 
--- a/projects/canopy/src/lib/carousel/carousel.component.ts
+++ b/projects/canopy/src/lib/carousel/carousel.component.ts
@@ -41,6 +41,9 @@ import { LgAutoplayComponent } from './auto-play/auto-play.component';
     NgClass,
   ],
 })
+/**
+ * @deprecated This component will be removed in the future
+ */
 export class LgCarouselComponent implements AfterContentInit, OnDestroy {
   private unsubscribe: Subject<void> = new Subject<void>();
   selectedItem: LgCarouselItemComponent;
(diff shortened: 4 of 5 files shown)


projects/canopy/src/lib/carousel/carousel.component.html
```diff
@@ -13,7 +13,7 @@
     class="lg-carousel__wrapper"
     [style.width.%]="100 * carouselItemCount"
     [style.left.%]="-100 * selectedItemIndex"
-    [style.transition]="'left ease ' + slideDuration / 1000 + 's'"
```
st3fun1, line 16: It seems the latest versions of Angular don't take into account default values for inline styles, for example the the default value for the CSS property `transition-timing-function` is `ease`. Using it here is redundant.

projects/canopy/src/lib/carousel/carousel.component.spec.ts
```diff
@@ -127,7 +127,7 @@ describe('LgCarouselComponent', () => {
     expect(component.carouselItemCount).toBe(3);
 
     expect(wrapperElement.attributes['style']).toBe(
-      'width: 300%; left: 0%; transition: left 0.5s ease 0s;',
```
st3fun1, line 130: By removing `ease` from the inline styling, this test had to be updated.

projects/canopy/src/lib/carousel/carousel.component.html
st3fun1: This assumption can be checked by using a different timing function. Will provide a screenshot soon.

.storybook/main.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     // Components
     '../projects/canopy/src/lib/accordion/docs/guide.stories.mdx',
     '../projects/canopy/src/lib/accordion/docs/accordion.stories.ts',
+    '../projects/canopy/src/lib/carousel/docs/carousel.stories.ts',
```
st3fun1, line 25: I had to add the missing storybook story to this list as the Carousel wasn't available to be tested.

eduardhojbota (approved):

robbrownsell (approved):

projects/canopy/src/lib/carousel/carousel.component.ts
```diff
@@ -14,11 +14,12 @@ import { filter, map, takeUntil, withLatestFrom } from 'rxjs/operators';
 import { NgIf, NgFor, NgClass } from '@angular/common';
 
 import type { HeadingLevel } from '../heading';
-import { LgIconComponent } from '../icon';
+import { LgIconComponent, lgIconChevronLeft, lgIconChevronRight } from '../icon';
 import { LgHeadingComponent } from '../heading';
 
 import { LgCarouselItemComponent } from './carousel-item/carousel-item.component';
 import { LgAutoplayComponent } from './auto-play/auto-play.component';
+import { LgIconRegistry } from './../icon/icon.registry';
```
pauleustice, line 22: I'm surprised the linter didn't complain about this; the `../` paths should all be grouped together. I think you tricked it by starting with `./` :D

pauleustice (changes requested):
Just one tiny change needed with an import path

projects/canopy/src/lib/carousel/carousel.component.ts
st3fun1: ohhh, I used auto import, I will fix this

.storybook/main.js
elenagarrone, line 25: This was actually done on purpose as the use of the Carousel should be discouraged. This was obviously a decision that was taken with @alexcanning before leaving for maternity leave so not sure if anything changed in the meantime!

projects/canopy/src/lib/carousel/carousel.component.ts
st3fun1: I've fixed this

elenagarrone (approved):
Screenshot 2024-08-30 at 09 09 56

Looks good to me 👍

projects/canopy/src/lib/carousel/carousel.component.ts
elenagarrone: I thought what Paul meant was move it to line 18 and remove the `./` so you have `import { LgIconRegistry } from '../icon/icon.registry';` but i might be wrong!

elenagarrone (approved):

pauleustice (approved):

PR #1487, opened 2024-12-02
Add link styles to banner message

### LINKED ISSUE #1417: Banner message content link colours are not contrast compliant
## Prerequisites

* [x] I have checked the [open Issues](https://github.com/Legal-and-General/canopy/issues) to see if this has already been raised
* [x] I have searched the [closed Issues](https://github.com/Legal-and-General/canopy/issues?q=is%3Aissue+is%3Aclosed) to see if a similar bug has been fixed in the past

## Description

This bug affects:

* [ ] GitHub workflows
* [x] Component
* [ ] Config
* [ ] Performance
* [ ] Tooling / Frameworks
* [ ] Documentation
* [ ] Something else

### Steps to Reproduce

Using links inside the content of the banner message component results in non compliant link colours.

#### Expected behaviour:

The contrast of the links should be compliant.

#### Actual behaviour:

The links fail contrast checks.

#### Recurrence frequency:

* [x] Occurs every time
* [ ] Occurs intermittently
* [ ] Unsure

## Package version

* [ ] This problem started happening recently
* [ ] This problem started happening a while ago
* [x] Unsure

## Configuration and environment

* [x] This pertains to **local development**
* [ ] This pertains to **the CI**
* [x] This pertains to **Storybook**

## Additional Information

I understand from the documentation that this component is a work in progress but we should change the standard blue link colour  for both variants until we decide the link/button styling.

Screenshot 2024-06-10 at 16 53 00

Screenshot 2024-06-10 at 16 54 17

It would be good to include some link text in the Storybook example similar to the Inline message component.

Screenshot 2024-06-10 at 17 07 24

mchand:
# Description

- adds link colours to the banner message component so any links used in the projected content are contrast compliant
- adds a container to the projected content so links are aligned correctly (and to allow for dismiss functionality later)
- adds link text to the component example

Fixes #1417

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/banner/banner.component.scss
+++ b/projects/canopy/src/lib/banner/banner.component.scss
@@ -1,13 +1,42 @@
+@import '../../styles/mixins';
+
 .lg-banner {
   display: block;
   background-color: var(--color-battleship-grey);
   color: var(--color-white);
   box-shadow: 0 0.25rem 0 0 rgba(0, 0, 0, 0.03);
+
+  a:not(.lg-btn) {
+    border-bottom-color: var(--color-white);
+    color: var(--color-white);
+
+    @include lg-link(
+      $default-color: var(--color-white),
+      $hover-color: var(--color-white),
+      $visited-color: var(--color-white),
+      $active-color: var(--color-white),
+      $active-bg-color: var(--color-charcoal),
+      $focus-color: var(--color-charcoal),
+      $focus-bg-color: var(--color-white)
+    );
+  }
 }
 
 .lg-banner-variant--warning {
   background-color: var(--color-dandelion-yellow);
   color: var(--color-charcoal);
+
+  a:not(.lg-btn) {
+    @include lg-link(
+      $default-color: var(--color-charcoal),
+      $hover-color: var(--color-charcoal),
+      $visited-color: var(--color-charcoal),
+      $active-color: var(--color-charcoal),
+      $active-bg-color: var(--color-white),
+      $focus-color: var(--link-focus-color),
+      $focus-bg-color: var(--color-charcoal)
+    );
+  }
 }
 
 .lg-banner__container {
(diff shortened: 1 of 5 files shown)


projects/canopy/src/lib/banner/banner.component.scss
```diff
@@ -1,13 +1,42 @@
...
   background-color: var(--color-battleship-grey);
   color: var(--color-white);
   box-shadow: 0 0.25rem 0 0 rgba(0, 0, 0, 0.03);
+
+  a:not(.lg-btn) {
+    border-bottom-color: var(--color-white);
+    color: var(--color-white);
+
+    @include lg-link(
+      $default-color: var(--color-white),
```
elenagarrone, line 14: Because Canopy is fully themable, we should really use hard-coded colour variables.
In this case, what you need is to use the `light-foreground` variables that you can find listed in `styles/variables/_variants.scss`.

Before making anymore changes - have you tried applying the `lgVariant` directive to it instead? that might work 🤔 it would need a bit of cleanup in the styles and in `set variant`. Happy to jump on a call tomorrow if you want to pair?

mchand: I did try to apply the `lgVariant` directive but the problem was the colours used in the variants aren't the same as what has been designed for the banner message component. The variant colours are more pastel than dark and bold in the banner message. Do we need to update the variant colours globally or maybe add more variants? 🤔 Yes, always happy to jump on a call and pair 🙂.

mchand: As discussed, I have replaced the colours with the light/dark variables and created some others for the banner background and text colours.

elenagarrone (approved):
Thanks for fixing this! 👏 and congrats on your first Canopy PR 🎉

elenagarrone:
Just deployed and it's looking good:
image
image

Xinchro (approved):

PR #1506, opened 2025-02-28
FEP-3096: Upgrade date-fns to v4.1.0

AndreiRosu:
# Description

This PR:
- Updates date-fns package to latest version (v4.1.0)
- Fixes the resulting build issues by changing tsconfig compiler options for libs from es2018 to ES2022
- Fixes issues that resulted after default string conversion has been removed from the package (https://github.com/date-fns/date-fns/pull/3141/commits/1e50bced1aa131b79ed2172a62c3727fe8e716e2#diff-ff2296eb27d158df704d066e8e91bbfc4191b0aa87003b24e232081c8b10440c)

All the unit tests are passing (`npm run test`):
Screenshot 2025-03-03 at 11 38 19

# Checklist:

### DIFF
--- a/projects/canopy/src/lib/forms/date/afterDate.validator.ts
+++ b/projects/canopy/src/lib/forms/date/afterDate.validator.ts
@@ -9,7 +9,7 @@ export function afterDateValidator(dateToCompare: Date): ValidatorFn {
   }
 
   return (control: AbstractControl): ValidationErrors | null => {
-    const date = parseISO(control.value);
+    const date = parseISO(control.value || '');
 
     return !isValid(date) || isAfter(date, dateToCompare)
       ? null
--- a/projects/canopy/src/lib/forms/date/futureDate.validator.ts
+++ b/projects/canopy/src/lib/forms/date/futureDate.validator.ts
@@ -3,9 +3,9 @@ import { isFuture, isValid, parseISO } from 'date-fns';
 
 export function futureDateValidator(): ValidatorFn {
   return (control: AbstractControl): ValidationErrors | null => {
-    const date = parseISO(control.value);
+    const date = parseISO(control.value || '');
 
-    return !isValid(date) || isFuture(parseISO(control.value))
+    return !isValid(date) || isFuture(parseISO(control.value || ''))
       ? null
       : { futureDate: true };
   };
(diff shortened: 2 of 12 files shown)


projects/canopy/src/lib/forms/date/afterDate.validator.ts
```diff
@@ -9,7 +9,7 @@ export function afterDateValidator(dateToCompare: Date): ValidatorFn {
   }
 
   return (control: AbstractControl): ValidationErrors | null => {
-    const date = parseISO(control.value);
```
AndreiRosu, line 12: The default conversion to string [has been removed](https://github.com/date-fns/date-fns/pull/3141/commits/1e50bced1aa131b79ed2172a62c3727fe8e716e2#diff-ff2296eb27d158df704d066e8e91bbfc4191b0aa87003b24e232081c8b10440c), which means providing an invalid type will result in an error, because it will try to use the `split` method on a non-string:

`Chrome 133.0.0.0 (Mac OS 10.15.7) afterDate returns null if the date is null FAILED | TypeError: Cannot read properties of null (reading 'split')`

projects/canopy/src/lib/forms/date/futureDate.validator.ts
```diff
@@ -3,9 +3,9 @@ import { isFuture, isValid, parseISO } from 'date-fns';
 
 export function futureDateValidator(): ValidatorFn {
   return (control: AbstractControl): ValidationErrors | null => {
-    const date = parseISO(control.value);
+    const date = parseISO(control.value || '');
 
-    return !isValid(date) || isFuture(parseISO(control.value))
+    return !isValid(date) || isFuture(parseISO(control.value || ''))
```
pauleustice, line 8: Couldn't this be `!isValid(date) || isFuture(date)`?

pauleustice (changes requested):
One small improvement we can make while we're here

elenagarrone (dismissed):
Looks good to me 👍

projects/canopy/src/lib/forms/date/futureDate.validator.ts
AndreiRosu: Definitely. Good spot! I've updated it.

pauleustice (approved):

elenagarrone (approved):

Issue #1567, opened 2025-09-12
Broken underline/icon alignment for lg-button "link" variant


AndreiRosu:
## Prerequisites

* [x] I have checked the [open Issues](https://github.com/Legal-and-General/canopy/issues) to see if this has already been raised
* [x] I have searched the [closed Issues](https://github.com/Legal-and-General/canopy/issues?q=is%3Aissue+is%3Aclosed) to see if a similar bug has been fixed in the past

## Description

The underline styling for the "link" variant of the `lg-button` appears broken when the text wraps across multiple lines. Additionally, the icon should be positioned after the text.

Note: Take into consideration the hover effect as well

This bug affects:

* [ ] GitHub workflows
* [x] Component
* [ ] Config
* [ ] Performance
* [ ] Tooling / Frameworks
* [ ] Documentation
* [ ] Something else

### Steps to Reproduce

1. Add an lg-button "link" variant with an icon

#### Expected behaviour:

<img width="794" height="178" alt="Image" src="https://github.com/user-attachments/assets/edb9b7da-716e-4c34-8ec7-36450f702f5b" />

#### Actual behaviour:

<img width="828" height="202" alt="Image" src="https://github.com/user-attachments/assets/71200c6e-5901-4126-bffe-51b1642a6a0b" />

#### Recurrence frequency:

* [x] Occurs every time
* [ ] Occurs intermittently
* [ ] Unsure

## Package version

* [ ] This problem started happening recently
* [x] This problem started happening a while ago
* [ ] Unsure

## Configuration and environment

* [x] This pertains to **local development**
* [ ] This pertains to **the CI**
* [ ] This pertains to **Storybook**

eduardhojbota:
I've simply replaced border-bottom for links with text-decoration underline and it looks exactly the same. Please review https://github.com/Legal-and-General/canopy/pull/1573.

PR #1573, opened 2025-09-15
Fix link styling

eduardhojbota:
# Description

**DO NOTICE** that this fix removes the underline from the icon.
Previously:
image
Now:
image
This should be fine as it follows GOV.UK design standards https://design-system.service.gov.uk/components/details/ and from that I read it's also not recommended to underline the icon.

Fixes # (issue)
https://github.com/Legal-and-General/canopy/issues/1567

Screenshots:
lg-button of type link with very long text:
link_button_icon_long_text
lg-button of type link with alternating state: normal, hover, active, focus-visible
link_normal_hover_active_focus-visible

# Checklist:

### DIFF
--- a/projects/canopy/src/styles/mixins.scss
+++ b/projects/canopy/src/styles/mixins.scss
@@ -33,31 +33,26 @@
   $focus-color: var(--link-focus-color),
   $focus-bg-color: var(--link-focus-bg-color)
 ) {
-  $box-shadow-inset-width: -0.063rem;
-
-  border-bottom: 0.125rem solid $default-color;
   color: $default-color;
   line-height: initial;
-  padding: 0 0.125rem;
-  text-decoration: none;
+  text-decoration: underline;
+  text-decoration-color: $default-color;
 
   &:hover,
   &:hover:visited {
     color: $hover-color;
-    border-bottom: 0;
-    box-shadow:
-      inset 0 $box-shadow-inset-width 0 0 $hover-color,
-      0 0.188rem 0 0 $hover-color;
+    text-decoration-color: $hover-color;
+    text-decoration-thickness: 0.188rem;
   }
 
   &:visited {
     color: $visited-color;
-    border-color: $visited-color;
+    text-decoration-color: $visited-color;
   }
 
   &:active {
     background-color: $active-bg-color;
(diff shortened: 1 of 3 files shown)


projects/canopy/src/styles/mixins.scss
```diff
@@ -33,39 +33,34 @@
...
   $focus-bg-color: var(--link-focus-bg-color)
 ) {
-  $box-shadow-inset-width: -0.063rem;
-
-  border-bottom: 0.125rem solid $default-color;
   color: $default-color;
   line-height: initial;
-  padding: 0 0.125rem;
-  text-decoration: none;
+  text-decoration: underline;
```
elenagarrone, line 38: I think if we change this mixin, we will have to change the link itself too, otherwise we'll have discrepancies in how buttons styled as links and links will look.

If Alex agrees to change the pattern, we should also add:
```
    text-decoration-thickness: 0.125rem;
    text-underline-offset: 0.125rem;
```

eduardhojbota: Links will be changed and will look the same as they use the same mix in.
https://github.com/Legal-and-General/canopy/blob/5dd34a8bc74cd6cd92e8248b40f6f9b215381050/projects/canopy/src/styles/mixins.scss#L280
I've glanced in Storybook over the elements that use this mix in and I've not seen visible changes.

elenagarrone: There is quite a lot of difference in how the links/buttons styled as links look now, even if we don't extend the sides, we should at least make them look more like the design: https://percy.io/ffaa5218/web/LG-Canopy/builds/43112747

I think those property should do the job. Might need to look at the other states though.

elenagarrone:
Alex and I talked through this change and we think it would lead to a lot of visual differences and potential bugs across apps that use links or buttons styled as links. Especially with the brand modernisation work starting now, and knowing that link styles will be updated as part of that, it feels like the wrong time to make this change. 

We'd suggest going with the approach [mentioned in the issue](https://github.com/Legal-and-General/canopy/issues/1567#issuecomment-3290759935) instead as it’s more in line with what’s coming and avoids unnecessary disruption.

PR #1580, opened 2025-09-17
feat(brand-modernisation): update codeql workflow

IdaHW:
# Description

Update so the codeql analysis workflow will be triggered on pull requests to the master-bm branch as well. Include scan of workflows and actions.

# Checklist:

### DIFF
--- a/.github/workflows/codeql_analysis.yml
+++ b/.github/workflows/codeql_analysis.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ "master", "next", "master-bm" ]
   pull_request:
-    branches: [ "**" ]
+    types: [ opened, synchronize ]
 
 jobs:
   analyze:
@@ -13,6 +13,8 @@ jobs:
 
     permissions:
       security-events: write
+      pull-requests: write
+      statuses: write
       packages: read
       actions: read
       contents: read
@@ -20,11 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-        - language: actions
-          build-mode: none
-        - language: javascript-typescript
-          build-mode: none
+        language: [ 'actions', 'javascript-typescript' ]
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql_analysis.yml
```diff
@@ -1,30 +1,30 @@
...
-# or to provide custom queries or build logic.
 name: 'CodeQL'
 
 on:
   push:
-    branches:
-      - master
-      - next
+    branches: [ "master", "next", "master-bm" ]
   pull_request:
```
elenagarrone, line 6: I think we can change this to:

```
  pull_request:
    types: [ opened, synchronize ]
```
I think they do exactly the same thing but I find it more readable.

IdaHW: OK, I will amend this as that makes sense

.github/workflows/codeql_analysis.yml
```diff
@@ -1,30 +1,30 @@
...
-      - next
+    branches: [ "**" ]
 
 jobs:
   analyze:
-    name: Analyze
+    name: Analyze (${{ matrix.language }})
     runs-on: ubuntu-latest
 
+    permissions:
```
elenagarrone, line 14: Just comparing this to the monorepo, I think we need a couple more:

```
  pull-requests: write
  statuses: write
```
I think it requires these two as it writes comments on the PR itself if something is not right and also adds the status check.

elenagarrone (commented):
Nice work! just a couple of comments

.github/workflows/codeql_analysis.yml
```diff
@@ -1,30 +1,30 @@
...
+      packages: read
+      actions: read
+      contents: read
+
     strategy:
       fail-fast: false
       matrix:
-        language: ['javascript']
+        include:
+        - language: actions
```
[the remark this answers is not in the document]
IdaHW: I think it can be either way, but don't mind updating as a one liner

PR #1826, opened 2026-05-08
feat(inline-message): update inline message to new brand

anca-david:
BREAKING CHANGE: making the icon costumisable for generic and info statuses and adding design tokens variables

# Description

Updated the Alert component to support customizable icons and new design tokens:

**Icons:** Added [icon] input allowing custom icons for generic and info statuses; other statuses use fixed branded icons
**All Statuses Now Show Icons**: Previously generic had no icon; now all statuses display icons by default
**Design Tokens**: Replaced generic spacing utilities with design-specific tokens (--inline-message-padding-y, --inline-message-padding-x, --inline-message-gap)
**Docs**: Updated Storybook with typed args, and documented the new feature in guide and skill

## Requirements

Inline message has five variants: info, success, warning, error, and generic.
All variants except info and generic have fixed icons; generic allows custom icon selection from the icon library.
The default for the info status should be the info icon
Inline message conforms to content width, has rounded edges, and uses status tokens for colour.
Heading/title is not built-in but can be added by users via HTML.
There is a separate optional link at the end of the content (under the text). 

Design link: https://www.figma.com/design/FveBgdquAdUyB26zut7z9z/Core---2025?node-id=4780-12304&m=dev

# Checklist:


elenagarrone:
Apologies if this is following designs and I'm wrong, but this doesn't look right to me:

image

The spacing looks off and I would have thought the "Important update" would have to be bold? 

Also the "read more" link seems to have some side padding that make it misaligned to the rest of the content.
Should the `i` in `inline link` be capitalised since there is a full stop before it?

elenagarrone:
image

In the controls, the `genericIcon` seems to set the icon for the component, but not `icon` which I think is the correct input.

EDIT:
I see what is happening but this approach makes it very confusing in my opinion 😅 You should be able to use `icon` from there to select the correct icon.

IdaHW (approved):

PR #1939, opened 2026-07-27
Cn 340 card changes

IdaHW:
# Description

This PR modernises the Card component to align with the new brand, updating its variants, layout structure, and Storybook/docs to reflect the refreshed patterns.

### Changes:

- Renames card variants (e.g. promotion → promo, navigation → interactive) and updates docs/stories/tests accordingly.
- Refactors card layout/styling (new .lg-card__body wrapper, updated header/content/footer spacing, new orientation rules, updated hero image behaviour).
- Removes legacy card CSS variable definitions and updates mixins/components to use new --card-* design tokens.

# Checklist:


anca-david (dismissed):

.github/workflow-scripts/__tests__/check-docs-skills.spec.cjs
```diff
@@ -119,6 +123,24 @@ describe('checkDocsSkills', () => {
     });
   });
 
+  describe('when only deleted .mdx files are in the PR', () => {
```
elenagarrone: When mdx files are in the PR the skill should be removed so I think how it was before it's correct 🤔 
Remember there is also the label that you can add to ignore the check.

What is your scenario exactly? I might be missing something!

projects/canopy/src/lib/card/card-footer/card-footer.component.spec.ts
```diff
@@ -1,14 +1,81 @@
...
 
+import {
+  LgLinkMenuComponent,
+  LgLinkMenuItemComponent,
+  LgLinkMenuItemTextComponent,
+} from '../../link-menu';
+
 import { LgCardFooterComponent } from './card-footer.component';
 
+@Component({
```
elenagarrone: Could you please use `ng-mocks` instead of creating dummy components? I think that's a better pattern 🙂

projects/canopy/src/lib/card/card-hero-img/card-hero-img.component.scss
```diff
@@ -11,238 +43,168 @@
...
-  &.lg-orientation--sm--reverse-horizontal {
-    .lg-card-hero-img__icon {
-      margin: 0 0 0 var(--space-6);
-    }
-  }
+lg-card-hero-img {
+  display: block;
+  width: 100%;
+  flex: 0 0 auto;
+  aspect-ratio: 4 / 3;
```
elenagarrone, line 54: Should this come from a design token?

projects/canopy/src/lib/card/docs/promotions/guide.mdx
```diff
@@ -1,15 +1,12 @@
...
 import CardHeroImageComponent from '../card-hero-image-component.md';
 import Feedback from '../../../docs/common/feedback.md';
-import { CustomBadge } from 'storybook-addon-tag-badges/manager-helpers'
 
-<Meta title="Patterns/Promotions/Guide" tags={['pending']}/>
+<Meta title="Patterns/Promotions/Guide"/>
 
 # Promotions
 
-<CustomBadge text="Pending" style={{ backgroundColor: '#005dba', borderColor: '#005dba', color: '#fff' }} /> <span class="small-text">Due for brand modernisation</span>
```
elenagarrone, line 10: Should we add the small text for the version it will go in (placeholder)?

projects/canopy/src/lib/card/card.component.ts
```diff
@@ -30,15 +36,30 @@ import { LgCardNavigationTitleComponent } from './card-navigation-title/card-nav
   standalone: true,
 })
 export class LgCardComponent implements AfterContentInit, OnDestroy {
-  private subscription: Subscription;
-  private uniqueId = randomUniqueId();
+  private static readonly minWidths = {
```
elenagarrone, line 39: I'm not sure I like that the breakpoints are hardcoded here :/ Ideally, TypeScript shouldn't be handling responsive breakpoints as that's the responsibility of CSS and makes things prone to errors. From what I see, it looks like this is just being used to toggle the orientation class? Why don't we use the existing `LgOrientation` directive here instead?

projects/canopy/src/lib/card/card.interface.ts
```diff
@@ -1,3 +1,3 @@
 export const lgCardToggleIdPrefix = 'lg-card-toggle-';
 export const lgCardPanelIdPrefix = 'lg-card-panel-';
-export type CardVariant = 'default' | 'navigation' | 'promotion';
+export type CardVariant = 'default' | 'promo' | 'interactive';
```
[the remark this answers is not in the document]
elenagarrone: @IdaHW I know you marked this as resolved but could you please explain why? e.g. I have checked and this is a false positive. It makes it a bit clearer for the reviewer and means they don't have to double check 😅

.github/workflow-scripts/__tests__/check-docs-skills.spec.cjs
IdaHW: It was throwing an error as the promo-card skill was deleted in the pr but was still expecting it. Can use the label instead if it's better

projects/canopy/src/lib/card/card-hero-img/card-hero-img.component.scss
IdaHW: There's no tokens for the aspect ratios in Figma

projects/canopy/src/lib/card/docs/promotions/guide.mdx
IdaHW: Can do!

projects/canopy/src/lib/card/card.component.ts
IdaHW: This is required to make the `default` variant, with Pictogram and horizontal layout work. This has been the trickiest part of the card layout as the Pictogram has to be vertically aligned within the card content only in this scenario and in all other variations the pictogram sits before `lg-card__body`.

Here it checks what screen size it is, and for that size is the card horizontal or vertical, if horizontal and hero is a pictogram - > move the pictogram into the content, `lg-card-content` so it can sit correctly within the card. If not stay the same as all other variants and layouts.
The breakpoints are used to pick the active orientation at the current viewport width.

I agree it would be better to use CSS but couldn't find a way to get that to work.

There was one other option to get the Pictogram align vertically within the card content and that was to absolute position the pictogram instead, to get it vertically centred at all times (when missing either card header or footer). That involved observing the card contents, check if there's a header or footer and set an inline centre offset on the Pictogram element. If you think that can be better I can change it back to that?

Screenshot 2026-08-12 at 15 07 47

Screenshot 2026-08-12 at 15 08 53


Screenshot 2026-08-12 at 15 09 37

PR #2022, opened 2026-09-16
fix(date-field): amend box sizing

IdaHW:
# Description

--Testing--

Fix the date field year being cut of on mobile devices

# Checklist:


projects/canopy/src/lib/forms/date/date-field.component.scss
```diff
@@ -23,3 +23,7 @@
 .lg-date-field__date {
   width: 4rem;
 }
+
+.lg-date-field__year {
+  width: 5rem;
```
elenagarrone, line 28: I think there is a bit of a clash here. In the template, we set the `size` of the field too which would affect this and the width on line 24 - we should either set it here or there.

I also think, if we set it here, we need to put these into design tokens.

Generally I cannot replicate the issue in Canopy (`master` version) - have you been able to?

IdaHW: Thanks @elenagarrone The issue only appears on real mobile devices.
I think the date field is being re-build shortly so we can look at updating either the template or this then.

elenagarrone (approved):

